PR #15, opened 2018-08-16
DNM / Discussion: Fix for Bone Scaling

Menithal:
This is a specific fix for the issue of bones being too tiny and not scaled correctly. 

DNM is for clarification and discussion on the subject and to double check conventions, and due to still incomplete fix: tested with Binary version of Celsius Man and Monster (which seems to be offset, separate issue). Will post an issue related to the Monster, later.. 

Not Perfect, as joint ends are most likely not upto spec (will look into just using the previous bone length in the future).

Bones in Blender inherently have no scale: instead they have a start point (head, or joint position) and an endpoint (tail, child point.). This is something to keep in mind when designing the nodes.

From what I understood, the idea is to put most of the correction logic into the Nodes. However we cannot rely on the information in them while the scene is building, as information comes in out of order.


julienduroure:
Hello,

Problem is much more complicated.
Your solution will work for bone chains, but every bone are not connected ( even if a bone have only 1 child), so this code won't work for not connected bones.
A still missing feature is bone scale animation: you remove TODOs in code, but it has to be implemented, because bones have scale in pose mode (even if there don't have scale in edit mode).

Currently working on a more robust solution ( not committed yet)

PR #56, opened 2018-10-23
Disallow uint8 indices.

donmccurdy:
/cc https://github.com/KhronosGroup/glTF-Blender-Exporter/pull/329

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_generate.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_generate.py
@@ -1136,9 +1136,9 @@ def generate_meshes(operator,
 
             max_index = max(indices)
 
-            if max_index < 65536:
+            if max_index < 65535:
                 componentType = GLTF_COMPONENT_TYPE_UNSIGNED_SHORT
-            elif max_index < 4294967296:
+            elif max_index < 4294967295:
                 componentType = GLTF_COMPONENT_TYPE_UNSIGNED_INT
             else:
                 print_console('ERROR', 'Invalid max_index: ' + str(max_index))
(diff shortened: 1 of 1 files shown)


UX3D-nopper:
see my comment https://github.com/KhronosGroup/glTF/issues/1471

We cannot just remove it, as some people maybe need it.

lexaknyazev:
For compatibility, can we keep uint8 import while removing uint8 export?

lexaknyazev:
Also, regardless of the related spec issue outcome, `65536` should be changed to `65535` to avoid primitive restart.

emackey:
> We cannot just remove it, as some people maybe need it.

Are there people who specifically need UINT8 export from Blender via this plugin?  That seems like too small an edge case.

FWIW, in my conversations with artists, the "maximum indicies" options with different integer sizes are the most confusing of all this plugin's options.  I think it's the only one that doesn't reference a feature they know about from Blender, but rather references an internal detail of the file storage format that they've never heard of.

donmccurdy:
Thanks! I've reduced the maximum index values by 1, per https://github.com/KhronosGroup/glTF/pull/1476/files.

addons/io_scene_gltf2/blender/exp/gltf2_blender_generate.py
```diff
@@ -1132,15 +1132,13 @@ def generate_meshes(operator,
...
-            if max_index < 256:
-                componentType = GLTF_COMPONENT_TYPE_UNSIGNED_BYTE
-            elif max_index < 65536:
+            if max_index < 65535:
                 componentType = GLTF_COMPONENT_TYPE_UNSIGNED_SHORT
-            elif max_index < 4294967296:
+            elif max_index < 4294967295:
                 componentType = GLTF_COMPONENT_TYPE_UNSIGNED_INT
             else:
                 print_console('ERROR', 'Invalid max_index: ' + str(max_index))
```
lexaknyazev: Could the error message here be more descriptive? Something like:
> A mesh contains too many vertices and needs to be split before export.

donmccurdy: Done

PR #106, opened 2018-11-29
Rename extension options.

donmccurdy:
I think it may be better to name these options for what they do, rather than the extension name.

/cc @emackey

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -258,19 +258,19 @@ class ExportGLTF2_Base:
     )
 
     export_lights = BoolProperty(
-        name='Export KHR_lights_punctual',
+        name='Export punctual lights',
         description='',
         default=False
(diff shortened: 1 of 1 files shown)


addons/io_scene_gltf2/__init__.py
```diff
@@ -258,19 +258,19 @@ class ExportGLTF2_Base:
     )
 
     export_lights = BoolProperty(
-        name='Export KHR_lights_punctual',
+        name='Export punctual lights',
         description='',
```
emackey: I'm fine with this rename.

Question, does this description show up on the UI?  Is it worth reporting the extension name here?

donmccurdy: The descriptions do show up in the UI, yes. Great idea. I'll do a pass on this and fill in all of the descriptions — seems like they may be helpful even for non-extension options.

Issue #118, opened 2018-11-30
Process for contributing


donmccurdy:
Now that this addon is in the Blender 2.8 beta (🎊🎊🎊), I'm not sure how to develop against the GitHub repository. The addon folder created by Blender 2.8 seems to be readonly, and on my work computer I cannot remove it. At the moment I am:

1. Copy GitHub folder to a folder with a different name, 'io_scene_gltf2_custom'
2. Edit `__init__.py` and give a different name than the existing addon.

Is there a better way to contribute? Just disabling the official extension doesn't seem to do it. We should probably add this to the documentation for would-be external contributors.

donmccurdy:
Hm, unfortunately the steps above aren't working for me well — when making changes, it seems like the official code is still interfering, even though the addon is disabled.

UX3D-nopper:
In Blender 2.8 - for developers - first I completely delete the current glTF I/O scripts. Then, like before, I point to the Khronos GitHub repository.

donmccurdy:
Ok, I was able to get this working on my personal machine at least. Will write up some docs for contributors later.

donmccurdy:
Ed's documentation in https://github.com/KhronosGroup/glTF-Blender-IO/blob/master/DEBUGGING.md is great. Closing this issue.

Issue #109, opened 2018-11-29
Review of export options


emackey:
The exporter gives the user a ton of options, but I think a few of them can be simplified.

- [X] ~~There are two separate file -> export entries, one for glTF and one for GLB.  The Blender devs asked us to combine this into a single menu entry, so the choice would migrate to the export options panel.  The default would be GLB, to produce a single file for non-power users.~~

- [X] ~~Embed buffers, embed images - These shouldn't be shown for GLBs, and I'm not sure we need to offer this for text glTFs either.  If we do, perhaps there could be a combo box for "Export type" with three options [ `Binary GLB`, `glTF Embedded`, and `glTF Separate` ]~~

- [X] ~~Strip whitespace - Don't offer this.  Automatically strip whitespace for GLB, don't strip it from glTF.~~

- [X] ~~Max indicies and Force Max indicies - Let's get rid of these.  If we really need to keep them they can be hidden behind the "experimental" flag in user prefs.  But really the extension should just automatically do the right thing here (16 for smaller, 32 for larger meshes).  Artists shouldn't have to think about this stuff.~~

- [X] ~~Export Tangents, export Morph Tangents - Currently these default to True, I think we can switch them to False.  Most viewers work fine without tangents, for example the Damaged Helmet model works most places without tangents.~~

- [ ] Export all layers - See https://github.com/KhronosGroup/glTF-Blender-IO/issues/193#issuecomment-453543304

- [X] ~~Apply Modifiers - Should this default to True?  Or does that mess up skinning if true?~~ [(comment)](#issuecomment-463713414)

- [X] ~~Save settings pushpin - Let's change the pushpin to a traditional checkbox.  It could be renamed "Remember settings" with description "Store glTF export settings in the Blender project."~~

- [X] ~~Infinite perspective cameras - Let's remove this.~~

/cc @donmccurdy Thoughts on all this?

> **EDIT**: Edited by @donmccurdy to add infinite perspective cameras to the list, and update status of other items.

donmccurdy:
> Embed buffers, embed images

I like the idea of just having a dropdown. Suggested naming:

- Binary (`.glb`)
- Unpacked JSON (`.gltf`, `.bin`, + images)
- Embedded JSON (`.gltf)

[...]

PR #120, opened 2018-12-01
Remove infinite camera, whitespace options, set default tangents=false.

donmccurdy:
Per https://github.com/KhronosGroup/glTF-Blender-IO/issues/109.

Unfortunately I'm having trouble testing this; see https://github.com/KhronosGroup/glTF-Blender-IO/issues/118. Will need this to be tested before it is merged, I think.


donmccurdy:
^Note above that I have not been able to test this, due to addon installation issues on my work machine. Could someone confirm this doesn't break everything? 😇

donmccurdy:
OK, I was able to test this.

Issue #130, opened 2018-12-03
Images are converted to PNG


donmccurdy:
From glTF-Blender-Exporter, I think were we ended up was trying to use the input format wherever possible, and if the file exists on disk, try to copy it rather than re-saving (which will apply various unwanted scene preferences...).

See: https://github.com/KhronosGroup/glTF-Blender-Exporter/pull/221 and https://github.com/KhronosGroup/glTF-Blender-Exporter/pull/220.

I'm testing with this 600KB JPG, which on export becomes a 4MB PNG:

![awan biru](https://user-images.githubusercontent.com/1848368/49358184-a0e1c000-f686-11e8-8220-ff9c49f1206f.jpg)

(from https://www.blendswap.com/blends/view/80706)

viperscape:
I just wanted to note that when converted, the naming scheme in the gltf file is then not the same for uri and name:
```json
            "mimeType" : "image/png",
            "name" : "rock-mat_diffuseOriginal.bmp", // original file used
            "uri" : "rock-mat_diffuseOriginal.png" // recreated png from exporter
```

emackey:
@viperscape I think that's fine, `name` is just an arbitrary name and can match the original source material.  It doesn't need to match files shipped with the glTF.  Also, BMP format is not allowed by glTF, so it needs to be converted to either JPG or PNG.

But the up-conversion from smaller JPGs to larger PNGs is an un-wanted regression.  @donmccurdy had this fixed in the previous converter, and it would be good to see that fix preserved or re-created here.

emackey:
Thanks @UX3D-schmithuesen.  I didn't try this yet, but does your fix still convert BMPs and other disallowed formats to PNG?  Looking briefly at the code it reads as if it will preserve any original format.

donmccurdy:
Thanks! :)

PR #135, opened 2018-12-04
cleanup

UX3D-schmithuesen:


addons/io_scene_gltf2/__init__.py
```diff
@@ -74,11 +74,11 @@ class ExportGLTF2_Base:
 
     export_format = EnumProperty(
         name='Format',
-        items=(('GLB', 'Binary (.glb)', 'Exports a single file, with all data packed in binary form. Most efficient and portable, but more difficult to edit later'),
-               ('GLTF', 'JSON (.gltf)', 'Exports a single file, with all data packed in JSON. Less efficient than binary, but easier to edit later'),
-               ('GLTF_SEPARATE', 'JSON (.gltf + .bin + textures)', 'Exports multiple files, with separate JSON (.gltf), binary (.bin), and texture data. Easiest to edit later')),
+        items=(('GLB', 'glb', 'Exports a single file, with all data packed in binary form. Most efficient and portable, but more difficult to edit later'),
+               ('GLTF_EMBEDDED', 'glTF embedded)', 'Exports a single file, with all data packed in JSON. Less efficient than binary, but easier to edit later'),
```
donmccurdy: Got an extra closing parenthesis on `glTF embedded)`.

I'm not sure whether I prefer these names... We are assuming the user knows the difference already? I can see where the term "JSON" is not something an artist wants to know about, but for example the FBX exporter has "Binary" and "ASCII" options similar to this. And a glb is still a glTF file...

/cc @emackey any preferences on this?

emackey: You've got a trailing close paren in `glTF embedded)`

addons/io_scene_gltf2/__init__.py
```diff
@@ -74,11 +74,11 @@ class ExportGLTF2_Base:
...
+               ('GLTF', 'glTF', 'Exports multiple files, with separate JSON (.gltf), binary (.bin), and texture data. Easiest to edit later')),
         description='Output format and embedding options. Binary is most efficient, but JSON (embedded or separate) may be easier to edit later',
-        default='GLB'
+        default='GLTF'
```
donmccurdy: I would prefer that the default be GLB — that's the most efficient representation, and it's the only form that works in certain environments like Facebook 3D Posts.

emackey: Are we sure we want this default?  I would think typical users are more likely to be sharing GLBs.  I'm fine with either though.

addons/io_scene_gltf2/blender/com/gltf2_blender_image_util.py
```diff
@@ -1,120 +0,0 @@
...
-
-    def png_pack(png_tag, data):
-        chunk_head = png_tag + data
-        return struct.pack("!I", len(data)) + chunk_head + struct.pack("!I", 0xFFFFFFFF & zlib.crc32(chunk_head))
-
-    return b"".join([
-        b'\x89PNG\r\n\x1a\n',
-        png_pack(b'IHDR', struct.pack("!2I5B", width, height, 8, 6, 0, 0, 0)),
-        png_pack(b'IDAT', zlib.compress(raw_data, 9)),
-        png_pack(b'IEND', b'')])
```
donmccurdy: Note that some of this code was used before to solve the regression we now have in https://github.com/KhronosGroup/glTF-Blender-IO/issues/130.

addons/io_scene_gltf2/blender/exp/gltf2_blender_animate.py
```diff
@@ -1,637 +0,0 @@
...
-#
-
-JOINT_NODE = 'JOINT'
-
-NEEDS_CONVERSION = 'CONVERSION_NEEDED'
-CUBIC_INTERPOLATION = 'CUBICSPLINE'
-LINEAR_INTERPOLATION = 'LINEAR'
-STEP_INTERPOLATION = 'STEP'
-BEZIER_INTERPOLATION = 'BEZIER'
-CONSTANT_INTERPOLATION = 'CONSTANT'
```
donmccurdy: Do we still support CUBICSPLINE here? The code remaining seems much simpler than before, if that's still working as intended that's awesome. 👍

donmccurdy:
Heh, didn't see your comments before adding my own @emackey :)

addons/io_scene_gltf2/__init__.py
donmccurdy: I'm happy with that, thanks.

addons/io_scene_gltf2/blender/com/gltf2_blender_image_util.py
UX3D-schmithuesen: Since I'm assigned to that one anyway, I might as well fix that first and come back to this then.

addons/io_scene_gltf2/blender/exp/gltf2_blender_animate.py
UX3D-schmithuesen: It's mentioned multiple times in the new animation code so I do believe it.

donmccurdy:
Updates in 6ac4658da88a12a83243acccdaeba5f90af171f6 look good, thanks!

addons/io_scene_gltf2/blender/exp/gltf2_blender_animate.py
donmccurdy: Great! Yes, tested and aside from one issue with custom spline tangents (https://github.com/KhronosGroup/glTF-Blender-IO/issues/140) it seems to be working as before. :)

Issue #136, opened 2018-12-04
Export Custom Properties for Scene


EJTH:
Custom properties for scenes are not exported. I think that would be nice to have.

EJTH:
#138 I made a pull request, since the change is so simple even I can do it. :-)

EJTH:
Maybe `glTF2ExportSettings` should be excluded from the extras generated?

PR #138, opened 2018-12-04
Export Custom Properties for Scene

EJTH:
#136 Export Custom Properties for Scene

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather.py
@@ -62,6 +62,6 @@ def __gather_animations(blender_scene, export_settings):
 
 
 def __gather_extras(blender_object, export_settings):
-    if export_settings['gltf_extras']:
+    if export_settings[gltf2_blender_export_keys.EXTRAS]:
         return gltf2_blender_generate_extras.generate_extras(blender_object)
     return None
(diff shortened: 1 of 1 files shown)


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather.py
```diff
@@ -58,3 +59,9 @@ def __gather_animations(blender_scene, export_settings):
     for blender_object in blender_scene.objects:
         animations += gltf2_blender_gather_animations.gather_animations(blender_object, export_settings)
     return animations
+
+
+def __gather_extras(blender_object, export_settings):
+    if export_settings['gltf_extras']:
```
UX3D-schmithuesen: Please don't use a literal string here.
```suggestion
    if export_settings[gltf2_blender_export_keys.EXTRAS]:
```

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather.py
```diff
@@ -18,6 +18,7 @@
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_nodes
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_animations
 from io_scene_gltf2.blender.exp.gltf2_blender_gather_cache import cached
+from io_scene_gltf2.blender.exp import gltf2_blender_generate_extras
```
UX3D-schmithuesen: add this import for my suggestion below
```suggestion
from io_scene_gltf2.blender.exp import gltf2_blender_export_keys

```

Issue #146, opened 2018-12-05
Saved export format setting seems broken


emackey:
In a blend file I made to test something unrelated, I ended up with this as a custom scene property:

```
glTF2ExportSettings

{'export_format': 1, 'export_selected': 1}
```

With that in place, if I file -> export glTF, I see this error before I even get to the export settings UI.

```
Traceback (most recent call last):
  File "..\2.80\scripts\addons\io_scene_gltf2\__init__.py", line 272, in invoke
    setattr(self, k, v)
  File "..\2.80\scripts\modules\bpy_types.py", line 662, in __setattr__
    return setattr(properties, attr, value)
TypeError: bpy_struct: item.attr = val: expected a string enum, not int
```

/cc @donmccurdy

UX3D-becher:
Does this also happen in blender 2.79?
At first glance this seems to be a bug in blender

willstott101:
I've also stumbled into this issue. I didn't really know hoe to debug it though and ended up just making the patch at https://github.com/KhronosGroup/glTF-Blender-IO/pull/149

EJTH:
Hi! I made a fix that seems to work. The problem is that EnumProperties are not stored/read properly depending on what was intended.

emackey:
Further discussion in #148.

PR #149, opened 2018-12-06
Catch wrong Property type when loading settings.

willstott101:
I'm not sure how it happened but I think a setting saved in my blender file was of the incorrect type for it's associated Property. I had to make this change to be able to export again.

It seems fairly inconspicuous to me, but I would understand if you'd refuse the patch without a test case or example blend file (as mine no longer has the culprit setting saved).

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -272,7 +272,7 @@ class ExportGLTF2_Base:
                     setattr(self, k, v)
                 self.will_save_settings = True
 
-            except AttributeError:
+            except (AttributeError, TypeError):
                 self.report({"ERROR"}, "Loading export settings failed. Removed corrupted settings")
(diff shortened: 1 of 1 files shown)


emackey:
Related to #146?  Seems like this would "fix" it by throwing away the saved settings.  I would rather allow the settings to load correctly if that can be worked out.  I think this just sweeps a live bug under the rug.

willstott101:
Yeah you're right, it's definitely just a workaround.

emackey:
@willstott101 No worries.  I hope you get the chance to send us another PR sometime!

emackey:
So actually, now that the root cause of the bug is fixed in #150, there are still some `.blend` files hanging around that have the old corrupted settings still baked into them.  Re-opening and merging this would fix those files by discarding the corrupt settings.

Any objections from UX3D or others?

donmccurdy:
+1 from me

PR #150, opened 2018-12-06
Exporter: Save Settings Broken

EJTH:
Fix for T58830 (#146) Exporter Save Settings Broken. Im not into python really so Im not sure this is the right way, but it works for me.


emackey:
@EJTH Thanks so much for this!

Is there an easy way to make this tolerant of the malformed saved export settings from the last few days?  Or should we just tell people to manually delete them?

If needed we could even pull in @willstott101's abandoned patch from #149, which would discard the bad export settings, and allow a fresh export to begin.

PR #173, opened 2018-12-16
Implementation of punctual lights

julienduroure:


UX3D-nopper:
Have you cross checked, if your solution also works, when the light is animated?
Actually, the same problem occurs with the camera.

julienduroure:
@UX3D-nopper I added support for light animation (tested with light animation exported from blender, and imported back)

I checked for camera, but it seems to work (tested with camera animation exported from blender, and imported back)

julienduroure:
@UX3D-schmithuesen If possible, I would like to manage myself PR merging / stuff regarding importer part of the addon

UX3D-schmithuesen:
Okay, sure! So you just want some comments when you create a PR?

PR #195, opened 2018-12-28
Ignore non-deforming vertex weights when exporting from Blender

WindyDarian:
* Fixed an issue where weight was added to joint 0 when the bone doesn't exist in armature (the case for non-deforming vertex groups)
* Don't add joint weight to vertex if weight is zero

Was trying a Blender 2.80->gltf2->Godot workflow, but some vertices behaved funny in Godot. This fixed the look in Godot


donmccurdy:
Thanks!

PR #201, opened 2019-01-01
Use unsigned short or unsigned int for indices

klacansky:


UX3D-schmithuesen:
Any reason why we would do this?

lexaknyazev:
Probably this one: https://github.com/KhronosGroup/glTF/issues/1471.

PR #206, opened 2019-01-04
Re-apply the fix for primitive restart values.

### LINKED ISSUE #204: Indices accessor contains primitive restart value (65535)
Managed to create a file that throws this validation error: 

> Indices accessor contains primitive restart value (65535) at index 78974.

Maybe off-by-one on a check somewhere?

[Archive.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/2726218/Archive.zip)

emackey:
Fixes #204.


lexaknyazev:
Could the error message update from #56 be re-applied here as well?

emackey:
@lexaknyazev Thanks for catching that.  Done.

PR #223, opened 2019-01-09
Feature: export base color factor

### LINKED ISSUE #191: BaseColorFactor not exported
Hey,

I was playing around with this exporter / importer for a while, and noticed the following things:

- BaseColor factor is imported only at half strength. I was about to open up a new issue, and even fix it with a PR, when I found this, so I guess it's already fixed now: https://github.com/KhronosGroup/glTF-Blender-IO/issues/187
- However, when I export the same model, it loses information: the BaseColorFactor is not included in the export.

So I assume, right now there's no way to have a texture AND a basecolorfactor on the model simultaneously (at least, the colorfactor part won't get exported).

UX3D-schmithuesen:
fixes #191 
Only multiply nodes with a value set directly on it are supported (see screenshot from #191).


Ali-RS:
@UX3D-schmithuesen it seems this PR not solves this issue : https://github.com/KhronosGroup/glTF-Blender-IO/issues/393

PR #229, opened 2019-01-12
Importer: Manage KHR_materials_unlit

julienduroure:
Note that Background Node is not working for objects on 2.79.
So for now, I replaced Background node by emission node, even if this is wrong.

One solution for 2.79 is to create a tree with Light Path 'Is camera ray'.
Other solution is to keep it wrong in 2.79, with a warning


julienduroure:
I know the shadeless checkbox for Blender Internal Render, but I can't find it for Cycles

julienduroure:
Some feedback before I merge it?

PR #242, opened 2019-01-21
Fixed exporting non-number array custom properties

Capital-Asterisk:
So far, only number array custom properties can be exported as extras. An array of directories or strings would not show up in the gltf file at all.

Try setting a custom property like this, and it will not show up in the extras category.
`bpy.data.objects["Cube"]["mixed array"] = [{"foo": "bar"}, "stringy"]`

However, this works just fine.
`bpy.data.objects["Cube"]["array of 42"] = [42, 42, 42, 42]`

One of the projects I'm working on relies heavily on custom properties that are arrays of dictionaries. I've done some changes on gltf2_blender_generate_extras.py to fix this.

 * Moved type checks for custom properties into its own function
 * Added a check for list classes to allow arrays of strings and dictionaries
 * Replaced some if statements with elif

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_generate_extras.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_generate_extras.py
@@ -16,6 +16,37 @@
 import bpy
 from io_scene_gltf2.blender.com import gltf2_blender_json
 
+def to_json_compatible(value):
+    """Make a value (usually a custom property) compatible with json"""
+
+    if isinstance(value, bpy.types.ID):
+        return value
+
+    elif isinstance(value, str):
+        return value
+
+    elif isinstance(value, (int, float)):
+        return value
+
+    # for list classes
+    elif isinstance(value, list):
+        value = list(value)
+        # make sure contents are json-compatible too
+        for index in range(len(value)):
+            value[index] = to_json_compatible(value[index])
+        return value
+
+    # for IDPropertyArray classes
+    elif hasattr(value, "to_list"):
+        value = value.to_list()
+        return value
+
+    elif hasattr(value, "to_dict"):
+        value = value.to_dict()
+        if gltf2_blender_json.is_json_convertible(value):
+            return value;
+        
+    return None
 
 def generate_extras(blender_element):
     """Filter and create a custom property, which is stored in the glTF extra field."""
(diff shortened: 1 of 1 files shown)


addons/io_scene_gltf2/blender/exp/gltf2_blender_generate_extras.py
```diff
@@ -16,6 +16,37 @@
...
+
+    elif isinstance(value, str):
+        return value
+
+    elif isinstance(value, (int, float)):
+        return value
+
+    # for list classes
+    elif isinstance(value, list):
+        value = list(value)
```
UX3D-schmithuesen: is this really necessary if it's already of type `list`?

addons/io_scene_gltf2/blender/exp/gltf2_blender_generate_extras.py
```diff
@@ -16,6 +16,37 @@
 import bpy
 from io_scene_gltf2.blender.com import gltf2_blender_json
 
+def to_json_compatible(value):
```
UX3D-schmithuesen: please make this private (rename to `__to_json_compatible`) and move it below `generate_extras`

addons/io_scene_gltf2/blender/exp/gltf2_blender_generate_extras.py
```diff
@@ -16,6 +16,37 @@
...
+
+    # for IDPropertyArray classes
+    elif hasattr(value, "to_list"):
+        value = value.to_list()
+        return value
+
+    elif hasattr(value, "to_dict"):
+        value = value.to_dict()
+        if gltf2_blender_json.is_json_convertible(value):
+            return value;
```
UX3D-schmithuesen: for [PEP8](https://www.python.org/dev/peps/pep-0008/) conformity, please remove the `;`

addons/io_scene_gltf2/blender/exp/gltf2_blender_generate_extras.py
```diff
@@ -16,6 +16,37 @@
 import bpy
 from io_scene_gltf2.blender.com import gltf2_blender_json
```
UX3D-schmithuesen: for [PEP8](https://www.python.org/dev/peps/pep-0008/) conformity, there have to be 2 empty lines between imports and the first `def`

addons/io_scene_gltf2/blender/exp/gltf2_blender_generate_extras.py
Capital-Asterisk: It's suppose to clone the list to make sure the original isn't modified. Incompatible custom properties would be set to None in the upcoming loop.

UX3D-schmithuesen: okay, that's fine then 👍

UX3D-schmithuesen:
Since it's only some small formalities, I'll merge it and do the changes myself :+1:

PR #250, opened 2019-01-24
Add normal strength/scale to glTF export.

### LINKED ISSUE #241: Normal scale/strength
Apologies if this is a duplicate, I couldn't find exactly this issue but I know I've run into it before.

glTF has a parameter [`normalTextureInfo.scale`](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#normaltextureinfoscale) that controls the strength of the normal-mapping.  Blender has a parameter called "Strength" in its Normal Map node that appears to control the same thing:

normal strength

Looks like neither the importer or exporter copy this value out or back into glTF files.  Can/should this be added to both?

emackey:
Fixes #241.

/cc @donmccurdy


donmccurdy:
Looks good!

Issue #76, opened 2018-11-15
Importer performance: use foreach_set / foreach_get


julienduroure:

julienduroure:
Foreach_get & foreach_set can improve performance.
Example : 

```
obj = bpy.context.object
len = len(obj.data.vertices)
coords = [0] * len * 3
obj.data.vertices.foreach_get('co', coords)
```

scurest:
Yeah, using `keyframe_insert` for animations is craaaazy slow. Here's an example of how to make it _much_ faster

````diff
diff --git a/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation_node.py b/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation_node.py
index b3e8291..2e6cd61 100644
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation_node.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation_node.py
@@ -76,61 +76,54 @@ class BlenderNodeAnim():
                 # We can't remove Yup2Zup oject
                 gltf.animation_object = True
 
+                if animation.samplers[channel.sampler].interpolation == "CUBICSPLINE":
+                    # TODO manage tangent?
+                    values = [values[idx * 3 + 1] for idx in range(0, len(keys))]
+
                 if channel.target.path == "translation":
                     blender_path = "location"
-                    for idx, key in enumerate(keys):
-                        if animation.samplers[channel.sampler].interpolation == "CUBICSPLINE":
-                            # TODO manage tangent?
-                            obj.location = Vector(loc_gltf_to_blender(list(values[idx * 3 + 1])))
-                        else:
-                            obj.location = Vector(loc_gltf_to_blender(list(values[idx])))
-                        obj.keyframe_insert(blender_path, frame=key[0] * fps, group='location')
-
-                    # Setting interpolation
-                    for fcurve in [curve for curve in obj.animation_data.action.fcurves
-                                   if curve.group.name == "location"]:
-                        for kf in fcurve.keyframe_points:
-                            BlenderNodeAnim.set_interpolation(animation.samplers[channel.sampler].interpolation, kf)
+                    group = action.groups.new("location")
+                    num_components = 3
+                    values = [loc_gltf_to_blender(vals) for vals in values]
 
                 elif channel.target.path == "rotation":
                     blender_path = "rotation_quaternion"
-                    for idx, key in enumerate(keys):
[...]

PR #253, opened 2019-01-25
Much faster animation import

scurest:
[Persuant to #76](https://github.com/KhronosGroup/glTF-Blender-IO/issues/76#issuecomment-456704838), I changed the importer for node TRS and bone TRS curves to fill in Blender's FCurves in batches. This gives enormous speedups for models with lots of animation data. To measure this I added the import time to the "glTF import finished" log statement and tried the slowest sample models, plus the model I've been testing with.

|Model|Before|After|% Improvement|
|---|---|---|---|
|Brainstem | 63s | 26s | 142%|
|VC | 23s | 4s | 475%|
|[b00A](https://github.com/KhronosGroup/glTF-Blender-IO/files/2794270/animation_test.zip) | 7.2s | 0.5s | 1340%|

I didn't change the importer for morph weight curves, but you can do something similar to that one too.

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -17,6 +17,7 @@
 #
 
 import os
+import time
 import bpy
 from bpy_extras.io_utils import ImportHelper, ExportHelper
 from bpy.types import Operator, AddonPreferences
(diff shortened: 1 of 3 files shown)


julienduroure:
Thanks!
You did it before I get the time to do it myself :)

Looks good to me, but want to have a deeper look/tests before merging

PR #255, opened 2019-01-25
Reconnect the double-sided flag from the old custom node system.

emackey:
Most exporter inputs still have fallback behavior to look for the old custom node system, but DoubleSided was missing this hookup.  Talked offline with @donmccurdy and there still isn't a great way to get this flag from within Blender, but, it's worth maintaining compatibility with the old system here.


emackey:
Tests pass now.  This is ready.

PR #258, opened 2019-01-27
Added support for explicitly setting the glTF "doubleSided" material …

### LINKED ISSUE #154: Checking `Double Sided` doesn't export as double-sided geometry
Checking the `Double Sided` checkbox under `Object Data` doesn't export a double-sided geometry.

image

Is this the preferred method for doing so, or is there another way?

image

atteneder:
…property.

This is done via custom Blender material property. Works for import and export and adds a UI Panel in the material property tab for easy (un-)checking.

fixes #154

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -491,6 +491,21 @@ class ImportGLTF2(Operator, ImportHelper):
 
         return {'FINISHED'}
 
+class GltfMaterialPanel(bpy.types.Panel):
+    bl_label = "glTF"
+    bl_idname = "MATERIAL_PT_glTF"
+    bl_space_type = 'PROPERTIES'
+    bl_region_type = 'WINDOW'
+    bl_context = "material"
+
+    @classmethod
+    def poll(self, context):
+        return context.active_object is not None and context.active_object.active_material is not None
+
+    def draw(self, context):
+        layout = self.layout
+        mat = context.active_object.active_material
+        layout.prop(mat, "gltf_double_sided")
 
 def menu_func_import(self, context):
     self.layout.operator(ImportGLTF2.bl_idname, text=ImportGLTF2.bl_label)
@@ -498,7 +513,8 @@ def menu_func_import(self, context):
 
 classes = (
     ExportGLTF2,
-    ImportGLTF2
+    ImportGLTF2,
+    GltfMaterialPanel,
 )
 
 
@@ -515,6 +531,7 @@ def register():
         bpy.types.TOPBAR_MT_file_export.append(menu_func_export)
         bpy.types.TOPBAR_MT_file_import.append(menu_func_import)
 
+    bpy.types.Material.gltf_double_sided = BoolProperty( name="glTF double sided", description="When this value is true, back-face culling is disabled and double sided lighting is enabled in exported glTF files" )
 
 def unregister():
     for c in classes:
@@ -528,3 +545,5 @@ def unregister():
     else:
         bpy.types.TOPBAR_MT_file_export.remove(menu_func_export)
         bpy.types.TOPBAR_MT_file_import.remove(menu_func_import)
+
+    del bpy.types.Material.gltf_double_sided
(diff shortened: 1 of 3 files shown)


mholthausen:
I think this conflict is related to what I expected, when I have tested the magnificent add-on: the assigned material (e.g. diffuse color) isn't exported into the glTF/glb-File, although the "Export Materials" is checked in the export menu.

atteneder:
> I think this conflict is related to what I expected, when I have tested the magnificent add-on: the assigned material (e.g. diffuse color) isn't exported into the glTF/glb-File, although the "Export Materials" is checked in the export menu.

I doubt that this is related to your problem since I cannot reproduce. Is the error happening on master as well? If not, could you provide me with your setup, steps and test data?

mholthausen:
> > I think this conflict is related to what I expected, when I have tested the magnificent add-on: the assigned material (e.g. diffuse color) isn't exported into the glTF/glb-File, although the "Export Materials" is checked in the export menu.
> 
> I doubt that this is related to your problem since I cannot reproduce. Is the error happening on master as well? If not, could you provide me with your setup, steps and test data?

It's the current Blender 2.80 Beta downloaded ob blender.org
Date: 2019-01-27 19:17
Hash: 690478027bd7
Branch: blender2.7 

Running on Linux Mint 19 

The worklfow consists of 
1) Importing a shapefile into Blender with the BlenderGIS-Plugin (https://github.com/domlysz/BlenderGIS/tree/Blender28) as mesh where the normals of some triangles are up and some others are down

2) Assigning a material and color to the mesh and checking the "glTF double sided" of the box in the "Material" pane

3) Checking "Double Sided" in Normals section of "Object Data" pane

4) Exporting the scene as glb

5) Uploading the scene into the glTF-Viewer (https://gltf-viewer.donmccurdy.com/) and the assigned color is missing

Behavoiur occurs in master of gltF-Blender-IO, too.
In BLender 2.79 with the current master of glTF-Blender-Exporter works like expected (without double sided, of course)

So, i guess it's a miscommunication between Blender 2.8 Beta and glTF-Blender-IO for 2.8

atteneder:
> Behavoiur occurs in master of gltF-Blender-IO, too.

Thanks. In that case, you should create a proper issue with a copy of your description. It's not related to this PR at all, as far as I can tell.

donmccurdy:
Note that https://github.com/KhronosGroup/glTF-Blender-IO/pull/255 was just merged, allowing `doubleSided` to be exported from the old custom nodes. We (intentionally) do not create those old nodes when importing glTF models, but at least it provides an escape hatch for exporting the property.

If I'm reading correctly, this option appears in the main Blender materials panel, not the materials panel at the export step? Back when glTF-Blender-IO required an opt-in installation by users, I think this fix would have been OK. However, as the addon comes enabled by default now, we're essentially adding new UI for all Blender users by doing this. As Blender already has (multiple) similar options, I'm not sure we can justify it.

A more robust fix, perhaps, would be to prescan all Blender meshes for a double sided flag, and then set `material.doubleSided` for materials on those meshes. A material may be reused by a mesh that isn't double sided, and ideally that would be handled too, but some false positives/negatives may be OK for an initial fix.

atteneder:
> Note that #255 was just merged, allowing `doubleSided` to be exported from the old custom nodes. We (intentionally) do not create those old nodes when importing glTF models, but at least it provides an escape hatch for exporting the property.
I noticed, but I too don't want to use the obsolete custom nodes.

> If I'm reading correctly, this option appears in the main Blender materials panel, not the materials panel at the export step? Back when glTF-Blender-IO required an opt-in installation by users, I think this fix would have been OK. However, as the addon comes enabled by default now, we're essentially adding new UI for all Blender users by doing this. As Blender already has (multiple) similar options, I'm not sure we can justify it.
I'm not sure either :)
> A more robust fix, perhaps, would be to prescan all Blender meshes for a double sided flag, and then set `material.doubleSided` for materials on those meshes. A material may be reused by a mesh that isn't double sided, and ideally that would be handled too, but some false positives/negatives may be OK for an initial fix.
Good idea. I thought about it briefly, but I wanted a quick solution without digging too much into the addon's code.

emackey:
Closing this in favor of the solution in #304.

@atteneder Sorry we couldn't take this one.  There are a number of open issues remaining in this project, so we'd be glad to see more PRs with fixes, we appreciate them!

PR #267, opened 2019-01-30
Bring back save settings option for exporter

pjoe:
- Use suggested name and description from #109

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -255,7 +255,10 @@ class ExportGLTF2_Base:
         default=False
     )
 
-    will_save_settings = BoolProperty(default=False)
+    will_save_settings = BoolProperty(
+        name='Remember Settings',
+        description='Store glTF export settings in the Blender project',
+        default=False)
 
     # Custom scene property for saving settings
     scene_key = "glTF2ExportSettings"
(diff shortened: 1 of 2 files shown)


addons/io_scene_gltf2/__init__.py
```diff
@@ -255,7 +255,10 @@ class ExportGLTF2_Base:
         default=False
     )
 
-    will_save_settings = BoolProperty(default=False)
+    will_save_settings = BoolProperty(
+        name='Remember Settings',
+        description='Store glTF export settings in the Blender project',
```
donmccurdy: maybe, `Remember Export Settings`? /cc @emackey

donmccurdy:
I don't have a strong preference on this... it hasn't been something I've needed, but I can see where others might. I have some reservations about writing extra data into the Blender project when other exporters don't, as it has caused issues in the past with "breaking" a file if exporter settings change.

emackey:
> maybe, `Remember Export Settings`

I'd be fine with the name changing to that.  I think the description is good as-is.

> it has caused issues in the past with "breaking" a file if exporter settings change

I just tested that, it drops the settings but doesn't crash.  At some point I think we added a try/catch or something to protect against those crashes.

addons/io_scene_gltf2/__init__.py
pjoe: Fixed

emackey:
Any objections?  @julienduroure @UX3D-nopper

pjoe:
I added a quick explanation in io_gltf2.rst

PR #281, opened 2019-02-06
Disable armature modifier when applying modifiers and exporting skins

pjoe:
Makes things work when you want to non-destructively apply other modifiers (e.g decimate) and still be able to export skinning and animation

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
@@ -177,11 +177,24 @@ def __gather_mesh(blender_object, export_settings):
             edge_split.split_angle = blender_object.data.auto_smooth_angle
             edge_split.use_edge_angle = not blender_object.data.has_custom_normals
 
+        armature_modifiers = {}
+        if export_settings[gltf2_blender_export_keys.SKINS]:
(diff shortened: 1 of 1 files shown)


donmccurdy:
Thanks! The FBX exporter does the same thing; +1 from me on ignoring armature modifiers.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
```diff
@@ -177,12 +177,25 @@ def __gather_mesh(blender_object, export_settings):
...
+
         if bpy.app.version < (2, 80, 0):
             blender_mesh = blender_object.to_mesh(bpy.context.scene, True, 'PREVIEW')
         else:
             blender_mesh = blender_object.to_mesh(bpy.context.depsgraph, True)
         skip_filter = True
 
+        if export_settings[gltf2_blender_export_keys.SKINS]:
+            # restore Armature modifiers
+            for key, val in armature_modifiers.items():
```
donmccurdy: nit: let's use the `idx, modifier` variable names to match the block above.

pjoe: I'll settle for `idx, show_viewport` as that is really what's in the dict

donmccurdy:
Thanks! :)

PR #283, opened 2019-02-07
Auto limit and normalize skin weights

pjoe:
Limit to 4 joints and normalize when 'all vertex influences' export setting is off

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitive_attributes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitive_attributes.py
@@ -165,10 +165,10 @@ def __gather_skins(blender_primitive, export_settings):
         joint_id = 'JOINTS_' + str(bone_index)
         weight_id = 'WEIGHTS_' + str(bone_index)
         while blender_primitive["attributes"].get(joint_id) and blender_primitive["attributes"].get(weight_id):
-            if bone_index >= 4:
-                gltf2_io_debug.print_console("WARNING", "There are more than 4 joint vertex influences."
-                                                        "Consider to apply blenders Limit Total function.")
+            if bone_index >= 1:
                 if not export_settings['gltf_all_vertex_influences']:
+                    gltf2_io_debug.print_console("WARNING", "There are more than 4 joint vertex influences."
+                                                            "The 4 with highest weight will be used (and normalized).")
                     break
 
             # joints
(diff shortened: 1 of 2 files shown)


pjoe:
CI: https://circleci.com/gh/p3d-in/glTF-Blender-IO/31

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitive_attributes.py
```diff
@@ -165,10 +165,10 @@ def __gather_skins(blender_primitive, export_settings):
         joint_id = 'JOINTS_' + str(bone_index)
         weight_id = 'WEIGHTS_' + str(bone_index)
         while blender_primitive["attributes"].get(joint_id) and blender_primitive["attributes"].get(weight_id):
-            if bone_index >= 4:
-                gltf2_io_debug.print_console("WARNING", "There are more than 4 joint vertex influences."
-                                                        "Consider to apply blenders Limit Total function.")
+            if bone_index >= 1:
```
donmccurdy: Why `>= 1` here?

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitive_attributes.py
```diff
@@ -192,6 +192,15 @@ def __gather_skins(blender_primitive, export_settings):
 
             # weights
             internal_weight = blender_primitive["attributes"][weight_id]
+            # normalize first 4 weights, when not exporting all influences
+            if not export_settings['gltf_all_vertex_influences']:
+                for idx in range(0, len(internal_weight), 4):
```
donmccurdy: This is normalizing weights 1-4, 5-8, 9-12, each group in batches of four, correct? That seems a little confusing, doesn't look like we should need this outer loop.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitive_attributes.py
pjoe: Heh, was pretty confused about why the 4 wasn't working, however bone_index here refers to the output groups in gltf (JOINTS_X, WEIGHTS_X) which each hold 4 bones.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitive_attributes.py
pjoe: Yep, it needs to normalize each bundle of 4. Not sure how you would do that without the outer loop in line 197

donmccurdy: Oh I see, `internal_weights` is the weights for all vertices, and each "bundle of 4" is one vertex? I think I'd misread and assumed this was normalizing WEIGHTS_0, then normalizing WEIGHTS_1, etc.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitive_attributes.py
donmccurdy: Oh that's confusing. 😅 Do you mind renaming this variable to `group_index` or `set_index`?

emackey:
@donmccurdy I haven't dug much into bones & skinning, so this one is for you & Norbert.

donmccurdy:
This is working correctly for me. Thanks @pjoe!

PR #287, opened 2019-02-08
Correct handling of objects directly parented to bones, fix #286

### LINKED ISSUE #286: Meshes parented to a bone is exported as parented to the bone's armature
If you have a mesh (or other object I guess) that in blender is parented to a bone directly (i.e. not skinned with armature modifier), then it gets exported as though it was parented to the armature instead.

I'm trying to have simple mesh hit-boxes parented to bones of a character, but this doesn't work :(

I can't skin the hit boxes as the ray caster I'm using won't work on skinned meshes.

pjoe:
A bit tricky, but looks to be working correctly now :)

Fixes #286

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
@@ -99,6 +99,9 @@ def __gather_children(blender_object, export_settings):
     # standard children
     for child_object in blender_object.children:
         if child_object.parent_bone:
+            # this is handled further down,
+            # as the object should be a child of the specific bone,
+            # not the Armature object
             continue
         node = gather_node(child_object, export_settings)
         if node is not None:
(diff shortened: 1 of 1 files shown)


pjoe:
Hold off merging this ... discovered it doesn't work in all cases. Trying to fix :)

pjoe:
Fixed now, ready for merge again :)

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
```diff
@@ -114,9 +116,47 @@ def __gather_children(blender_object, export_settings):
...
+                parent_joint = find_parent_joint(joint.children, name)
+                if parent_joint:
+                    return parent_joint
+            return None
+        for child in direct_bone_children:
+            # find parent joint
+            parent_joint = find_parent_joint(root_joints, child.parent_bone)
+            if parent_joint:
+                child_node = gather_node(child, export_settings)
+                if child_node is not None:
```
donmccurdy: nit: could you change the `if parent_joint:` and `if child_node is not None:` checks to use a "guard clause" pattern, avoiding the extra nesting? For example:

```python
if not parent_joint:
  continue
# ...
```

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
```diff
@@ -95,6 +95,8 @@ def __gather_children(blender_object, export_settings):
     children = []
     # standard children
     for child_object in blender_object.children:
+        if child_object.parent_bone:
+            continue
```
donmccurdy: Huh, so these objects are both parented to the armature and the bone, in blender's representation? That's surprising, a comment in the code here might be helpful.

pjoe: Yeah it's a bit confusing ... it's because in blender bones are apparently not objects like everything else, but special 'sub objects' of the armature object

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
pjoe: done

donmccurdy:
Testing against this model...

 https://stackoverflow.com/questions/54162120/blender-how-to-correctly-export-gltf-glb-with-armature-applied

... we're close, but some at least one node is rotated 180º —

screen shot 2019-02-08 at 1 21 31 pm

Maybe the rotation should be +90 rather than -90? Does that work for your case as well?

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
```diff
@@ -114,9 +119,48 @@ def __gather_children(blender_object, export_settings):
...
+                if parent_joint:
+                    return parent_joint
+            return None
+        for child in direct_bone_children:
+            # find parent joint
+            parent_joint = find_parent_joint(root_joints, child.parent_bone)
+            if not parent_joint:
+                continue
+            child_node = gather_node(child, export_settings)
+            if child_node is not None:
```
donmccurdy: let's do a guard clause here as well, thanks!

pjoe:
Must admit I came to the -90 by trial and error ... +90 did not work for me. I suspect it may be something with the swizzling/y-up handling that happens multiple times in the hierarchy, but not really sure.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
pjoe: done

donmccurdy:
Could you share an example `.blend` where `-90` is working correctly?

pjoe:
Unfortunately the `.blend` I've been working on is for a commercial project, so can't share here :(

I'm testing with the lamp now and +90 doesn't work either .. the shade gets corrected, but the base gets upside down ... hmmm

pjoe:
FWIW: the lamp exports correctly - without y-up conversion.

pjoe:
Finally figured this out! After several days of nightmares about matrix math in general, parent inverse, quaternions, bone roll and architecture lamps :S

I'm attaching another test file (2.80)
[bone_children.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/2966722/bone_children.zip)

And some screens:
image

image

image

donmccurdy:
Many thanks @pjoe, and sorry for the lamp-themed nightmares! 😅 The lamp and the new example you provide both look correct to me.

For a more fun example, I've have a version of [this robot](https://www.patreon.com/posts/free-animated-22105894) which I've loaded into Blender and added facial expressions:

[RobotExpressive.blend 2.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/2967293/RobotExpressive.blend.2.zip)

It's never worked with this exporter or the older exporter (which didn't support bone parenting at all). With your fix, it's looking _much_ closer than before, and I think the remaining issue – note the fingers – is probably outside the scope of this PR:

Screen Shot 2019-03-14 at 8 57 19 AM

emackey:
@donmccurdy Can you do the honors and merge this?  I'm not a bones/skeleton user.

donmccurdy:
Thanks again @pjoe! :)

pjoe:
Wohoo, merged !

@donmccurdy when I get a spare moment, I'll try taking a look at the skinning weirdness on the robot hand

PR #288, opened 2019-02-08
accessibility through search bar

Vinc3r:
not such a big thing but could make search easier (plus it acts like other export file formats)

img


donmccurdy:
Does this not affect the File -> Export menu too? 🤔

Vinc3r:
Since I don't modified line 435 (and also after tested of course), it should not.

```PYTHON
def menu_func_export(self, context):
   self.layout.operator(ExportGLTF2.bl_idname, text='glTF 2.0 (.glb/.gltf)')
```

donmccurdy:
Thanks!

donmccurdy:
Good catch! Also tweaked this to remove the extensions, better matching the rest of the list:

screen shot 2019-02-08 at 12 53 40 pm

PR #291, opened 2019-02-10
Exporter: correctly convert UV texture transforms

scurest:
Previous to this change, the values in the Mapping node in Blender were copied straight into the KHR_texture_transform object. This is incorrect because the Mapping node is applied in Blender's UV space, which is related to glTF's with the u,v->u,1-v conversion. That meant that a Mapping node that eg. translates up would be exported as a KHR_texture_transform that translates down because of the
differing directions of the V axis.

The correct thing to do is to calculate the equivalent transform in the other UV space.

This was added to the common module because the importer would need to apply the same conversion to get the right Mapping node from a KHR_texture_transform.

### DIFF
--- a/addons/io_scene_gltf2/blender/com/gltf2_blender_conversion.py
+++ b/addons/io_scene_gltf2/blender/com/gltf2_blender_conversion.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from mathutils import Matrix, Quaternion
-from math import sqrt
+from math import sqrt, sin, cos
 
 def matrix_gltf_to_blender(mat_input):
     """Matrix from glTF format to Blender format."""
@@ -46,3 +46,22 @@ def correction_rotation():
     # Correction is needed for lamps, because Yup2Zup is not written in vertices
     # and lamps has no vertices :)
     return Quaternion((sqrt(2)/2, -sqrt(2)/2, 0.0, 0.0)).to_matrix().to_4x4()
+
+def convert_texture_transform(texture_transform):
+    """
+    Converts a KHR_texture_transform object in one UV space (glTF or Blender)
+    into the equivalent in the other UV space. The returned transform is the
+    same as switching UV spaces (with u,v -> u,1-v), applying texture_transform,
+    then switching back.
+    """
+    offset = texture_transform.get('offset', [0, 0])
+    rotation = texture_transform.get('rotation', 0)
+    scale = texture_transform.get('scale', [1, 1])
+    return {
+        'offset': [
+            offset[0] - scale[1] * sin(rotation),
+            1 - offset[1] - scale[1] * cos(rotation),
+        ],
+        'rotation': -rotation,
+        'scale': [scale[0], scale[1]],
+    }
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
@@ -20,6 +20,7 @@ import bpy
 
 from . import gltf2_blender_export_keys
 from ...io.exp import gltf2_io_get
+from ...blender.com.gltf2_blender_conversion import convert_texture_transform
 from io_scene_gltf2.io.com import gltf2_io_debug
 #
 # Globals
@@ -381,6 +382,8 @@ def get_texture_transform_from_texture_node(texture_node):
         texture_transform["offset"] = [mapping_node.translation[0], mapping_node.translation[1]]
         texture_transform["rotation"] = mapping_node.rotation[2]
         texture_transform["scale"] = [mapping_node.scale[0], mapping_node.scale[1]]
+
+    texture_transform = convert_texture_transform(texture_transform)
 
     if all([component == 0 for component in texture_transform["offset"]]):
         del(texture_transform["offset"])


donmccurdy:
Thank you! :)

donmccurdy:
@emackey can you think of any reason we shouldn't enable the "Texture Transforms" checkbox by default? I'd see more of an argument for `default=false` if the alternative was to bake the transform into UVs, but currently we just ignore it. Seems like adding a transform is a pretty manual and explicit step in Blender, and we may as well use the extension when appropriate.

emackey:
Yeah, that sounds good.  Can the exporter be smart enough (or is it already?) to avoid writing the extension when it isn't needed?  Certainly when the transform node is missing, but maybe also when the transform is just the identity transform.

donmccurdy:
Not sure, but I agree that should be a requirement.

scurest:
It is already (smart enough).

https://github.com/KhronosGroup/glTF-Blender-IO/blob/c48a4891e2a7cd09df031c817825ac4dd295219f/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py#L388-L396

emackey:
That being the case, who would ever disable the tex transform box?  Can we just remove the UI for it?

donmccurdy:
Fine to remove it, yeah 👍

emackey:
@scurest The math here is better than before, but doesn't seem quite right yet.  The Z rotation direction appears reversed, and I don't think the offset is correct when a rotation is applied.

scurest:
Hm. convert_texture_transform is computing (change UV space)∘(texture transform)∘(change UV space). Changing UV space is (u, v) -> (u, 1-v); write it as Trans[0, 1] Scale[1, -1]. Write the texture transform as Trans[t_x, t_y] Rot[ϑ] Scale[s_x, s_y]. Then I did

Derivation

And evaluated the translation

Evaluate offset

I'm fairly confident in this calculation.

I tested a rotation in the three.js viewer (is this what you're testing with?) and the rotation is indeed different. But looking at the source, I don't think three.js is correct! Looking at GLTFLoader, I see that it sets flipy to false.

https://github.com/mrdoob/three.js/blob/9e49e886d0c11f12e9df978c50dcab1ee0eb6ca7/examples/js/loaders/GLTFLoader.js#L2197

three.js has a UV coordinate system where the bottom-left is (0, 0), like Blender. flipy is accounting for this. But I think  (based on searching the three.js repo for uses of "flipy") flipy is only used to pass `UNPACK_FLIP_Y_WEBGL` to pixelStorei. It doesn't seem to affect the texture transform in any way and the texture transform values are copied directly into the three.js texture object without any modification 

https://github.com/mrdoob/three.js/blob/9e49e886d0c11f12e9df978c50dcab1ee0eb6ca7/examples/js/loaders/GLTFLoader.js#L2197

According to the docs, by default the center of rotation is (0,0), the bottom left corner. So copying them without modification is incorrect because in glTF the rotation should be about the top-left corner.

donmccurdy:
For your second link, I think you meant this snippet?

https://github.com/mrdoob/three.js/blob/9e49e886d0c11f12e9df978c50dcab1ee0eb6ca7/examples/js/loaders/GLTFLoader.js#L587-L617

Hm, threejs does OK according to this test: https://gltf-viewer.donmccurdy.com/#model=https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/TextureTransformTest/glTF/TextureTransformTest.gltf. But we have a very small test set now, and your description sounds right to me.

scurest:
Oops, yeah, I meant `extendTexture`.

I also believe TextureTransformTest is wrong! How was it generated? cc @bghgary

Let's look at the bottom-left test, the one called "Rotation". Without a texture transform, it looks like this, with the arrow pointing to the yellow square

no-transform

Now the rotatation in the glTF is _positive_

https://github.com/KhronosGroup/glTF-Sample-Models/blob/ec880043ae0cead374d7e3a965cc8323c51b06e1/2.0/TextureTransformTest/glTF/TextureTransformTest.gltf#L155

So the rotation should be _counter-clockwise_, ie. towards the red square, not the green square!

emackey:
I was testing in VSCode, with both the BabylonJS and ThreeJS previews, which are in agreement with each other.

I'm tempted to say that since we have two implementations that can both show a "correct" result for the test model, then if that doesn't match the math explained by the spec, maybe it's the spec that's wrong.  If we can just fix it there and have the implementations continue to work as-is, that seems like the better path.

bghgary:
> How was it generated?

I created this asset by hand.

> Now the rotatation in the glTF is positive by about 22 degrees

The [spec](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_texture_transform#gltf-schema-updates) says this for rotation:
> Rotate the UVs by this many radians counter-clockwise around the origin. This is equivalent to a similar rotation of the image clockwise.

This matches the result of Babylon and three. I haven't personally checked to see if the example math is correct in the spec.

scurest:
Okay, I think I understand what's wrong. Essentially we think different directions are counter-clockwise :)

Here is a counter-clockwise rotation in a normal X-right Y-up coordinate system. The matrix is the standard counter-clockwise rotation matrix. This is the matrix that KHR_texture_transform says to use for rotation and is what I've been using.

ccw-rot

So for example, taking a point near (1,0) and rotating it towards (0,1) is a counter-clockwise rotation. So these are the exact same rotation (ie. they transform the coordinates in the same way); normal X-right Y-up on the left, glTF X-right Y-down on the right.

switch

So this is what I've been calling a counter-clockwise rotation even though the right side is going clockwise! I guess we should go like this instead (right-side showing the "equivalent similar rotation of the image clockwise")

cw

Then the matrix is

cw-matrix

ie. the same as a rotation by -ϑ if you use the other matrix.

So if this is what KHR_texture_transform means, its rotation matrix is wrong.

scurest:
In terms of what that changes for convert_texture_transform, the conversion glTF->Blender and Blender->glTF are no longer the same.

The calculation going from a Mapping node to a KHR_texture_transform (ie. whats the exporter uses) is the same, its just that when we read off the final line, we just have to negate the rotation angle. This means that the rotation ultimately does not change (instead of being negated). The offset should still be calculated the same way.

For the importer, when we write down the first line of the calculation, we'd similarly need to negate the rotation angle. This means the rotation ultimately does not change (instead of being negated), but the first component of the offset becomes 

    offset[0] + scale[1] * sin(rotation)

ie. with a + instead of a -.

Sanity check: these two functions are still inverses of each other.

I pushed these changes to my [texture_transform_conversion](https://github.com/scurest/glTF-Blender-IO/tree/texture_transform_conversion) branch. With this change it appears to work in three.js and Babylon.js

o

emackey:
@scurest Sounds great!  This PR is already merged, can you open a new PR from the same branch please?

Issue #185, opened 2018-12-19
Emissive factor missing?


emackey:
I downloaded this car model:  https://sketchfab.com/models/885d9f60b3a9429bb4077cfac5653cf9

After import/export trip through Blender, the emissive channel is not active.  Turns out that the exporter wrote an `emissiveTexture` but not an `emissiveFactor`.  The default `emissiveFactor` is `[0, 0, 0]` so as not to blast all models to full white.  But without the `emissiveFactor` being set, the `emissiveTexture` is multiplied by zero and has no effect.

(Also looks like blending may have been lost from the car windows, that's separate...)

Cyp:
It looks to me as if it's assuming `[1, 1, 1]` is the default `emissiveFactor` here: https://github.com/KhronosGroup/glTF-Blender-IO/blob/master/addons/io_scene_gltf2/blender/imp/gltf2_blender_map_emissive.py#L43 I'd guess that the other `[1, 1, 1]`'s in that file are ok.

But that's in the importer, not the exporter, so I'm not sure that it's related to this bug.

emackey:
I looked more closely at what was imported from this model into Blender.  The emissive map does appear to be hooked up correctly, and works in Eevee and Cycles.  I still need to dig into the code a bit more before I understand what that particular line is for.  But I'm pretty sure this issue is an export issue, since Blender has the right nodes for emission.

julienduroure:
Regarding importer part, it may be a bug (have to have a deeper look on it). If I correctly understand the specification, having an emissiveTexture without any emissiveFactor is the same as having no emissiveTexture, because it will be multiply by [0.0,0.0,0.0] ?

emackey:
> If I correctly understand the specification, having an emissiveTexture without any emissiveFactor is the same as having no emissiveTexture, because it will be multiply by [0.0, 0.0, 0.0] ?

Yes, that is a correct understanding of the glTF spec.

In this case, the original Sketchfab glTF does have the correct emissiveFactor, but the Blender exporter keeps only the texture and loses the factor, resulting in the situation described above.

PR #295, opened 2019-02-12
Add more documentation on how materials get exported.

emackey:
I've added a section explaining how the different material channels and factors can be controlled.

@donmccurdy I think you had a way to build and host a sample copy of the docs before they were pushed to the official Blender doc repo, is that right?  Can we use that to test this PR?


docs/blender_docs/io_gltf2.rst
```diff
@@ -30,7 +30,198 @@ This importer/exporter supports the following glTF 2.0 features:
...
+- Metallic
+- Roughness
+- Baked Ambient Occlusion
+- Normal Map
+- Emissive
+
+.. note::
+
+   When images are used to drive these channels, glTF requires that PNG or JPG format
+   is used to pack the images.
```
donmccurdy: Suggested: `When images are used for these inputs, glTF requires that images be in PNG or JPEG format.`.

docs/blender_docs/io_gltf2.rst
```diff
@@ -30,7 +30,198 @@ This importer/exporter supports the following glTF 2.0 features:
...
+
+.. note::
+
+   glTF does not directly contain Blender's nodes, however, this glTF exporter
+   and importer will use nodes to describe the contents of the glTF file.
```
donmccurdy: I'm not sure what this note is meant to convey – do you mean this _manual_ will use nodes to describe the contents? Users can work in the Blender material sidebar without opening the node graph now, it generates a parallel node graph internally.

docs/blender_docs/io_gltf2.rst
```diff
@@ -30,7 +30,198 @@ This importer/exporter supports the following glTF 2.0 features:
...
+
+These values are read from the "Principled BSDF" node.  If both of these inputs
+are disconnected, the node will display sliders to control their respective
+values between 0.0 and 1.0, and these values will be copied into the glTF.
+
+When using an image, glTF expects the metallic values to be encoded in the
+Blue (``B``) channel, and roughness to be encoded in the Green (``G``) channel of the
+same image.  If images are connected to the Blender node in a manner that
+does not follow this convention, this addon may attempt to adapt the image
+to the correct form during export, costing some performance.
```
donmccurdy: ~costing some performance~ increasing export time.

docs/blender_docs/io_gltf2.rst
```diff
@@ -30,7 +30,198 @@ This importer/exporter supports the following glTF 2.0 features:
...
+   with the glTF standard, allowing it to be used verbatim inside
+   an exported glTF file.
+
+Baked Ambient Occlusion
+^^^^^^^^^^^^^^^^^^^^^^^
+
+glTF is capable of storing a baked ambient occlusion map.  Currently there
+is no arrangement of nodes that causes Blender to use such a map in exactly
+the same way as glTF intends it to be used.  However, if the exporter finds
+a custom node group by the name of ``glTF Metallic Roughness``, and finds an
```
donmccurdy: I don't actually know how to create a custom node group, other than importing one from the old exporter.. this guidance seems a bit unclear without that.

Arguably it would be nice if we detected an Image Texture named `Occlusion`, regardless of  what it's connected to.

docs/blender_docs/io_gltf2.rst
```diff
@@ -30,7 +30,198 @@ This importer/exporter supports the following glTF 2.0 features:
...
+
+To use a Normal Map in glTF, connect an "Image Texture" node's color output
+to a "Normal Map" node's color input, and then connect the "Normal Map" normal
+output to the "Principled BSDF" node's "Normal" input.  The Image Texture node
+for this should have its "Colorspace" setting configured to "Non-Color Data".
+
+The "Normal Map" node must remain on its default setting of "Tangent Space" as
+this is the only type of normal map currently supported by glTF.  The strength
+of the normal map can be adjusted on this node.  The exporter isn't exporting
+these nodes directly, but will use them to locate the correct image and will
```
donmccurdy: This last sentence (`The exporter isn't exporting these nodes directly...`) seems unnecessary with the disclaimer at the top of the manual.

docs/blender_docs/io_gltf2.rst
```diff
@@ -30,7 +30,198 @@ This importer/exporter supports the following glTF 2.0 features:
...
+
+When ``false`` (the default), backface culling is used, and the backs of faces in
+the glTF will not be visible in other software.  When ``true``, backface culling
+is disabled, and double-sided lighting is used, automatically reversing the normal
+vectors of any visible back faces.
+
+To set this value to true, create a custom node group by the name of
+``glTF Metallic Roughness``, add an input value named ``DoubleSided`` with a range
+of 0.0 to 1.0, and set it to 1.0.  There will be no equivalent effect in Blender,
+but the exporter will enable double-sided mode in glTF for this material.
```
donmccurdy: Per above, I'm not sure how to create a custom node group. Long term I still consider this a bug to be fixed in the exporter (by duplicating materials and applying mesh double-sided settings to them, if necessary) rather than something we should guide users through hoops to work around. It seems unlikely that the Principled BSDF node will ever have a doubleSided input.

docs/blender_docs/io_gltf2.rst
```diff
@@ -30,7 +30,198 @@ This importer/exporter supports the following glTF 2.0 features:
...
+
+- **Opaque** - Alpha values are ignored (the default).
+- **Alpha Blend** - Lower alpha values cause blending with background objects.
+- **Alpha Clip** - Alpha values below the ``Clip Threshold`` setting will cause portions
+of the material to not be rendered at all.  Everything else is rendered as opaque.
+
+.. note::
+
+   Be careful using materials with blend mode: **Alpha Blend**.  Some real-time rendering
+   engines may show artifacts if multiple blended polygons overlap.
```
donmccurdy: I think `Some real-time rendering engines may show artifacts if multiple blended polygons overlap` is somewhat understating the complexity here 😅 – suggested:

>Be aware that transparency (or **Alpha Blend** mode) is complex for real-time engines to render, and may behave in unexpected ways after export. Where possible, use **Alpha Clip** mode instead.

docs/blender_docs/io_gltf2.rst
```diff
@@ -30,7 +30,198 @@ This importer/exporter supports the following glTF 2.0 features:
...
+How these values are treated by glTF depends on the selected blend mode.
+
+With the "Eevee" rendering engine selected, each material has a "Blend Mode" on the
+material settings panel.  Use this setting to govern how alpha values from the
+Base Color channel are treated in glTF.  Three settings are supported by glTF:
+
+- **Opaque** - Alpha values are ignored (the default).
+- **Alpha Blend** - Lower alpha values cause blending with background objects.
+- **Alpha Clip** - Alpha values below the ``Clip Threshold`` setting will cause portions
+of the material to not be rendered at all.  Everything else is rendered as opaque.
```
donmccurdy: ^indentation needed to keep this line within the bullet point.

docs/blender_docs/io_gltf2.rst
emackey: Deleted.

docs/blender_docs/io_gltf2.rst
emackey: Yes.  I was thinking we still have to so something better about Occlusion, but I couldn't write documentation on what that might be until after it gets implemented.  So the PR for that should include an update to this section, or a new issue to update this section.

Creating custom node groups must be covered elsewhere in the Blender manual.  It seemed inappropriate to delve into those details here.

docs/blender_docs/io_gltf2.rst
emackey: Yeah, but I think I just removed that...

docs/blender_docs/io_gltf2.rst
emackey: When the bug gets fixed, I'll update this section too.

emackey:
Thanks @donmccurdy!  If you get a chance to refresh the doc build, that would be great.

docs/blender_docs/io_gltf2.rst
j-conrad: As a point of reference, I knew nothing of how to create Node Groups but a quick search found the answer (CTRL+G in the Node Editor). However, following the instructions for adding an 'Occlusion' input didn't cause any AO information to be exported. I tend to agree with the suggestion of simply having an Image Texture named `Occlusion` without the need for a Node Group, although it does keep things tidier.

Image shows the Node Group in the foreground, main graph in background with collapsed Node Group on the far left.
image

dsinni: @j-conrad Creating custom node groups seems a bit tricky (at least in 2.80), since it doesn't seem possible to create an empty group when no other node groups are selected. I found that I needed to create some input, select it, press `Ctrl + G` in the Shader Editor and then customize it in the properties panel (press `N`).

That said, I believe the node graph above is incorrect; the texture shouldn't be part of the group at all and should just be passed to an input named `Occlusion`. It's also important that the node group is named `glTF Metallic Roughness`.

This worked for me, for both double-sided materials and baked occlusion:

image

dsinni: One thing I'm unclear on, however, is whether or not `Separate RGB` is needed as shown below:

image

The result appeared the same when not using a `Metallic` or `Roughness` map, but I did notice that without `Separate RGB`, the texture was monochromatic black and white, and monochromatic cyan when splitting off the red channel.

No `Separate RGB`:

image

`Separate RGB` using R channel:


image

When combining with `Metallic` (`B`) and `Roughness` (`G`), I can only assume that the separation is necessary, but would appreciate any clarification on this.

j-conrad: I would think your selection of ‘non-color data’ would be enough to not require splitting out the red channel, but I’m new enough to Blender to not know for sure.

I think what confuses me is that we’re putting Occlusion in a separate image texture to begin with. Why can’t it live within the Red channel of the Metallic/Roughness image? I understand that the Principled BSDF node doesn’t have an AO input (which is why we’re having to do any of this in the first place). But it seems like the exporter could simply look for data in that Red channel and assume it’s the Ambient Occlusion map.

j-conrad: It’s worth noting that it’s a fairly common practice to already have your Ambient Occlusion stores there (Red channel). Common enough that you see people referring to the texture as “ARM”

emackey: "ORM" please, not "ARM" 😸 

But yes, the exporter was never made to understand the "Separate RGB" node being used with occlusion, even though glTF stores occlusion in Red.  This deficiency is just a quirk of the history of the project, not a real design decision.

Ideally, there should be a single ORM image connected to a single Separate RGB node, and G and B should be connected to Principled BSDF, and the R side should be connected to... something.  What the "something" could be, is still up for debate.  We need to figure it out, implement it, and then document whatever it is.

j-conrad: oops..  Yeah, I definitely meant 'ORM'.

Using the Node Group is working for me now too, including 'DoubleSided'. Although, I noticed that naming the input `doubleSided` (camel case, as per the glTF spec) does not work. Seems better to use the glTF spec names verbatim, unless there's some special reason not to.

Also, what do you think about the idea of using the Node Group as the single place to put any additional inputs, for example, `baseColorFactor`, `metallicFactor` and `roughnessFactor`? I understand that those two values automatically appear in the JSON when a texture is not specified, but since the spec states that it's possible to use a texture AND have it multiplied by a Factor, it seems like a good place to specify in such a scenario since right now, if a texture gets used, those factors disappear.

emackey: Actually the factors work by creating extra nodes that multiply with the Image Texture nodes.  I mention this in the PR here, but I don't have a screenshot for it.  I think if you have such a glTF already, the importer may generate the correct pattern of nodes for this.

The input names, and for that matter the entire behavior of the `glTF Metallic Roughness` custom node, is legacy.  They date from the previous exporter which worked exclusively off a custom node grouping, and was not compatible with Principled BSDF.  The names are meant to be user-facing, so they do not follow a camelCase pattern.

j-conrad: Regarding camelCase, I suppose that makes sense. Moving forward, most users will be using this new Blender documentation for reference without needing to look at the glTF specs. In the interim, for people like me, it has the potential to cause some mistakes but I can deal with it.

I had completely forgotten about the glTF **importer**-- what a great way to understand how a glTF is set up by seeing nodes get generated.  :)

dsinni: @j-conrad If any clarification is still needed for using a single ORM texture, I believe you can use something like the graph below:

image

Please let me know if I'm mistaken.

j-conrad: That’s exactly what my graph looks like and is also working. Thanks!

emackey:
The [Blender writing guide](https://docs.blender.org/manual/en/dev/about/contribute/guides/writing_guide.html) mentions:

> Do not simply copy the tooltips from Blender. – People will come to the manual to learn *more* than is provided by the UI.

I'm thinking of removing the section where we list out all the tooltips from all the import/export UI checkboxes.  The users can see all this information directly on the UI, and there's no new information here.  In its place I could put a description of `.gltf` vs `.glb` and why users could choose one vs another.

donmccurdy:
> I'm thinking of removing the section where we list out all the tooltips from all the import/export UI checkboxes.

^Note that [the other import/export addons ](https://docs.blender.org/manual/en/dev/addons/index.html#import-export) all include this section, even if only with "TODO" as the description. I'd prefer that we not remove it, but if we can add more detail that's even better.

emackey:
@donmccurdy I had some RST issues but I think they're resolved now.  Can you run your build when you get a moment?  It's pretty out of date from what's in the PR now.  Thanks!

j-conrad:
I'm following the instructions on exporting an Emissive Map with my model and I'm a little confused. While the Emissive map clearly shows up in the JSON, it appears that its `emissiveFactor` is zero by default.
Piping the Emissive image through a MixRGB did not include `emissiveFactor` in the JSON upon export. Ok, so maybe I'm just doing it wrong...

So to figure out the proper graph, I edited the JSON manually to include an `emissiveFactor` of 1,1,1. This immediately showed up in all the various glTF viewers so no problem there.
However, when I **import** that proper glTF into Blender, the graph shows no signs of an `emissiveFactor` anywhere that I can find. And when I re-export that file from Blender, once again, the `emissiveFactor` is not in the JSON and thus, the emissive texture doesn't show.
Any ideas what might be happening?

emackey:
> I'm following the instructions on exporting an Emissive Map with my model and I'm a little confused. While the Emissive map clearly shows up in the JSON, it appears that its `emissiveFactor` is zero by default.

That's a bug.  Issue #185.

emackey:
There's a test failure here, but I think it's the result of test configuration changes being made elsewhere.

j-conrad:
Despite creating a single 'ORM' (occlusion, roughness, metallic) texture and wiring the RGB up to a Node Group (with 'Occlusion', Roughness, and Metallic respectively, the exported glTF produces two separate textures-- 1 for Occlusion and 1 for Roughness/Metallic. Not sure why it's creating an extra copy of the same texture but I assume this is bumping up the file size.
image

donmccurdy:
Updated demo of the docs: https://gltf-blender-docs-ufucdfzqtu.now.sh/addons/io_gltf2

The documentation looks good to me, and can be merged. The duplicate texture would be a bug (let's file a new issue). I also think the custom node should be allowed to be named `glTF Export` (case insensitive) and we should document that rather than `glTF Metallic Roughness`, but that can be changed later.

emackey:
@julienduroure Note that the `images` folder has moved to the root of the repo, and the `copy_doc.py` script is updated with the new location.  There are several new images here.

julienduroure:
@emackey thanks!
I will merge it into blender doc during the following week

emackey:
Thanks @julienduroure!

For anyone else following this, the new documentation is now published!  If you find problems or have further questions, make a new issue in this repo.  Thanks!

https://docs.blender.org/manual/en/2.80/addons/io_gltf2.html

donmccurdy:
Very nice, thanks @emackey and @julienduroure!

PR #298, opened 2019-02-14
Export option cleanup

### LINKED ISSUE #109: Review of export options

emackey:
- ~~**Apply Modifiers** set to `true` per https://github.com/KhronosGroup/glTF-Blender-IO/issues/109#issuecomment-463713414~~ Edit: Reverted this, it caused test failures.

- **Texture Transforms** removed from the UI, always on now.  If you don't want it, don't use the mapping nodes, or leave the mapping nodes set to the default (identity transform).

- **Materials** was alone on a tab by itself, it has been moved over to the Mesh tab.

Fixes #109.

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -156,7 +156,7 @@ class ExportGLTF2_Base:
     export_apply = BoolProperty(
         name='Apply Modifiers',
         description='Apply modifiers (excluding Armatures) to mesh objects',
-        default=True
+        default=False
     )
 
     export_animations = BoolProperty(
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -82,7 +82,6 @@ class ExportGLTF2_Base:
         items=(('GENERAL', "General", "General settings"),
                ('MESHES', "Meshes", "Mesh settings"),
                ('OBJECTS', "Objects", "Object settings"),
-               ('MATERIALS', "Materials", "Material settings"),
                ('ANIMATION', "Animation", "Animation settings")),
         name="ui_tab",
         description="Export setting categories",
@@ -157,7 +156,7 @@ class ExportGLTF2_Base:
     export_apply = BoolProperty(
         name='Apply Modifiers',
         description='Apply modifiers (excluding Armatures) to mesh objects',
-        default=False
+        default=True
     )
 
     export_animations = BoolProperty(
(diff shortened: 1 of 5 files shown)


emackey:
Test failure - Don't merge yet

emackey:
Still 26 failing tests here.  I need to learn more about the test framework to understand what's going on with this.  Any advice would be appreciated.

emackey:
OK, tests pass now.  I reverted the change to "Apply Modifiers", apparently setting that to True by default was causing 26 test failures.

donmccurdy:
Thanks! I'm fine with leaving Apply Modifiers disabled by default, but haven't dug into the test failures yet. We can look at that later if needed.

PR #299, opened 2019-02-15
Fix texture transform conversion

scurest:
See #291 for details. Specifically https://github.com/KhronosGroup/glTF-Blender-IO/pull/291#issuecomment-464199974

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
@@ -20,7 +20,7 @@ import bpy
 
 from . import gltf2_blender_export_keys
 from ...io.exp import gltf2_io_get
-from ...blender.com.gltf2_blender_conversion import convert_texture_transform
+from ...blender.com.gltf2_blender_conversion import texture_transform_blender_to_gltf
 from io_scene_gltf2.io.com import gltf2_io_debug
 #
 # Globals
@@ -373,17 +373,17 @@ def get_texture_transform_from_texture_node(texture_node):
     if not isinstance(mapping_node, bpy.types.ShaderNodeMapping):
         return None
 
-    texture_transform = {}
+    mapping_transform = {}
     if mapping_node.vector_type == 'TEXTURE':
-        texture_transform["offset"] = [-mapping_node.translation[0], -mapping_node.translation[1]]
-        texture_transform["rotation"] = -mapping_node.rotation[2]
-        texture_transform["scale"] = [1.0 / mapping_node.scale[0], 1.0 / mapping_node.scale[1]]
+        mapping_transform["offset"] = [-mapping_node.translation[0], -mapping_node.translation[1]]
+        mapping_transform["rotation"] = -mapping_node.rotation[2]
+        mapping_transform["scale"] = [1.0 / mapping_node.scale[0], 1.0 / mapping_node.scale[1]]
     elif mapping_node.vector_type == 'POINT':
-        texture_transform["offset"] = [mapping_node.translation[0], mapping_node.translation[1]]
-        texture_transform["rotation"] = mapping_node.rotation[2]
-        texture_transform["scale"] = [mapping_node.scale[0], mapping_node.scale[1]]
+        mapping_transform["offset"] = [mapping_node.translation[0], mapping_node.translation[1]]
+        mapping_transform["rotation"] = mapping_node.rotation[2]
+        mapping_transform["scale"] = [mapping_node.scale[0], mapping_node.scale[1]]
 
-    texture_transform = convert_texture_transform(texture_transform)
+    texture_transform = texture_transform_blender_to_gltf(mapping_transform)
 
     if all([component == 0 for component in texture_transform["offset"]]):
         del(texture_transform["offset"])
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
@@ -17,6 +17,7 @@
 #
 
 import bpy
+from mathutils import Vector, Matrix
 
 from . import gltf2_blender_export_keys
 from ...io.exp import gltf2_io_get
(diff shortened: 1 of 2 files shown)


emackey:
@scurest Thanks for sticking with this thorny problem.  I can confirm this fix works correctly (for tranlate, rotate, and scale) when the Mapping node is in `Point` mode.

When the mapping node is in `Texture` mode, the offset appears incorrect.  I think this may be the result of pre-existing code that simply negates all the input values.  `Texture` mode is described as the inverse transform of `Point`, so I'm guessing that simply negating the offset & rotation is not the proper way to calculate the inverse transform.

If you like we could consider that a separate issue and merge this as-is.  But if you're willing to take on `Texture` mode as part of this PR, that would be great!

scurest:
Yeah, I was just assuming that after that `if`, the equivalent `POINT` transform had been calculated, haha.

It is indeed the inverse transform ([Blender source code](https://github.com/sobotka/blender/blob/27c72e22aa41e13cb94bb3e5f9a59b960ba5aa09/intern/cycles/render/nodes.cpp#L105-L110)). That means you not only have to invert all the TRS properties, you need to compose them in the opposite order too. Unfortunately, this runs into the problem that [the inverse of a TRS transform is not a TRS transform in general](https://github.com/KhronosGroup/glTF/issues/1507#issuecomment-446747632).

In the special case that the scaling is homogeneous (scale[0] == scale[1]) OR the rotation is zero, then the inverse of a `POINT` transform can be computed with

````diff
diff --git a/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py b/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
index 6339e57..0a2a59a 100644
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
@@ -17,6 +17,7 @@
 #
 
 import bpy
+from mathutils import Vector, Matrix
 
 from . import gltf2_blender_export_keys
 from ...io.exp import gltf2_io_get
@@ -373,15 +374,32 @@ def get_texture_transform_from_texture_node(texture_node):
     if not isinstance(mapping_node, bpy.types.ShaderNodeMapping):
         return None
 
+    if mapping_node.vector_type not in ['TEXTURE', 'POINT']:
+        return None
+
     mapping_transform = {}
+    mapping_transform["offset"] = [mapping_node.translation[0], mapping_node.translation[1]]
+    mapping_transform["rotation"] = mapping_node.rotation[2]
+    mapping_transform["scale"] = [mapping_node.scale[0], mapping_node.scale[1]]
+
     if mapping_node.vector_type == 'TEXTURE':
-        mapping_transform["offset"] = [-mapping_node.translation[0], -mapping_node.translation[1]]
-        mapping_transform["rotation"] = -mapping_node.rotation[2]
-        mapping_transform["scale"] = [1.0 / mapping_node.scale[0], 1.0 / mapping_node.scale[1]]
-    elif mapping_node.vector_type == 'POINT':
-        mapping_transform["offset"] = [mapping_node.translation[0], mapping_node.translation[1]]
-        mapping_transform["rotation"] = mapping_node.rotation[2]
-        mapping_transform["scale"] = [mapping_node.scale[0], mapping_node.scale[1]]
+        def inverted(mapping_transform):
+            offset = mapping_transform['offset']
+            rotation = mapping_transform['rotation']
+            scale = mapping_transform['scale']
+            if bpy.app.version >= (2, 80, 0):
+                new_offset = Matrix.Rotation(-rotation, 3, 'Z') @ Vector((-offset[0], -offset[1], 1))
+            else:
+                new_offset = Matrix.Rotation(-rotation, 3, 'Z') * Vector((-offset[0], -offset[1], 1))
+            new_offset[0] /= scale[0]; new_offset[1] /= scale[1]
+            inverted_mapping_transform = {
+                'offset': new_offset[0:2],
+                'rotation': -rotation,
+                'scale': [1/scale[0], 1/scale[1]],
+            }
+            return inverted_mapping_transform
+
+        mapping_transform = inverted(mapping_transform)
 
     texture_transform = texture_transform_blender_to_gltf(mapping_transform)
 
````

Test:

texture_mapping

But I don't know what to do when those conditions aren't satisfied. There isn't a perfectly accurate way to export the mapping node in that case.

**Comment by emackey** [comment]
Right, since we're targeting a glTF extension that only encodes TRS, it makes sense that an inversion of that may not be able to be correctly encoded in the extension.  Sounds like this would still be a better option than just negating all the values though.

**Comment by scurest** [comment]
So but what exactly did you want me to do when those conditions aren't satisfied?

**Comment by emackey** [comment]
Elsewhere we import and use [print_console](https://github.com/KhronosGroup/glTF-Blender-IO/blob/32a6f05f64dd33e273e979d36d508835ca4243bf/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py#L77) to dump out warnings, is that an option?

**Comment by scurest** [comment]
Yes, can do. I printed a warning in all cases where we find a Mapping node but can't calculate an equivalent KHR_texture_transform.

As a bonus, I implemented VECTOR transforms too (since they're easy :p).

**Comment by scurest** [comment]
However, bad news. I found a file that Three and Babylon disagree on. @donmccurdy 

disagree
[disagree.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/2886226/disagree.zip)

I believe Babylon is correct here.

File: addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py [hunk 1]
```diff
@@ -373,17 +374,64 @@ def get_texture_transform_from_texture_node(texture_node):
...
+        )
+        return None
+
+    if mapping_node.rotation[0] or mapping_node.rotation[1]:
+        # TODO: can we handle this?
+        gltf2_io_debug.print_console("WARNING",
+            "Skipping exporting texture transform because it had non-zero "
+            "rotations in the X/Y direction; only a Z rotation can be exported!"
+        )
+        return None
```
donmccurdy: Consider logging the warning but continuing to export here – seems like the X/Y rotation can be ignored.

donmccurdy:
Thanks! I'll take a look. 👍

addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
scurest: They definitely affect the transform. You'll get a different results in glTF and Blender if the exporter just ignores them.

donmccurdy: Ok, guess I'm misremembering.

addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
```diff
@@ -373,17 +374,64 @@ def get_texture_transform_from_texture_node(texture_node):
...
+        def inverted(mapping_transform):
+            offset = mapping_transform["offset"]
+            rotation = mapping_transform["rotation"]
+            scale = mapping_transform["scale"]
+
+            # Inverse of a TRS is not always a TRS. This function will be right
+            # at least when the following don't occur.
+            if abs(rotation) > 1e-5 and abs(scale[0] - scale[1]) > 1e-5:
+                return None
+            if abs(scale[0]) < 1e-5 and abs(scale[1]) < 1e-5:
```
scurest: This should be `or` to make sure the scaling is invertible. Amending.

donmccurdy:
Thanks! Filed https://github.com/mrdoob/three.js/issues/15831 to track the ThreeJS issue.

PR #304, opened 2019-02-19
Export blender double-sided mesh flag to gltf materials.

Cyp:
Just as making this pull request, it showed another new https://github.com/KhronosGroup/glTF-Blender-IO/commit/cb1f27b6106f77538004b5b331ebc44a7c69969d commit which does the same thing. So I suppose only either that one or this one will be useful, but not both.

Edit: Looking at that version, I think it wouldn't duplicate materials if needed, but this version should.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
@@ -27,7 +27,7 @@ from io_scene_gltf2.blender.exp import gltf2_blender_get
 
 
 @cached
-def gather_material(blender_material, export_settings):
+def gather_material(blender_material, mesh_double_sided, export_settings):
     """
     Gather the material used by the blender primitive.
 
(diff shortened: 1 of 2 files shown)


emackey:
I'm working on this too.  I tried this approach originally, but discarded it because it caused unpredictable results when materials were shared among meshes.

I'll have my own PR open soon, so we can compare in more detail.

emackey:
I tested this and it works really well, thanks!  I do like that it will automatically duplicate materials when needed for differences in double-sidedness, that's great.  I don't have any worries about letting go of that commit of mine you linked to.

I'll leave this open until @donmccurdy gets a peek at it, then we'll merge.

emackey:
Fixes #154.

emackey:
@donmccurdy The behavior resulting from this is:

- When "double-sided lighting" is enabled on a Blender mesh, the materials used by that mesh will be double-sided in glTF.
- When a double-sided mesh and a single-sided mesh share the same Blender material, the exporter will export two different copies of that material with the same name (allowed in glTF), wired correctly so the double-sided mesh gets the double-sided copy of the material.
- When meshes share materials and also share double-sided settings, they will continue to share glTF materials as before.
- Only single-sided materials can be influenced by the old custom nodes.  But with this change, the custom nodes shouldn't be needed for this input anymore.

emackey:
Merging this, thanks @Cyp.

donmccurdy:
Nice clean fix, thanks! :)

PR #305, opened 2019-02-20
Feature: export at frame 0

### LINKED ISSUE #302: Node rotation not exported if node scale is 0
The node rotation is not exported if the node's scale is 0.

A case where this actually causes an issue is where there is a static rotation and an animated scale, and the export happens to be done during a frame where the scale is 0.

This happens since `blender_object.matrix_local` doesn't contain rotation information when the scale is 0, and `gltf2_blender_gather_nodes.__gather_trans_rot_scale` uses `matrix_local`.

I'm not sure what a good clean solution is. Just using `rotation_*` wouldn't work since the `matrix_parent_inverse` has to somehow be taken into account. Not sure it's solvable in general, since the transformation might not even be decomposable. One non-ideal solution might be to detect if a scale axis is 0, and if so temporarily change the scale to 1 and recompute the rotation.

A workaround is to only animate the scale down to 0.00001.

UX3D-schmithuesen:
fixed #302


donmccurdy:
Is this compatible with both states of the `Keyframes Start at 0` checkbox?

UX3D-schmithuesen:
> Is this compatible with both states of the `Keyframes Start at 0` checkbox?

All this does, is set the current frame to 0 before exporting and then restore it after exporting is done. I don't know how the `Keyframes Start at 0` functions, but I wouldn't expect that they interfere. I will look it though!

UX3D-schmithuesen:
Okay, `Keyframes Start at 0` is not referenced in the code in any productive way. Also, I didn't have a problem exporting animations with the first keyframe at 0 or 1. That settles your other question, I guess :D

donmccurdy:
That answers that, thanks! 😅

PR #313, opened 2019-02-21
Feature: handle 'reload scripts'

### LINKED ISSUE #279: Handle reload scripts properly
When working on the addon, it would be very useful if Blenders 'reload scripts' feature worked properly. However to support reloading of submodules it needs some importlib functionality added.

Old exporter did this: https://github.com/KhronosGroup/glTF-Blender-Exporter/blob/master/scripts/addons/io_scene_gltf2/__init__.py#L23-L40

UX3D-schmithuesen:
fixes #279

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -16,9 +16,38 @@
 # Imports
 #
 
+import importlib
+from pathlib import Path
+
+
+modules = locals()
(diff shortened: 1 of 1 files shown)


pjoe:
Looks awesome, nice way to dynamically figure out which modules to reload. Was worried this would have to been handled all over the place. Very nice :)

Issue #323, opened 2019-02-24
Linked data not shared after export


Kaesebrot84:
I am trying to export two objects which are supposed to share mesh data (`NORMAL`, `POSITION`, `indices` etc.). I also want the objects' materials not being shared.

In order to achieve this I linked the objects' meshes via `Link Data`. I also linked the materials to objects instead of linking them to data.

![structure](https://user-images.githubusercontent.com/13865115/53298395-bdc50400-382d-11e9-8f0d-f987b8494cfe.PNG)

![linked_to_object](https://user-images.githubusercontent.com/13865115/53298396-c289b800-382d-11e9-8792-de6c13e48339.PNG)

![pawns](https://user-images.githubusercontent.com/13865115/53298399-c4ec1200-382d-11e9-9a14-c71090f09d86.PNG)

When I export to gltf it seems like according primitives are still pointing to different accessors. I would have expected them to share buffer data.

```json
        {
            "name" : "pawn_1.002",
            "primitives" : [
                {
                    "attributes" : {
                        "NORMAL" : 6,
                        "POSITION" : 4,
                        "TEXCOORD_0" : 5
                    },
                    "indices" : 7,
                    "material" : 1
                }
            ]
        },
        {
            "name" : "pawn_1.002",
            "primitives" : [
                {
                    "attributes" : {
                        "NORMAL" : 10,
                        "POSITION" : 8,
                        "TEXCOORD_0" : 9
                    },
                    "indices" : 11,
                    "material" : 0
                }
            ]
        }
```

The according accessors (`NORMAL` in this case) are also pointing to different `bufferView` elements:

```json
        {
            "bufferView" : 7,
            "componentType" : 5126,
            "count" : 5504,
            "type" : "VEC3"
        },        
        {
            "bufferView" : 12,
            "componentType" : 5126,
            "count" : 5504,
            "type" : "VEC3"
        }
```

How can I make primitives in exported gltf files share buffer information? According to Blender's Add-On screen I am using `Import-Export: glTF 2.0 format, 0.0.1`.

donmccurdy:
When duplicating objects with Alt+D the meshes are reused. I'm not sure what's different in this case – do you get the same result if they share a material?

Kaesebrot84:
> When duplicating objects with Alt+D the meshes are reused. I'm not sure what's different in this case – do you get the same result if they share a material?

[...]

Issue #55, opened 2018-10-23
Animations and meshes duplicated, even with same animation, parent inverse and modifier set


Cyp:
If multiple nodes use the same animation, and the nodes have different parent inverses, the animations are wrong for nodes with a different parent inverse to the first node.

I'm not sure what a sane solution is that doesn't involve duplicating the animation (with different parent inverses applied to it), but reporting in case there is a possible solution.

UX3D-nopper:
Let's see, if dublicating the nodes can be avoided ...

UX3D-becher:
A refactored exporter is now enabled by default. It now creates one animation per node, possibly fixing your issue. @Cyp, could you verify that the parent inverses are now correct?

Cyp:
Sorry for the late response.

Seemed to still be getting the issue, but actually it's completely failing to save anything, so I'm not sure…

Doesn't seem to be working, after trying to reset the settings, when saving I'm getting this:
```
Traceback (most recent call last):
  File "~/.config/blender/2.79/scripts/addons/io_scene_gltf2/__init__.py", line 376, in execute
    return gltf2_export.save(self, context, export_settings)
  File "~/.config/blender/2.79/scripts/addons/io_scene_gltf2/gltf2_export.py", line 85, in save
    generate_glTF(operator, context, export_settings, glTF)
  File "~/.config/blender/2.79/scripts/addons/io_scene_gltf2/gltf2_generate.py", line 2872, in generate_glTF
    generate_animations(operator, context, export_settings, glTF)
  File "~/.config/blender/2.79/scripts/addons/io_scene_gltf2/gltf2_generate.py", line 802, in generate_animations
    process_object_animations(blender_object, action)
  File "~/.config/blender/2.79/scripts/addons/io_scene_gltf2/gltf2_generate.py", line 628, in process_object_animations
    if blender_action.name not in animations:
AttributeError: 'NoneType' object has no attribute 'name'

location: <unknown location>:-1
```

And (not really related, I think), I couldn't even get back in to the export menu after selecting export indices to `UNSIGNED SHORT`.
```
Traceback (most recent call last):
  File "~/.config/blender/2.79/scripts/addons/io_scene_gltf2/__init__.py", line 291, in invoke
    setattr(self, k, v)
  File "/usr/share/blender/2.79/scripts/modules/bpy_types.py", line 611, in __setattr__
    return setattr(properties, attr, value)
[...]

PR #357, opened 2019-03-12
Deduplicate accessors.

### LINKED ISSUE #353: Exporter should reuse animation sampler's input when they are the same
Consider (again) this model
[simple.zip](https://github.com/KhronosGroup/glTF-Blender-Exporter/files/1235360/simple.zip)
And the resulted exported file
[SimpleAnimPose.zip](https://github.com/KhronosGroup/glTF-Blender-Exporter/files/1235393/SimpleAnimPose.zip)
It's been exported checking the "force sample animation" option.
In the animations/samplers section you can see that each input points to a different accessor. If you look at those accessors, they look similar, but each one points toward a different bufferView. Those buffer views, in turn, points to the same buffer, but with a different byteOffset each time, meaning that the data is duplicated in the bin file too.
If I check the data when loading the files it is strictly identical from one input to another.

This is perfectly valid in a spec point of view, so that's not really a bug, but maybe an enhancement. It would greatly reduce the files size if this data was reused.

Cyp:
Meshes with modifiers remain duplicated, but should now at least share accessors.

Fixes #353 and helps a bit with #323 / #55. Doesn't help with performance due to regenerating the same mesh multiple times.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_accessors.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_accessors.py
@@ -1,0 +1,47 @@
+# Copyright 2018 The glTF-Blender-IO authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


donmccurdy:
Thanks! :)

PR #383, opened 2019-03-23
Fix a bug in import of shape keys

capnm:
io_scene_gltf2/blender/imp/gltf2_blender_mesh.py", line 135, in set_mesh
    if i >= len(prim.targets):
TypeError: '>=' not supported between instances of 'tuple' and 'int'

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
@@ -109,21 +109,21 @@ class BlenderMesh():
             obj.shape_key_add(name="Basis")
 
         current_shapekey_index = 0
-        for i in range(max_shape_to_create):
+        for sk in range(max_shape_to_create):
 
             # Check if this target has POSITION
-            if 'POSITION' not in prim.targets[i].keys():
-                gltf.shapekeys[i] = None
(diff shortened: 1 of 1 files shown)


julienduroure:
Do you have any sample file that have issue you want to solve with this PR ?
This will help me testing it

capnm:
Sorry at the moment I have nothing I could share, but it's an obvious regression introduced in
https://developer.blender.org/rBA80ad5c85bd7b8818ec21f7019241a8f8409aa30d

capnm:
Finally, I could produce a minimal sample:
[sk-test-01-005.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3000115/sk-test-01-005.zip)


Please try to include links to Github issues or PRs in your Blender 2.8 commits.
Many Thanks.

emackey:
Actually if I were to nitpick this a tiny bit, perhaps it should be the outer loop's variable that should be renamed, as that loop fills more than a page of code and so is deserving of more than a one-letter variable name.

capnm:
> Actually if I were to nitpick this a tiny bit, perhaps it should be the outer loop's variable that should be renamed, as that loop fills more than a page of code and so is deserving of more than a one-letter variable name.

Good point, I hope now it has become clearer.
Thanks for the review!

PR #395, opened 2019-04-01
Feature: google draco exporter

UX3D-schmithuesen:
A Blender build with https://developer.blender.org/D4501 is required for Draco to work, but nothing breaks without it.

- [x] works on Linux
- [x] works on Windows
- [x] works on MacOS


donmccurdy:
@UX3D-schmithuesen just to double-check, is the contribution to Blender using the `master` branch of the Draco repository, or the [`gltf_2.0_draco_extension`](https://github.com/google/draco/tree/gltf_2.0_draco_extension) branch? Maybe they're both in sync anyway, but I'm not sure and we may need to double-check.

PR #397, opened 2019-04-01
Fix/prune unused code

UX3D-becher:
removed some unused functions


emackey:
@julienduroure Just a heads up, this PR deletes a file from the repo.  After this gets merged, I think you have to take an extra manual step to make the delete happen in the Blender repo.

`addons/io_scene_gltf2/io/exp/gltf2_io_get.py`

julienduroure:
@emackey : Ok, I will try to remember it when I will push it. Don't hesitate to poke me if I forgot

PR #398, opened 2019-04-01
fix the ordering of inverse bind matrices

### LINKED ISSUE #232: Problem exporting skinned mesh with bones that have no skins
After I export a gltf, my skinned mesh appears to be very odd.

In Blender:
screenshot 2019-01-15 19 21 16

In gltf viewer:
screenshot 2019-01-15 19 24 08

While very visually interesting, this isn't the desired result 😄 

I am manually adding leaf bones to my armature so I can see the lone child bones in my engine.

I can manually "fix" this by reordering the skins, joints list.

`    
"skins" : [
        {
            "inverseBindMatrices" : 14,
            "joints" : [
                54,
                43,
                42,
                41,
                2,
                1,
                0,
                21,
                20, ...
`
If I take the indexes of my new leaf joints and move them to the bottom, it will render properly.

I wonder if the joints are not ordered properly on export?

I'm also curious if this is related to #184 

SOURCE BLENDER: 
[female-adult.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/2762044/female-adult.blend.zip)

UX3D-becher:
should fix #232 

Most skinning issues should be fixed by this commit, we should test and potentially close all of them

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_skins.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_skins.py
@@ -63,33 +63,44 @@ def __gather_extras(blender_object, export_settings):
     return None
 
 def __gather_inverse_bind_matrices(blender_object, mesh_object, export_settings):
-    inverse_matrices = []
-
     axis_basis_change = mathutils.Matrix.Identity(4)
     if export_settings[gltf2_blender_export_keys.YUP]:
         axis_basis_change = mathutils.Matrix(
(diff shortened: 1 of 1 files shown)


donmccurdy:
This is looking much better on initial tests, thanks! In a couple examples (attached) there's still something unusual going on, as if parts of the model have been mirrored left to right. See the arms in the "lain" example when playing either the HoodUp or HoodDown animation clips.

[Archive.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3034482/Archive.zip)

PR #404, opened 2019-04-03
Feature/jpg export

UX3D-becher:
Refactored the whole image processing to easily allow jpg export.


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
```diff
@@ -149,64 +150,26 @@ def split_pixels_by_channels(image: bpy.types.Image, export_settings) -> typing.
...
-                target_channel = 0
-                source_channel = 0
-                source_channels_length = len(pixels)
+            image = gltf2_blender_image.ExportImage.from_blender_image(result.shader_node.image)
+
+            if composed_image is None:
+                composed_image = gltf2_blender_image.ExportImage.zero_image(image.width, image.height)
 
             # Change target channel for metallic and roughness.
             if elem.to_socket.name == 'Metallic':
```
emackey: This line, and the one 2 below it, didn't change, but I think they have been harboring a bug.  `elem.to_socket.name` is the name of the output of the farthest-left link, the last one traversed by the `for` loop above.  I'm pretty sure these lines should both change to just `if socket.name ==`, which will look at the final destination of the links.

addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
```diff
@@ -0,0 +1,122 @@
...
+
+    @classmethod
+    def from_blender_image(cls, blender_image: bpy.types.Image):
+        img = np.array(blender_image.pixels)
+        img = img.reshape((blender_image.size[0], blender_image.size[1], blender_image.channels))
+        return ExportImage(img=img)
+
+    @classmethod
+    def zero_image(cls, width, height, num_channels: int = 4):
+        img = np.zeros((width, height, num_channels))
```
emackey: ones are safer than zeros as a default fill, particularly for AO and Roughness.

addons/io_scene_gltf2/__init__.py
```diff
@@ -107,6 +107,20 @@ def __init__(self):
...
+                'encode and save textures as .jpeg files. Be aware of a possible loss in quality.'),
+               ('PNG', 'png image format (.png)',
+                'encode and save textures as .png files.')
+               ),
+        description=(
+            'Output format for images. PNG is lossless and generally preferred, but JPEG might be preferable for web '
+            'applications due to the smaller file size'
+        ),
+        default='PNG'
+    )
```
donmccurdy: We circled around different ways of doing this in glTF-Blender-Exporter. In https://github.com/KhronosGroup/glTF-Blender-Exporter/pull/181 I tried to be clever about using presence or absence of alpha to determine output format, which we eventually reverted in favor of https://github.com/KhronosGroup/glTF-Blender-Exporter/pull/221, to simply have export match the input format wherever possible.

In hindsight, I think matching the input format is the best approach. Consider:

- Exporting a JPEG source image as JPEG should not result in any loss of quality.
- Users retain greater control of their textures, because they can choose format and compression settings before bringing them into Blender.
- Users may not want the same format for all images. In particular, base color should almost always use JPEG on web, whereas normal maps should almost always use PNG.

For the reasons above, I'm not sure we should expose a UI checkbox for the format of all textures, and if we do, "Source format" should be the default option.

donmccurdy:
We circled around different ways of doing this in glTF-Blender-Exporter. In https://github.com/KhronosGroup/glTF-Blender-Exporter/pull/181 I tried to be clever about using presence or absence of alpha to determine output format, which we eventually reverted in favor of https://github.com/KhronosGroup/glTF-Blender-Exporter/pull/221, to simply have export match the input format wherever possible.

In hindsight, I think matching the input format is the best approach. Consider:

- Exporting a JPEG source image as JPEG should not result in any loss of quality.
- Users retain greater control of their textures, because they can choose format and compression settings before bringing them into Blender.
- Users may not want the same format for all images. In particular, base color should almost always use JPEG on web, whereas normal maps should almost always use PNG.

For the reasons above, I'm not sure we should expose a UI checkbox for the format of all textures, and if we do, "Source format" should be the default option.

pjoe:
How is JPEG quality handled?

pjoe:
FWIW: jpg export (even from packed texs) is working for me with the feature/jpg_export branch
👍

UX3D-schmithuesen:
> How is JPEG quality handled?

We just use Blender's image saving mechanism which has some automatic handling – I guess that they have a default setting for jpg quality.

> Exporting a JPEG source image as JPEG should not result in any loss of quality.

Images run through modifications in Blender's node tree a lot so in many cases, we can not just export the same image that the user provided so it will be a newly created jpg, not just a copy of the input.

pjoe:
@UX3D-schmithuesen: ok, it does look intriguing that blender UI has a setting for quality though:

image

UX3D-becher:
@donmccurdy as @UX3D-schmithuesen said, as images can be manipulated by blenders shader node tree, we would need to specifically check if a modification in the node tree would require a new image or if the input could just be copied to the output location. I would rather not introduce such a complexity and generate new images in all cases. This however would as you already pointed out always result in a loss of quality with the approach of determining the output format from the input format. Therefore an alternative approach could be to specify the output format from the image name in blender. If an Image would for example be called "MyTexture.jpg" the output would be .jpg. This way a .png input could be converted to .jpg by the exporter without a loss in quality.

pjoe:
How would this work for combined textures (ORM)? If anyone of them is PNG then output is PNG?

I suppose that would work, until other lossless formats are added (e.g. WebP)

UX3D-becher:
@pjoe fallback would always be .png
I don't see how other formats would be a problem, they could be handled similarily

UX3D-becher:
In the most recent commit 85ef7eb the output format for combined textures such as metallic roughness is determined by the name of the metallic image

pjoe:
@UX3D-becher: my thought was that if you had ORM texs in e.g.: .wepb, .webp and .png - then .png wouldn't necessarily be what I wanted as default output format. Anyway that can be handled once there are other lossless formats - for now rule should be to always pick .png if any of the inputs are .png

PR #407, opened 2019-04-04
Fixed dict access key error when image file extension is not lower case.

atteneder:
Exporting a separate glTF failed, when a texture had a non-lower-case extension.
The conversion to lower-case string was done in the check only, not in the final dictionary acces.

Example file:
[ext_case.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3042470/ext_case.zip)


atteneder:
Yeah, seems like this whole CLA signing is not working. Tried it in 3 different browsers :/

emackey:
@atteneder Thanks!  The file you edited got overhauled a bit in #404, but it doesn't look like this concern was addressed.  Do you mind updating this PR to resolve the merge conflict?

atteneder:
@emackey sure, there you go!

emackey:
Thanks @atteneder!

PR #408, opened 2019-04-05
Keep custom properties when applying modifiers to meshes

### LINKED ISSUE #403: Custom properties on meshes aren't retained when applying modifiers
It appears that the exporter doesn't support exporting custom attributes as extras for meshes. Attaching custom properties to the containing object works fine, but properties on the mesh datablock itself seem to be ignored.

Tested in Blender 2.80 beta from a couple of days ago.

I'd be willing to try to scrape together a PR to implement support for exporting custom mesh properties.

blm768:
This should fix #403. I haven't included any new tests since the issue is so obscure, but I can put one together if you think the extra coverage would be valuable.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
@@ -243,6 +243,8 @@ def __gather_mesh(blender_object, export_settings):
             blender_mesh = blender_object.to_mesh(bpy.context.scene, True, 'PREVIEW')
         else:
             blender_mesh = blender_object.to_mesh(bpy.context.depsgraph, True)
+        for prop in blender_object.data.keys():
+            blender_mesh[prop] = blender_object.data[prop]
         skip_filter = True
 
         if export_settings[gltf2_blender_export_keys.SKINS]:


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
```diff
@@ -243,6 +243,8 @@ def __gather_mesh(blender_object, export_settings):
             blender_mesh = blender_object.to_mesh(bpy.context.scene, True, 'PREVIEW')
         else:
             blender_mesh = blender_object.to_mesh(bpy.context.depsgraph, True)
+        for prop in blender_object.data.keys():
+            blender_mesh[prop] = blender_object.data[prop]
```
UX3D-becher: this would change the state of the scene, right? If so, we should search for another way, maybe make a copy of the mesh and delete it again

donmccurdy: Related: https://github.com/KhronosGroup/glTF-Blender-IO/pull/413 by @pjoe – in that case the solution was to _avoid_ copying the object.

blm768:
Based on what I remember seeing in testing, `blender_mesh` ends up being a totally new datablock that is unlinked from the scene a few lines later, which should leave it in the same state. I haven't tested whether `to_mesh` returns the original datablock if the mesh has no modifiers, but in that case, we'd be assigning properties from the mesh back to itself, which should be a no-op.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
blm768: Based on what I remember seeing in testing, blender_mesh ends up being a totally new datablock that is unlinked from the scene a few lines later, which should leave it in the same state. I haven't tested whether to_mesh returns the original datablock if the mesh has no modifiers, but in that case, we'd be assigning properties from the mesh back to itself, which should be a no-op.

PR #410, opened 2019-04-08
Fix: Do not create texture files with delimiter characters, subsequen…

### LINKED ISSUE #405: Invalid texture file names when exporting separate glTF
The [glTF 2.0  specification](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#uris) says the relative URI paths have to be as defined by RFC 3986, Section 4.2.

This means they cannot contain delimiter character like `:` `/` `?` `#` `[` `]` `@` `!` `$` `&` `'` `(` `)` `*` `+` `,` `;` `=`.
When exporting separate glTFs, the texture file's name/path is taken from the Image Texture nodes' names, which can contain invalid characters (like `#`).

To be fair, these files will still work in most cases (like loading them in the [glTF-viewer](https://gltf-viewer.donmccurdy.com)), but I stumbled across errors when trying to post-process a glTF with [gltf-pipeline](https://github.com/AnalyticalGraphicsInc/gltf-pipeline). It fails and rightfully complains about the relative path URI

Here's a demo file and export:
[wrong_uri.tar.gz](https://github.com/KhronosGroup/glTF-Blender-IO/files/3038891/wrong_uri.tar.gz)
.

atteneder:
…t underscores or dots (other than the extension separator) in them.

Fix: Avoid image file name collisions by adding a trailing count number.

Fixes #405

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
@@ -168,7 +168,8 @@ def __get_image_data(sockets_or_slots, export_settings):
                 target_channel = 1
                 source_channels_length = 1
 
-            file_name = os.path.splitext(result.shader_node.image.name)[0]
+            file_name = __safe_uri_name(result.shader_node.image.name,export_settings)
+
             if result.shader_node.image.packed_file is None:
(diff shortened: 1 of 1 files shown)


atteneder:
solved in f8cbd0897a7b32cc13acd907285d1989f271af84
closing this one.

PR #411, opened 2019-04-08
Move & adjust image format options to be more readable

emackey:
The space given to exporter options on the UI is very limited, and it's hard to fit something readable and understandable in there.  I've attempted to change the wording on the image export options, added in #404, to be more readable on the amount of space on the UI.

Also, I moved the options to the `Mesh` tab, under `Materials`, as I don't think this option has any effect if the materials are not being exported.  If you un-check Materials, the image format option disappears.


donmccurdy:
Thanks! The tooltips append `.` at the end automatically so I removed that.

PR #412, opened 2019-04-09
Fix/animation bug

### LINKED ISSUE #322: Armature animation exports incorrectly
Hi, I have a rigged model in Blender with a pair of animations, and I'm using the GLTF exporter to export this scene to a .glb file. Sometimes the exporter has been crashing, although other times it has worked. But now I seem to have reached a state where it's crashing every time I try to export.

The bottom of the call stack reads:

File "C:\Program Files\Blender Foundation\Blender\2.79\scripts\addons\io_scene_gltf2\blender\exp\gltf2_blender_gather_animation_samplers.py". line 115, in __gather_output
bone = blender_object.path_resolve(get_target_object_path(target_datapath))
ValueError: Object.path_resolve("") could not be resolved

I downloaded the latest GLTF exporter today, but it shows the same issue.

I've attached the .blend file (which is just a character from a freebie model site with my test anim) and a screenshot of the exporter crash:
[anim_export_crash.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/2896894/anim_export_crash.zip)

This seems to be a common problem rather than an issue with just this one .blend file, as a different freebie character model I downloaded from some other author also gets the same crashes on animation export too.

Thanks for your help.

UX3D-becher:
fixes #322 and other animation issues. 

Most animation related issues should also be fixed by this pull request, they should checked again.

We needed to bake all animations involving constrained bones. This is incredibly slow, but atm can't be avoided.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_channels.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_channels.py
@@ -21,6 +21,7 @@ from io_scene_gltf2.io.com import gltf2_io_debug
 from io_scene_gltf2.blender.exp.gltf2_blender_gather_cache import cached
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_animation_samplers
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_animation_channel_target
+from io_scene_gltf2.blender.exp import gltf2_blender_get
 
 
 @cached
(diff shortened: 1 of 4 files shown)


donmccurdy:
Thanks! Tested this on a couple models (attached). It certaintly improves things, but there are still some issues on these three:

[Archive.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3060819/Archive.zip)

> We needed to bake all animations involving constrained bones. This is incredibly slow, but atm can't be avoided.

That's sufficient for constraints I think, thanks.

I'm seeing something odd when exporting the "Robot" model above with baked animation – it seems like all of the animation clips contain the same animation data, just the "Robot_Death" clip. Maybe the baking is not compatible with multiple clips right now?

PR #413, opened 2019-04-09
Fix issues with auto smooth and other modifiers

pjoe:
- Other modifiers were lost before for objects with auto smooth

- Avoid copying object, just add edge split modifier, disable auto smooth and revert back after export

Sample file:
[auto_smooth.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3059266/auto_smooth.zip)


capnm:
Didn't tested, but I guess this doesn't work for custom normals made e.g. with the bevel operator:

bevel-op-harden-normals

pjoe:
@capnm will take a look. One thing I wasn't totally sure of is whether the `edge split` modifier should be put first in the stack (currently by default it will end up last). This may also be related to custom normals PR #414

pjoe:
@capnm The bevel operator creates custom normals, so you need #414 as well for that to work :)

capnm:
> @capnm The bevel operator creates custom normals, so you need #414 as well for that to work :)

Cool, I'll check it with that PR ;)

PR #414, opened 2019-04-09
Export custom normals correctly

pjoe:
Use custom normals if available during export

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
@@ -597,7 +597,10 @@ def extract_primitives(glTF, blender_mesh, blender_vertex_groups, modifiers, exp
 
             v = convert_swizzle_location(vertex.co, export_settings)
             if blender_polygon.use_smooth:
-                n = convert_swizzle_location(vertex.normal, export_settings)
+                if blender_mesh.has_custom_normals:
+                    n = convert_swizzle_location(blender_mesh.loops[loop_index].normal, export_settings)
+                else:
+                    n = convert_swizzle_location(vertex.normal, export_settings)
                 if use_tangents:
                     t = convert_swizzle_tangent(blender_mesh.loops[loop_index].tangent, export_settings)
                     b = convert_swizzle_location(blender_mesh.loops[loop_index].bitangent, export_settings)


SteffenDuenner:
Thanks a lot for fixing this. Just gave it a quick try and it fixes everything that currently forces us to use Maya to export our stuff (coming from Blender via FBX because of the vertex normals).

pjoe:
@SteffenDuenner fantastic! Always happy when another Mayan bites the dust and converts to our beloved Blender 😄

SteffenDuenner:
Hahaaaaa! The funny thing is: I'm not even a "Mayan", nothing could be further away from truth. I'm a Softimage refugee, and also I'm a long term Blender afficionado, I was only forced to use Maya in a current project just because of the faulty vertex normal export. These times are now over thanks to you! :D

pjoe:
@SteffenDuenner Ahhh, XSI - definitely one of the best commercial 3D software. I actually also used Maya a lot, back in the day, until I took the jump to Blender full on.

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
```diff
@@ -597,7 +597,10 @@ def extract_primitives(glTF, blender_mesh, blender_vertex_groups, modifiers, exp
 
             v = convert_swizzle_location(vertex.co, export_settings)
             if blender_polygon.use_smooth:
-                n = convert_swizzle_location(vertex.normal, export_settings)
+                if blender_mesh.has_custom_normals:
```
emackey: @pjoe This is awesome, thanks!

I'm testing a strange case now, trying to reproduce this workflow: https://github.com/KhronosGroup/glTF-Blender-IO/pull/413#issuecomment-481539184

In this comment, a non-smoothed cube was beveled with "Harden Normals", producing some custom normals on a non-smooth mesh.  But it looks like your custom normal export fix here is hidden behind `if use_smooth` type check, is that intentional?  Should the custom normals take priority over the lack of smooth shading?

pjoe: Hmm, custom normals require the mesh's auto smooth to be active, and I think it also needs smooth shading to properly show up in the viewport. If you choose flat shading you are basically telling blender you want to use the face normals instead of vertex normals (or custom split normals)

pjoe: More hmmm, does seem to behave a bit weird though :S

capnm: > If you choose flat shading you are basically telling blender you want to use the face normals instead of vertex normals (or custom split normals)

You can apparently use custom normals in flat shading (left). IIRC there are plans to improve the user interface for editing custom normals. This may be useful for some effects.

image
image

pjoe: Aye, should move the custom normal code out of the `use_smooth`check then

PR #415, opened 2019-04-09
Fix animation sampling making way too many keys

pjoe:
It was generating fps keys per frame :S

Adding small sample file:

[anim_sampling.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3060856/anim_sampling.zip)

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
@@ -25,9 +25,9 @@ from io_scene_gltf2.io.com import gltf2_io_debug
 
 
 class Keyframe:
-    def __init__(self, channels: typing.Tuple[bpy.types.FCurve], time: float):
-        self.seconds = time / bpy.context.scene.render.fps
-        self.frame = time
+    def __init__(self, channels: typing.Tuple[bpy.types.FCurve], frame: float):
+        self.seconds = frame / bpy.context.scene.render.fps
+        self.frame = frame
         self.fps = bpy.context.scene.render.fps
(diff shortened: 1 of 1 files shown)


capnm:
Shouldn't that be times the frame step `*= export_settings['gltf_frame_step']` + check for zero value?

pjoe:
@capnm not sure I follow, step is the number of frames it will advance between sampling points. It is a bit misleading that the 'frame' variable in this code is called time, maybe should rename it to frame

capnm:
Sorry, you're right, LGTM :-) It was more broken than I thought.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
```diff
@@ -116,13 +116,13 @@ def gather_keyframes(blender_object_if_armature: typing.Optional[bpy.types.Objec
             pose_bone_if_armature = None
 
         # sample all frames
-        time = start
-        step = 1.0 / bpy.context.scene.render.fps
-        while time <= end:
-            key = Keyframe(channels, time)
+        frame = start_frame
+        step = export_settings['gltf_frame_step']
```
UX3D-becher: Maybe we should drop this parameter, to avoid clutter in the export settings?

pjoe: How would user then specify step value? Depending on your animation adjusting the value to control quality vs. file size is very useful IMO :)

donmccurdy: I consider it a useful setting, as baked animation can otherwise be quite large. If there's no objection other than UI clutter I'd prefer to keep it.

PR #416, opened 2019-04-09
For meshes use location, rotation_quaternion and scale to export...

### LINKED ISSUE #409: Exporting rotation doesn't work when at least one scale dimension ist 0
This is related to #302 but the core issue hasn't been fixed so I'm opening a new issue.

I need to export an empty object that has scale 0 along the z-axis. This certainly isn't a common use case but I still think it will come up from time to time.

After a little investigation I found that currently translation, rotation, scale is extracted by decomposing the object's matrix:
https://github.com/KhronosGroup/glTF-Blender-IO/blob/d8aa684b87779d43ead627a6463ce2b6d7763f0b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py#L106

This of course doesn't work when some axis is scaled to 0. Blender provides access to it's objects' properties `position`, `rotation_quaternion` and `scale` directly. Is there a specific reason why these aren't used?

The above mentioned function is used in only two places.

The first one is where I suggest we use the object's properties directly (modified by export settings of course):
https://github.com/KhronosGroup/glTF-Blender-IO/blob/b44e4794490f7d526727cf40f2fe20097f961385/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py#L278

and the second one should be left unchanged:
https://github.com/KhronosGroup/glTF-Blender-IO/blob/90b0c1f640c6d1105f9f520d0bf230b3cf8fe87c/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_joints.py#L50

I'm happy to do a pull request. Please let met know.

Thanks.

maccesch:
… instead of deconstructing matrix_local.

Fixes #409.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
@@ -102,13 +102,8 @@ def convert_swizzle_scale(scale, export_settings):
         return Vector((scale[0], scale[1], scale[2]))
 
 
-def decompose_transition(matrix, context, export_settings):
+def decompose_transition(matrix, export_settings):
     translation, rotation, scale = matrix.decompose()
-    """Decompose a matrix depending if it is associated to a joint or node."""
-    if context == 'NODE':
-        translation = convert_swizzle_location(translation, export_settings)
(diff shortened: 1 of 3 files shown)


donmccurdy:
Thanks!

Issue #369, opened 2019-03-20
Resampled animations should use linear interpolation


donmccurdy:
When sampling is required (either for a technical reason or because the *Always Sample Animations* option is checked), the exporter currently forces STEP interpolation:

https://github.com/KhronosGroup/glTF-Blender-IO/blob/287b4af31313725a0c1197cdfda413b63cae3a15/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_samplers.py#L79-L84

I think it would be better to use LINEAR interpolation here for three reasons:

1. Animation is typically sampled at 24fps, but may be played back at 60+ FPS.
2. Once the *Sampling Rate* option is supported (is it doing anything now?), a user may want to reduce the sampling rate to every N frames, and still have interpolation across those frames.
3. Using CUBICSPLINE for sampled keyframes probably doesn't make sense.

emackey:
100% agree on LINEAR interpolation.  Some systems (particularly Cesium) may play back animations at a much faster or slower rate than they were originally sampled with, and using LINEAR smooths out these motions very nicely.

julienduroure:
Seems this is already commited

PR #417, opened 2019-04-10
Use LINEAR interpolation when sampling anims

pjoe:
Not sure why this ever used STEP :)

With LINEAR you can get decent result with sampling every few frames (e.g. 3-4 depending on your animation), instead of the chunky result you will get with STEP

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_samplers.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_samplers.py
@@ -89,7 +89,9 @@ def __gather_interpolation(channels: typing.Tuple[bpy.types.FCurve],
     if gltf2_blender_gather_animation_sampler_keyframes.needs_baking(blender_object_if_armature,
                                                                      channels,
                                                                      export_settings):
-        return 'LINEAR'
+        min_keyframes = min([len(ch.keyframe_points) for ch in channels])
+        # If only single keyframe revert to STEP
+        return 'STEP' if min_keyframes < 2 else 'LINEAR'
 
     blender_keyframe = channels[0].keyframe_points[0]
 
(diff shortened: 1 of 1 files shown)


pjoe:
Sample file, with sampling every 4 frames:

[anim_sampling.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3063001/anim_sampling.zip)

Tried capturing the difference in gifs

STEP:
step

LINEAR:
linear

pjoe:
@UX3D-becher OK, I understand the issue, but for end user this really should be handled automatically. I'll try to see how difficult it is to handle the case with only 1 keyframe.

Having to manually bake anims is a complication of the workflow, and also to some extend makes the export process less 'non-destructive'.

pjoe:
@UX3D-becher something like this should do it 92c50f7d32f34ebf891c9bb90d4a566e2b010a54

emackey:
@UX3D-becher I find that reasoning a little odd.  Wouldn't it be better to file bug reports against engines that can't handle an individual linear keyframe?

I guess having a workaround here doesn't hurt much, but thus far the pattern has been to insist on filing issues and fixing problems rather than bending the format spec to accommodate bugs.

emackey:
Fixes #369

donmccurdy:
Agreed that LINEAR interpolation should be used for baking whenever possible. Engines are not rendering at 24-30 fps, which the keyframes usually are. Whether single-keyframe tracks should use STEP or LINEAR I'm not sure.

donmccurdy:
Hm, but this will go to STEP for all channels, if any channel has a single keyframe? I might prefer LINEAR in that case...

emackey:
Ouch.  Would definitely prefer LINEAR in that case.

pjoe:
I am all for always using LINEAR and would consider it a weird edge case / bug if engines don't handle it correctly. So I can easily revert last commit if we agree on this? Or try to add further detection of single channel (won't there always be 3 or 4 together?) Anyway this PR improves behavior in most common cases (more then 1 key), so can also be merged as is and we can further tackle the single key/STEP issue in separate PR

pjoe:
After thinking/testing this some more:

I believe the best option is to use LINEAR if any of the channels in a sampler has more than 1 keyframe,
since loc/scale/rot will get each their sampler and glTF will batch together each subcomponent, i.e. glTF doesn't support having different interpolation (or number of keys) for e.g. loc.x and loc.y anyway.

If anyone can come up with a case where this would break - speak up 😄 

Otherwise I think it is good to go now

emackey:
Any objections @UX3D-becher @donmccurdy ?

donmccurdy:
Sounds good to me. I could imagine CUBICSPLINE causing problems with a single keyframe, but STEP and LINEAR seem completely reasonable. If there's a bug in e.g. threejs we can certainly fix those.

PR #421, opened 2019-04-11
Bump version to 0.9.0, write version number to output file.

### LINKED ISSUE #182: Include exporter version in 'generator' string.
Will likely be helpful down the road if assets include the version of the exporter used. I.e. replace `Khronos Blender glTF 2.0 exporter` with `Khronos Blender glTF 2.0 exporter (<version>)`, where `<version>` is either a commit hash or semver version.

emackey:
This PR bumps the version number to `1.0.0`, and I hope to establish a new process where we increment the `PATCH` version (the third number) in the master branch after any PR or group of PRs gets merged.

This PR also includes new code to write the version into the glTF `generator` string.

@donmccurdy Note that I've completely changed the signature of the string, so you may have to update the list of known generators.  The previous string included the glTF version in the generator string, which was redundant because the glTF version is already required to be listed in the version field of the same asset object.

The new addon panel looks something like this:

AboutAddon

And the new output looks like this:

```
    "asset" : {
        "generator" : "Khronos glTF Blender I/O v1.0.0",
        "version" : "2.0"
    },
```

Fixes #182.

/cc @julienduroure @UX3D-nopper

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -15,7 +15,7 @@
 bl_info = {
     'name': 'glTF 2.0 format',
     'author': 'Julien Duroure, Norbert Nopper, Urs Hanselmann, Moritz Becher, Benjamin Schmithüsen, Jim Eckerlein, and many external contributors',
-    "version": (1, 0, 0),
+    "version": (0, 9, 0),
     'blender': (2, 80, 0),
     'location': 'File > Import-Export',
     'description': 'Import-Export as glTF 2.0',
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -14,8 +14,8 @@
 
 bl_info = {
     'name': 'glTF 2.0 format',
-    'author': 'Julien Duroure, Norbert Nopper, Urs Hanselmann, Moritz Becher, Benjamin Schmithüsen, Jim Eckerlein',
-    "version": (0, 0, 1),
+    'author': 'Julien Duroure, Norbert Nopper, Urs Hanselmann, Moritz Becher, Benjamin Schmithüsen, Jim Eckerlein, and many external contributors',
+    "version": (1, 0, 0),
     'blender': (2, 80, 0),
     'location': 'File > Import-Export',
     'description': 'Import-Export as glTF 2.0',
@@ -26,6 +26,8 @@ bl_info = {
     'category': 'Import-Export',
 }
 
+def get_version_string():
+    return str(bl_info['version'][0]) + '.' + str(bl_info['version'][1]) + '.' + str(bl_info['version'][2])
 
 #
 # Script reloading (if the user calls 'Reload Scripts' from Blender)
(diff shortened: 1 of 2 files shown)


donmccurdy:
> I hope to establish a new process where we increment the PATCH version (the third number) in the master branch after any PR or group of PRs gets merged.

Could this be automated and associated with git tags?

emackey:
> Could this be automated and associated with git tags?

That would be great!  I don't know how to implement that one though.

donmccurdy:
A find/replace in `convert_to_strict_2.80.py` could bump `__init__.py` automatically. @julienduroure would it be reasonable for the script to also create a new git commit and tag, or does that seem too complicated?

julienduroure:
Think this is doable. I will work on it.
I'm not sure we should add a tag each time I merge into blender repo, because I'm currently doing it 2 or 3 times a week. This will be a lot of tags !
But definitively, I will increment version & commit it

donmccurdy:
Thank you! Glad to help if I can. Agreed that tags aren't necessary quite that often. :)

julienduroure:
Based on this blog post: https://code.blender.org/2019/04/2-80-release-plan/
2.80 release should land in July. 

Not sure we have to bump to 1.0.0 now, or if we have to wait until release. Maybe we can now bump to 0.1.0 (and automatically increase to 0.1.1 ... 0.1.x )

emackey:
Let's bump this to 1.0.0 today, and immediately begin incrementing the patch versions (1.0.1, 1.0.2, etc).  I don't think there's anything to be gained by having 1.0.0 be a grand release tied to a large Blender version.  In my mind it's far more important to start being able to look at sample output and easily tell which PRs were merged when a particular model was exported.  It's fine if our version 1.0.0 goes un-noticed, as Blender 2.80 will attract tons of attention.

Looking at other exporter version numbers, they seem to be completely independent of Blender release versions.  Also, from a "people who judge the number of bugs by the number of zeros in the version number" standpoint, it may make the addon appear more stable, or at least have a longer history, when it has some nonzero version numbers.

But primarily, having this number change with each PR or each daily build will allow us to track down whether a particular PR really fixed a particular bug, and we need that.

titansoftime:
Does this mean that this is no longer in beta? Is skinning exporting correctly now?

donmccurdy:
Many skinning issues were fixed in the past couple weeks, so it would be worth testing your models again to see what issues remain.

That said, the beta tag and version numbers are not synced to any particular issues, we are just ensuring that each release of the package not _has_ a unique version number so that we can debug issues from the upcoming Blender release. I expect that the beta messaging will be removed when Blender 2.8 itself leaves beta.

emackey:
I guess if we're keeping the Beta label, we can't put 1.0.0 yet then.  Should I update this to 0.9.0?  I'd rather that than just 0.1.0.

UX3D-nopper:
I prefer 0.9.0, as 0.1.0 looks like we just started.

emackey:
OK, the version number on this branch has been rolled back to 0.9.0.

julienduroure:
I committed a first change to add option to +1 minor version of addon. Currenlty, there is no commit & push done automatically, this is the next step (will work on it in following days)

emackey:
Looks like the version bump script went directly into master, so I'm going to merge this now.

PR #437, opened 2019-04-16
Roundtrip AO node

### LINKED ISSUE #123: Supporting occlusionTexture in Blender 2.8
I'm struggling to decide how to handle AO (occlusionTexture) in Blender 2.8. Overall the move to Principled BSDF is great, but it has no AO input.

One option would be to adopt a convention such that a disconnected `Image Texture` node, used in the same material as a Principled BSDF node, will automatically be treated as AO if (a) the node is named "AO", or (b) the node uses a texture suffixed with "_ao". Presumably the Importer and Exporter could use the same convention.

Other ideas? In general I don't want to hack glTF features into this exporter when Blender doesn't support them, but AO is a pretty critical feature...

### LINKED ISSUE #319: Combining Occlusion with Metallic/Roughness makes duplicate references to the image
Copied from @j-conrad in https://github.com/KhronosGroup/glTF-Blender-IO/pull/295#issuecomment-465385119

---

Despite creating a single 'ORM' (occlusion, roughness, metallic) texture and wiring the RGB up to a Node Group (with 'Occlusion'), Roughness, and Metallic respectively, the exported glTF produces two separate textures-- 1 for Occlusion and 1 for Roughness/Metallic. Not sure why it's creating an extra copy of the same texture but I assume this is bumping up the file size.
image

image

### LINKED ISSUE #321: Import occlusion map
@julienduroure I'm curious if you would be willing to modify the importer to preserve an occlusion map using an empty custom node.  For example, when occlusion is combined with metal/rough, perhaps the importer could produce something similar to this:

image

I think I've worked out the code for creating and using a node group like this.  This is what I have so far:

```
    # Put this definition somewhere common, it will likely change to "glTF Settings" or similar.
    gltf_node_group_name = 'glTF Metallic Roughness'

    # The following code runs per-material for materials with occlusion.

    # Check if the custom node group already exists
    if gltf_node_group_name in bpy.data.node_groups:
        gltf_node_group = bpy.data.node_groups[gltf_node_group_name]
    else:
        # Create a new node grouping
        gltf_node_group = bpy.data.node_groups.new(gltf_node_group_name, 'ShaderNodeTree')
        gltf_node_group.inputs.new("NodeSocketFloat", "Occlusion")
        gltf_node_group.nodes.new('NodeGroupOutput')
        gltf_group_input = gltf_node_group.nodes.new('NodeGroupInput')
        gltf_group_input.location = -200, 0

    # Place a new node grouping in the current material
    gltf_settings_node = bpy.data.materials[material_idx].node_tree.nodes.new('ShaderNodeGroup')

    # Set the newly placed node grouping to be the glTF grouping
    gltf_settings_node.node_tree = gltf_node_group
    gltf_settings_node.width = 220
```

From there you would connect the `R` channel of the occlusion image texture to the one socket on the new custom node.

emackey:
Fixes #123
Fixes #319
Fixes #321

This includes commits from @julienduroure from the `import_ao_node` branch.  On that branch, a minimal node group is created to contain the occlusion map.  On this `roundtrip_ao_node` branch, the new group is accepted as a means to export occlusion, and the material exporter specifically looks for an opportunity to combine occlusion, roughness, and metallic into a single image.

**TODO:**
- [x] ~~Fix tests~~
- [x] ~~Fix/remove remaining TODO in comments~~

/cc @donmccurdy


addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py
```diff
@@ -178,6 +178,10 @@ def pre_compute(gltf):
                     if 'glossinessFactor' not in material.extensions['KHR_materials_pbrSpecularGlossiness'].keys():
                         material.extensions['KHR_materials_pbrSpecularGlossiness']['glossinessFactor'] = 1.0
 
+        # images
+        for img in gltf.data.images:
```
emackey: > Error: Command failed: blender279b -b --addons io_scene_gltf2 -noaudio --python roundtrip_gltf.py -- roundtrip/01_color_attribute/01_color_attribute.gltf /out/roundtrip/01_color_attribute/outblender279b 
2019-04-16

> File "gltf2_blender_gltf.py", line 182, in pre_compute for img in gltf.data.images:
> TypeError: 'NoneType' object is not iterable

addons/io_scene_gltf2/blender/imp/gltf2_blender_map_occlusion.py
```diff
@@ -32,12 +33,75 @@ def create_nodetree(gltf, material_idx, vertex_color):
...
+
+        if found == False:
+            # Need to create the texture node & separate node
+            mapping = node_tree.nodes.new('ShaderNodeMapping')
+            uvmap = node_tree.nodes.new('ShaderNodeUVMap')
+            separate = node_tree.nodes.new('ShaderNodeSeparateRGB')
+            if pymaterial.occlusion_texture.tex_coord is not None:
+                uvmap["gltf2_texcoord"] = pymaterial.occlusion_texture.tex_coord  # Set custom flag to retrieve TexCoord
+            else:
+                uvmap["gltf2_texcoord"] = 0  # TODO set in pre_compute instead of here
```
emackey: Remove/resolve TODO?

emackey:
Behold the WaterBottle sample model, full ORM stack imports and exports on this roundtrip branch.

WaterBottleORM

emackey:
For anyone testing the exporter on this branch, here's a project with three grayscale images that combine to form a single Occlusion-Roughness-Metallic texture.  The combined image has a cyan A, magenta B, and yellow C.

[TestMultiChannelORM.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3086404/TestMultiChannelORM.zip)

addons/io_scene_gltf2/blender/imp/gltf2_blender_map_occlusion.py
julienduroure: This is a global TODO (I mean that this is not the only line of code with this TODO about initialisation of texture coordinate. I think we can merge with this TODO, and it will be removed when I will process it on entire source code of the addon

addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py
julienduroure: I committed a fix in the roundtrip_ao_node branch

addons/io_scene_gltf2/blender/imp/gltf2_blender_map_occlusion.py
```diff
@@ -32,12 +33,75 @@ def create_nodetree(gltf, material_idx, vertex_color):
         """Nodetree creation."""
         pymaterial = gltf.data.materials[material_idx]
 
+        material = bpy.data.materials[pymaterial.blender_material[vertex_color]]
+        node_tree = material.node_tree
+
         BlenderTextureInfo.create(gltf, pymaterial.occlusion_texture.index)
 
-        # Pack texture, but doesn't use it for now. Occlusion is calculated from Cycles.
+        # Pack texture. Occlusion is calculated from Cycles or Eevee, so do nothing
```
donmccurdy: Minor, but I didn't think you could bake occlusion in Eevee yet?

donmccurdy: Oh, or this means that rendering in Blender will calculate AO one way or the other?

donmccurdy:
Haven't had the time to test it yet, but this looks great – thank you! :)

addons/io_scene_gltf2/blender/imp/gltf2_blender_map_occlusion.py
julienduroure: Yes, I mean that AO is not provided by user, but the rendering engine process leads to an AO calculation. (And you're right, we can't bake it)

emackey:
Yes, I'm planning to change a number of related things in the docs, as a separate PR.

PR #444, opened 2019-04-19
Combine names when combining different images into one.

emackey:
When more than one image is used to build a texture (for example, metal + rough as separate images), the names of the source images need to be combined.

The exporter was doing this [until recently](https://github.com/KhronosGroup/glTF-Blender-IO/pull/404/files#diff-9f6a92be46d665dccef33199e2884a07L50) when a refactor dropped the feature.  My reimplementation here is a bit more complex, unfortunately, but it tries to "do the right thing" on a bunch of edge cases.

Time permitting, I may try to add tests for this behavior, so it can be called out if it breaks again.


emackey:
The latest commit to this branch includes the first few entries in a whole new section of the test suite.  Prior to this, the tests just confirm that valid glTFs are being produced, but this change introduces a new section of tests where we start to inspect certain aspects of the results.  This allows the tests to determine if filenames are being combined correctly, and ORM images are being assembled correctly.

There's tons of room to build additional specific tests, but for this PR I'll pause here and see if this new pattern of testing is acceptable to folks here.

pjoe:
Just to make sure I understand the principle: it checks the resulting texture files and their content, but doesn't try to do any validation of the json content?

emackey:
Wow.  Spent more than a day trying to get those image tests done, and when they finally worked I got excited and totally forgot to check the JSON output.

In my latest commit I'm using `JSON.parse` on the exported content, and then expecting certain values to be set correctly within that content.

As always, the glTF Validator also runs on the files, as part of the existing test suite, but you knew that.

pjoe:
Cool, LGTM

Maybe we can discuss some general solution to do comparison of JSON content (in separate issue or slack or something).

Btw. looks like Circle has changed something, I can't see uploaded test results or artifacts anymore

emackey:
Yes, the requirement to be logged into CircleCI in order to see artifacts threw me for a loop last week.

emackey:
Tests failed when I merged this, but it looks like the failure was during the "download Blender 2.80" step.  Hopefully that's just transient, maybe there's a corrupt bzip file on the server today...

pjoe:
From the logs transfer was interrupted after 22M. Trying same URL from here completes just fine, so maybe just some network glitch. If you have access to retry the build, that may fix things. And once Blender 2.80 goes final we should do a docker image with blender already installed and use that for tests :)

emackey:
Yes.  It looks to me like Blender 2.79b is [also being downloaded](https://github.com/KhronosGroup/glTF-Blender-IO/blob/efa4205618f071fcbdff3ccda5082319384c1dfb/.circleci/config.yml#L27) each time.  Would love to get that fixed too, but I'd have a lot of reading to do before I tried it myself.

pjoe:
Yeah I ended up making the build script download every time as I don't have access to setup Khronos properly in docker hub, and there didn't seem to be others that had or knew how to do. If someone has the rights I'll be happy to assist in setting it up. I originally had it (with our fork) so that it would trigger new build on docker hub on every push to master. In that way the build script would have the Blender 2.80 version from last push.

Issue #436, opened 2019-04-16
Custom normals on flat-shaded faces


emackey:
Just creating an issue so we don't forget.

In https://github.com/KhronosGroup/glTF-Blender-IO/pull/414#discussion_r274018041 I noticed a workflow involving bevel with hardened edges that applied custom normals on top of flat-shaded faces (as opposed to customized smooth-shaded faces).  In https://github.com/KhronosGroup/glTF-Blender-IO/pull/414#discussion_r274306611, @pjoe suggested changing the order of checks in the nested `if` statements.

pjoe:
Will try taking a look after Easter 🥚

Issue #399, opened 2019-04-01
Problem when exporting animated model. It seems like armature bones axis are changed.


cognatron:
Hi, 
I'm can't export this model with animations attached because the armature bones seems to be exported with wrong axis. It seems that the problem is related to the "children" bones and not with the "parent" base-bone.

*Model exported without skinning:*

![Exported model without skinning](https://user-images.githubusercontent.com/43038761/55341572-47936d00-547d-11e9-855c-936cb08d68d2.jpg)

*Model exported with skinning:*

![Exported model with skinning](https://user-images.githubusercontent.com/43038761/55341574-482c0380-547d-11e9-84f8-42af7767d9c4.jpg)

This is the animation working properly:
https://skfb.ly/6IrSJ

The .blend file is attached here.
[TrompaBeta-2.8_demo.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3030302/TrompaBeta-2.8_demo.blend.zip)

simongloor:
@cognatron, do your bones look correct when displaying the skeleton? (I suppose you used https://gltf-viewer.donmccurdy.com/ )

I've got the same problem with my model, both when I export it with animations and with only skinning. The bones look fine and animate properly, the skinning is off:
![image](https://user-images.githubusercontent.com/43432178/55384611-749e5900-552b-11e9-8786-9a1fd89bf168.png)

(The model should work as it is fully implemented in my game, using the fbx-export)

simongloor:
I think this might be related to #296
After updating to the latest version the skinning looks fine. Now I only need to fix the animations, but that's a different topic ;)

cognatron:
> I think this might be related to #296
> After updating to the latest version the skinning looks fine. Now I only need to fix the animations, but that's a different topic ;)

Hey, now it works with the new version! 
The only thing I must check now is how to put to work together keyshapes-based animations along with armature actions.

cognatron:
@UX3D-nopper The exported armature-based animations are broken again since the latest exporter updates.

cognatron:
(When using the same .blend file attached.)

UX3D-nopper:
Can you please track down, which commit did break this? If yes, please assign the bug to the person.

cognatron:
@UX3D-nopper I will check that. It looks like the problem is about the "child" bones of the armature.

emackey:
@cognatron Actually could you test the branch in #446 to see if it fixes the issue?

[...]

PR #446, opened 2019-04-23
Fix stuff

Cyp:
Tried updating, everything broke. This fixes it (or at least what I noticed so far, except for some weird dark surfaces which may be due to #436).

As a side effect, 09b1270b834364f0a0b4c59647b86122345a5144 might help with some bug reports about bad performance and/or large filesizes, but haven't tested or read the bug reports in detail. And 4549b3dadb48c5e88c3ba27d3c0e2043f6dc65d9 isn't really a fix.


emackey:
@Cyp Did you see #444?  😄

Cyp:
Yes, although I didn't look very carefully. If I understand right, it fixes something related, but a separate issue. There's no merge conflicts between our branches.

emackey:
@donmccurdy This has some animation fixes, are you available to take a look?

donmccurdy:
I'll do some testing here.

@pjoe would you be able to check this against some of your use cases as well?

Once things are working fully, I definitely think it's going to be critical that we get some automated tests to cover these cases as well.

pjoe:
Guess we need those 'golden file' tests sooner rather than later :) Would raise confidence about changes to fundamental things like how quaternion swizzling is handled. Anyway I'll try taking a closer look tomorrow, getting pretty late here. One thing to remember is to test both with/without Y_UP conversion.

emackey:
> it's going to be critical that we get some automated tests

Agreed!

> Guess we need those 'golden file' tests sooner rather than later

Golden file tests won't work, because the JSON in the sub-objects (under the sorted root level) is non-sorted, and is different with every export.  Even the bin files come out ordered differently, not sure why. 
Plus, any change to the overall structure will break every golden file test.

I'm thinking we could use regex patterns to locate specific things we need to see in the test output.  Not sure how we can carefully test for animation breakage like this.

@pjoe if you have a minute, please check out my new ORM tests in #444, I'm curious if you think this is a good pattern before I write more.

pjoe:
@emackey you are of course right :) I was more thinking along the line of some good json diff algorithm (js deep equal or something, maybe skipping parts that aren't relevant) But if even ordering of indexed entries and binary data changes, that will be difficult to make stable. We do need some level of automated test of exported json/binary content though 🤔

emackey:
Can anyone familiar with animation exports review this?  @donmccurdy or someone on the UX3D team?

I do have some ideas about testing animation export, but I'd like to get this PR resolved first.

pjoe:
Code changes look pretty OK to my naked eye 👁 

But this is complex enough that I won't rely on that alone, will try with some test files

pjoe:
Hmmm my test file is now broken even on master (and also with this PR), will have to take a closer look tomorrow.

pjoe:
Not sure how things broke :S ... Haven't had time to do a full bisect, but last working version I found is de31c4f

Any way maybe best to get this PR merged, and then fix afterwards. Obviously we very much need proper regression tests for avoiding these things. I will try and make a test file that I can share publicly.

emackey:
@pjoe Thanks.  For what it's worth, I did a full `git bisect` on another animation issue #399 (which is not fixed here).  The commit that broke that one was fa6e4afffe4c424.

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
```diff
@@ -105,9 +105,6 @@ def convert_swizzle_scale(scale, export_settings):
 def decompose_transition(matrix, export_settings):
     translation, rotation, scale = matrix.decompose()
 
-    # Put w at the end.
-    rotation = Quaternion((rotation[1], rotation[2], rotation[3], rotation[0]))
```
emackey: What are the implications of removing this?  Are we sure it's the right thing to do?

pjoe:
FWIW: After some more bisecting, it's 588a4749426edc39ce2ab0acf41c1bef74285ab3 and c6d1638183540d502472992a9930998e6e9c009f that really messes up my test file :S

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
Cyp: It was needed for consistency. The `convert_swizzle_rotation` function requires the quaternion to not have the components mixed. Also, it makes more sense in general to not change the order of the components until when converting from a quaternion to a JSON array, since changing the order of the components breaks quaternion multiplication (multiplication by 1 = `Quaternion((1, 0, 0, 0))` is the identity, and multiplication by k = `Quaternion((0, 0, 0, 1))` isn't).

emackey:
Thanks all.  Don said he needed a couple more days to get time to review this.

donmccurdy:
@emackey @pjoe OK to merge this?

pjoe:
@donmccurdy yeah let's get this merged and then work on fixes and adding regression tests after that

donmccurdy:
Thanks @Cyp!

emackey:
Thanks @Cyp.

PR #450, opened 2019-04-29
Rough without metal, and metal without rough

### LINKED ISSUE #240: Blender does not export the roughness map, instead it sets the metallic or roughness value to 1 if not both used
When I connect an image to the Metallic or Roughness input of the Principled shader, the image isn't exported, instead it just sets the value where it was connected to 1. But if I connect the image to the Metallic and Roughness slot the image is exported and a Separate RGBA node is added before it's connected to the Principled inputs.

Test file:
[roughness bug.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/2779401/roughness.bug.zip)

emackey:
This adds some fixes for when roughness exists without metallic, and vice-versa.  In such an arrangement, the "missing" channel will be filled with white pixels, but the `roughnessFactor` or `metallicFactor` will reflect the artist's choice of value on the node.

Tests have been updated to try seven different arrangements, all from separate grayscale images:

- only roughness
- only metallic
- roughness + metallic
- occlusion
- occlusion + roughness
- occlusion + metallic
- occlusion + roughness + metallic

In each case, the tests expect the missing R/M channel(s) to be specified with factors, and the existing channels to be supplying texture data with no factors.

Additionally, a "roundtrip" test checks what happens when all three channels come from the same image.

Fixes #240.


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
```diff
@@ -175,15 +175,28 @@ def __gather_orm_texture(blender_material, export_settings):
...
+    else:
+        result = (occlusion, roughness_socket, metallic_socket)
+
+    # Double-check this will past the filter in texture_info (otherwise there are different resolutions or other problems).
+    info = gltf2_blender_gather_texture_info.gather_texture_info(result, export_settings)
+    if info is None:
+        return None
 
-    return (occlusion, roughness_socket, metallic_socket)
+    return result
```
donmccurdy: This method could return either `(o, r)` or `(o, m)` among other options. I can't see where/how we later resolve whether that second entry should be written to the blue or green channel. Should we be returning `(o, None, m)` or something that matches up with the channels like that?

emackey: This is due to architecture that was already in place before my changes.  The channel assignments are (re-) computed in the [gather_image](https://github.com/KhronosGroup/glTF-Blender-IO/blob/d2405aedb0b5ecd5f65fd6cb851131e3d772e017/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py#L166) helper.  If `None` shows up anywhere in this result tuple, the entire image will be filtered out (by several different filter routines, sprinkled around).

With current architecture, `None` is forbidden here, and order is irrelevant.  The channel assignments come from inspecting the node socket connections.

emackey: This is why, in `master` currently, if either roughness or metallic is left disconnected, the entire RM map will be missing.  It gets filtered out.

donmccurdy: Ok, thank you! This code looks good to me, then.

PR #453, opened 2019-05-01
Add an "Image URI handling" option, for using original image file paths as the URIs

yoanlcq:
Greetings,

I have been working on several Blender scenes which share texture images. However, when exporting to `GLTF separate`, the image files are copied to the output directory, using their name from Blender.
This introduces a lot of redundant files when you have, say, multiple assets, which possibly use different output directories and different texture names, even though they refer to the same source image files.

In this case, I'd rather have the exporter reuse the original image paths, directly as the URIs in the JSON output (also removing the need to re-export these images); this allows to have a single source of truth as long as the images do not require conversion (i.e are JPG or PNG, as I've assumed so far).

This PR implements it as an extra option to the exporter; the current behavior remains the default.

The parts where I have some doubts are the way I retrieve the "final filepath" for the URI (paths in Blender may start with `//`, but is that all there is to it?) and the fact that sockets may specify multiple textures (is it OK to pick the first one?)

Thanks in advance!


donmccurdy:
I'm feeling cautious about this change, it offers users a potential footgun that could make their assets much less portable.

In general, this is the sort of change feature I'd love to see added to a processing tool like glTF-Pipeline, glTF-Toolkit, or glTF-Transform as a general-purpose "bundling" step, that resolves shared resources:

```
gltf-pipeline bundle -i a.glb b.glb c.glb -o bundled/
```

^That doesn't exist today, and I don't know what the process would look like to have developers contribute new pipeline steps.

emackey:
I think I would agree with @donmccurdy that this is a good job for some kind of script or pipeline outside of this exporter.

Some glTF readers won't take URIs that reach into parent folders, for security reasons.  So offering this option to users could reduce compatibility on that front.

Another problem is that sometimes the exporter has to re-map the channels of an image.  For example if roughness & metallic images aren't connected to the Principled BSDF node exactly the way glTF defines them, the addon will generate a new image with the correct mapping.  Handling this correctly will produce images localized to each model, so an external pipeline will have to look for duplicates after all of the exports have completed.

yoanlcq:
Thanks for your feedback!

> it offers users a potential footgun that could make their assets much less portable.

How so? Isn't it fine as long as all required files are provided to the application? Sure, it makes publishing a separate GLTF asset a bit more involved than "just copy its directory", but not that much.  
In any case, this is optional; I would expect users to understand the implications of having relative file paths.

> In general, this is the sort of change feature I'd love to see added to a processing tool

I would argue that, for the scope of this specific problem, there's not even a need to write such a tool.  

File paths are the true unique identifiers for files; Blender has that information, and so does the exporter. After export, it's too late... (unless we store it in `extras`, but this doesn't prevent images from being copied).  

When we are in situation where...
- We know our textures are PNG/JPG;
- We understand the consequences of having relative file path URIs and are OK with it;
- We know we'll comply to GLTF semantics when mapping image channels;

... Couldn't we keep file paths to these textures and use them as-is ? Preventing, to the maximum extent possible:
- Having many copies of the same 4K texture in the filesystem, in every export directory, possibly with different names;
- Having a tool not only figure out which images are the same, but also re-export them _yet again_, and having to figure out how to merge multiple GLTFs, also merging all references to identical textures.

The answer is yes, and I wish to provide it as an optional setting. Nobody forces anyone to use it, but it's there, because as the exporter, we are the ones who know file paths and know the mapping from these to the final URIs. We're also the ones who make the decision to copy images!

All in all, the exporter is, to me, the best place to solve this problem. Bundling multiple GLTFs assets is another, bigger, different problem (and therefore requires a different solution).

> Some glTF readers won't take URIs that reach into parent folders, for security reasons.

That's noteworthy! The tooltip could be changed to warn about this. Users can still decide for themselves if they are OK with this.

> Another problem is that sometimes the exporter has to re-map the channels of an image. For example if roughness & metallic images aren't connected to the Principled BSDF node exactly the way glTF defines them, the addon will generate a new image with the correct mapping.

That is the behaviour we want even with my option on. Its intent is "try to point to the original file paths if they can be used as-is, on a case-by-case basis"; any file for which this is not true should be re-exported along with the GLTF, using the correct format, as usual.

We might also emit a warning... In fact we probably should emit a warning for every image which required conversion/re-export.

yoanlcq:
In any case, this made me aware of some things my PR was missing :

First, it currently assumes that the export directory is the same as the `.blend` directory, but it shouldn't, so there's a path conversion step to do.

Also, its "file can be used as-is" check is currently just "is it a PNG or JPG ?", but it should also be "are its channels mapped correctly to the Principled BSDF node?".

There are also some notes to add in tooltips, so users are informed of possible issues (such as parent folder URIs and portability).

emackey:
Thanks @yoanlcq.  I'm going to close this as there appear to be design issues remaining to be worked out.  Feel free to open a new PR when ready.

scurest:
Thought I would mentions this has become easier. In particular,

> it should also be "are its channels mapped correctly to the Principled BSDF node?".

This can now be tested with `export_image.blender_image()`, which returns a bpy.types.Image when there's an existing image whose channels are all mapped correctly.

Here's a [branch](https://github.com/KhronosGroup/glTF-Blender-IO/compare/6786505...scurest:write-original-image-paths) with an up-to-date implementation.

donmccurdy:
> In general, this is the sort of change feature I'd love to see added to a processing tool....

Heh, I'd forgotten I made this comment, but I've since written a bundler to do something very much like that:

Screen Shot 2020-07-31 at 5 13 20 PM

scurest:
@donmccurdy What is it doing? If the exporter doesn't write the original image paths, there's no way to get them back in a post-processing step (though you can go the other way).

donmccurdy:
Ah I see. If the GLBs have duplicate textures you could consolidate them:

```
# merge glTF files
gltf-transform merge a.glb b.glb c.glb output/output.gltf

# merge duplicate textures and accessors
gltf-transform dedup output/output.gltf output/output.gltf
```

So the output file would combine the final assets into one and ensure they reuse the same textures. But that's not really the workflow being requested here, I realize.

I guess I need to think about this one a bit. It feels like https://github.com/KhronosGroup/glTF-Blender-IO/pull/836 has arguably solved this workflow — if you select the same output texture folder for each .blend, and that's also the folder that sources the textures, nothing should be overwritten, correct?

scurest:
> if you select the same output texture folder for each .blend, and that's also the folder that sources the textures

Well the textures don't have to all come from the same folder. (Also there's #743.)

donmccurdy:
I'm not sure we need to support _exactly_ the workflow that's being asked for here. I see a lot of models (especially DAE, OBJ/MTL) with absolute paths from the artist's hard drive, and "Original Paths" feel like a slippery slope toward that... it would be enough — in my opinion — to have a workflow that just allows the user to export glTF file from two different `.blend` sources and have some confidence that they'll point to the same textures.  I think #836 provides that, but fixing #743 would improve the situation for sure.

scurest:
Hm, I look somewhat warily on that slope too. It is strange that it's the default behavior of the other exporters but glTF can't do it at all though.

We could only allow it when exporting in Separate mode? That would keep .glb and embedded always standalone.

(I don't really like user extensions, but it also occurs to me this could be done with one.)

donmccurdy:
> We could only allow it when exporting in Separate mode? That would keep .glb and embedded always standalone.

Yeah, I think this should be the minimum... @emackey thoughts?

scurest:
Let's also ask @yoanlcq if that covers his use case.

yoanlcq:
Thanks for asking! I'll try to describe my use case as best as I can here, and address your suggestions.

First of all, I confirm that my (suggested) workflow is only desirable for the "glTF separate" mode. ".glb" and "embedded" shouldn't be concerned since (IIRC) they can be expected to be standalone.

The workflow is somewhat similar to the one we have at work (although proprietary formats instead of glTF are used); it's not a one-size-fits-all by any means, but it has many desirable qualities (especially the fact that no file is stored "redundantly").

What generally happens is as follows: there are many artists working on many models. All data such as models and textures are stored in a common folder, under version control such as Perforce or anything similar.  
Textures are responsible for a LOT of the total storage, especially these days when we are expected to perform 4K rendering and therefore store our textures in large resolution.  

"Models" (i.e blend files) are expected to be able to "reference" other assets such as models, materials, textures, whatever, but for the sake of simplicity we could pretend that the only "referable" asset type is textures, especially since they are the heaviest in terms of storage requirements. This is enough in practice, since models don't _generally_ reference others, OTOH it's very common for a texture to be shared by multiple models, especially for environment models.

So there is a `Models/` folder containing `.blend` files along with their exported `.gltf`, and a `Textures/` folder containing only the textures. This is a convention that the artists working on the project know to follow; they will always place (shared) textures in the `Textures/` folder, and make sure they are referenced as relative paths in the `.blend` file.  

It is important in order to allow, for instance, the following :
```
Models/Subdir1/level1.gltf   ----references (path relative to gltf)----> ../../Textures/Dirt/dirt_basecolor.png
Models/Subdir1/level1.gltf   ----references (path relative to gltf)----> ../../Textures/Lava/lava_basecolor.png
Models/Subdir2/Foo/level2.gltf   ----references (path relative to gltf)----> ../../../Textures/Dirt/dirt_basecolor.png
Models/Subdir2/Foo/level2.gltf   ----references (path relative to gltf)----> ../../../Textures/Lava/lava_basecolor.png

.... etc etc, except there is an insane amount of models, and an insane amount of large (4k+) textures, and very few space on the HDD or SSD.
In addition, Models/ and Textures/ have a large number of subdirectories.
This may appear exaggerated, but it's a real problem in practice.
```

For this to work, the glTF exporter has to be instructed to use relative paths as URIs. When this "option" is enabled, the files in `Textures/` are expected to be set up correctly such that the glTF exporter can use them directly without any change; if that is not possible, a warning or error should be emitted, allowing the artist to notice the issue, fix it, and move on.

That summarizes my use case.

About `gltf-transform`, it is great that such a tool now exists. However, AFAIK it doesn't fit my use case because:
- It adds a post-processing step which I would rather avoid;
- Merging glTFs in order to deduplicate textures seems prohibitive for a large number of glTF files. For other reasons, merging glTF files in the first place is not desirable in my workflow, although I am convinced it has its uses.

About #836, I guess that I could indeed use it to force textures to be exported into the same directory that contains them, but :
- As mentioned, there is still #743 which is slightly problematic;
- It could still lead to redundant copies of textures on disk (and therefore on Perforce, etc), for instance if my `.blend` file references a texture `../Textures/Foo/foo.png` but I set its _name_ in Blender to `bar`, then IIRC, the exporter would then create a file name `../Textures/bar.png`, which is not desired since I really want to use `../Textures/Foo/foo.png`.
  Let's also pretend that `Textures/` may have a large number of subdirectories, and a model may pick textures from many of them; exporting all textures at the root of `Textures/` is not desired.
  Artists should be able to name the textures in Blender however they want, since it's tracked separately from the file's path, without impacting the workflow I described (of course, only as long as the corresponding option is enabled).

>  I see a lot of models (especially DAE, OBJ/MTL) with absolute paths from the artist's hard drive, and "Original Paths" feel like a slippery slope toward that...

I agree that the occurrence of such models is unfortunate, however more often that not, this is an error that needs to be fixed. In a production setting, if an artist does export a model with absolute paths, one of the following happens:
- The exporter emits a warning or error (provided that it is configured accordingly).
- Whoever tests the application ends up noticing, the model is then fixed and a reminder is given to artists.
- The asset pipeline is actually able to deal with these paths, but can only do so with an ugly hack (e.g extracting the part after `... OurGameProject/Textures`).

The point is, I agree that absolute paths are "bad". Relative paths, however, may be very useful, especially when everyone in a team follows the same conventions (which is very often true, or rather, the opposite almost never works). 

There is nothing to fear about relative paths as long as : 
- Users choose to opt-in explicitly;
- Users are warned about the potential pitfalls. (e.g some glTF loaders will not accept relative paths as URIs.)

> it would be enough — in my opinion — to have a workflow that just allows the user to export glTF files from two different .blend sources and have some confidence that they'll point to the same textures.

This would be a nice improvement, but it's not "there" yet IMO. What I'm aiming for is exactly the following:
- There is a checkbox to opt-in to the feature when exporting. Users are informed of what they're getting into, and it is assumed they accept it.
- When the feature is enabled, texture files are not converted or "saved" at all. The file path as stored in Blender (NOT the name) is used directly as URI. This trivially prevents #743, and removes the need to use #836 as a workaround for this workflow.
- When the feature is enabled, warnings or errors are emitted for textures which would normally require conversion by the exporter (e.g a metallic-roughness texture with swapped channels, or when the source is not JPG/PNG). This could be "improved" later if someone needs.

That's "all" there is to it. I concede it's not that easy in practice, according to discussion in this ticket, but it's also quite simple in itself, to the point I feel there should be no need for workarounds.

I do agree that it might not be suitable as part of the "standard" exporter, since it's somewhat opinionated (although, again, it's just a checkbox).  
I do not know how to make user extensions, but that may be a better option. At least, better and cleaner than having to maintain my own fork. :)

yoanlcq:
I just noticed a case that wasn't covered by my last reply: what happens when the exported `.gltf` file is located in a folder other than the one containing the `.blend` file? That would require an extra step, which would do one of the following :
- Emit a warning or error. (Poor UX, but easy for the implementation, I guess);
- Fix all relative paths such that they are correct when taken relatively to to exported `.gltf`'s folder (this may be hard).

scurest:
I noticed DAE actually has an option (on by default) that's almost exactly this btw. We could steal it. I updated my branch.

copy

@yoanlcq 

> Fix all relative paths such that they are correct when taken relatively to to exported .gltf's folder (this may be hard).

Nah, it's [easy](https://github.com/scurest/glTF-Blender-IO/blob/8d62d09e2a5b6f5e25282b5180c8f84f1e9dba69/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py#L252-L256).

yoanlcq:
> I noticed DAE actually has an option (on by default) that's almost exactly this btw

That's nice to know; I tried it and could indeed notice that when `Copy` is turned off, textures are not copied, and their file path is directly included in the output DAE file. As you say, it's pretty much what I'd like.  
I could notice that the DAE exporter does not, however, perform any processing to "fix" relative paths.

> Nah, it's easy.

Woo! Looks like I've underestimated Python. That's great to know as well.

@scurest I've taken a quick look at [your branch](https://github.com/KhronosGroup/glTF-Blender-IO/compare/6786505...scurest:write-original-image-paths) and it looks very close to what I'd like.  
IIUC it still performs conversion if needed (instead of emitting a warning or error), but either that could be easily configured (?), or I could use that branch as a starting point and simply add custom behavior in my fork without bothering anyone.

In any case, I think that branch should be the starting point for any future work concerning this, instead of this PR. My `image-uri-handling` branch unfortunately became a catch-all for custom features, to the point is even includes hastily written code for supporting custom property animation (which actually "works" for me, but I digress).

I can't wait to test it as soon as I get the chance (not right now unfortunately), and I'm glad that the initial issue got some attention again.

scurest:
>  I could notice that the DAE exporter does not, however, perform any processing to "fix" relative paths.

Yeah, it looks like it always does absolute paths when it doesn't copy. We shouldn't steal that part lol.

>  IIUC it still performs conversion if needed (instead of emitting a warning or error), but either that could be easily configured (?), or I could use that branch as a starting point and simply add custom behavior in my fork without bothering anyone.

Yeah, correct. That's how DAE works too: if a texture has no file (eg. for GENERATED) it will write its own file for it even when Copy=off. I think that's the best thing to do. You could patch it around [here](https://github.com/scurest/glTF-Blender-IO/blob/8d62d09e2a5b6f5e25282b5180c8f84f1e9dba69/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py#L146) if you want to change it though (if the uri was None, print a warning, and return None if you don't want to convert it).

Is your goal here to ensure that no glTF has its "own" textures? ie. all textures come from the shared `Textures/` (or whatever) dir?
 
> even includes hastily written code for supporting custom property animation

(I was wondering about that!)

yoanlcq:
> Yeah, it looks like it always does absolute paths when it doesn't copy.

Actually, when I tested it, it always used the relative path exactly as provided in Blender. However, that is regardless of the directory where I save the output file, which is the problematic part.  
Like, if I specify `/some/absolute/path.jpg` as the image path in Blender, the exported DAE uses `/some/absolute/path.jpg`, but if I specify `//../some/relative/path.jpg` in Blender, the exported DAE always uses `../some/relative/path.jpg`, not an absolute path.  
If that can make a difference, I was testing this on Linux with Blender 2.79b (I'm not on the PC I usually work on, which has Blender 2.83 and all my projects).

> Is your goal here to ensure that no glTF has its "own" textures? ie. all textures come from the shared Textures/ (or whatever) dir?

Yes, pretty much, AND also that the textures can directly be used as-is, without transformation (if that is not the case, emit a warning/error), rather than "silently" adding (or worse, overwriting?) files, which are heavy. We want to have as few of them as possible.
I see it just like we, as programmers, use assertions to catch our own mistakes before disaster happens.

What I think would be great is to have a checkbox for that, like "Allow conversion" or something. By default it would be true, which is the most friendly (and current) behavior, but it could be set to false, in which case the exporter would say "hey, your file is not PNG/JPG (or metallic-roughness channels are swapped, etc) but you told me you didn't want me to convert your images, so please fix it yourself then try again".

About custom property animation in my branch, well, I recall there is not much to it. What I know is that, when I wrote it, glTF as a format did not have any support for that, not were there any plans, so I went "well, I'll just solve my own problem and move on".  
That did require me to also fork the glTF parser (the `glTF` Rust crate) I'm using, just for it to accept the arbitrary property names.

In a glTF, in the `"channels"` under an element of `"animations"`, that looks like this : 
```json
                {
                    "sampler" : 38,
                    "target" : {
                        "node" : 23,
                        "path" : "CUSTOM_PROPERTY:hf.animatable_props.play_stepping_sound"
                    }
                }
```

The target node looks like this: 

```json
        {
            "extras" : {
                "hf" : {
                    "animatable_props" : {
                        "play_stepping_sound" : 0
                    }
                }
            },
            "mesh" : 1,
            "name" : "Feet",
            "skin" : 0
        },
```

The custom property can be animated in Blender just like any property. Without any other changes, the application can perform special processing when a custom property keyframe is encountered during the playback of an animation, treating it as an "event". In my project, it is used to play stepping sounds during a walk/run animation.  

In any case this is very opinionated and I'm not sure how far it can be pushed, hence I kept it for myself. I think it only works on scalar numbers. Maybe it works on vectors, but I'm not sure.  
I really wish there was a serious proposal for custom property animation in glTF, especially since most of the infrastructure is in place; it's mostly about accepting arbitrary (though prefixed?) names in `target.path`. But I'm pretty sure a lot of issues could be found if this were to be standardized.

Issue #425, opened 2019-04-11
Transparency issues


SteffenDuenner:
**Describe the bug**
Exported PNGs have their alpha channel deleted on export?!

**To Reproduce**
- Add a grid with UVs
- Assign a principled shader
- Plug a PNG with an alpha channel in base color
- Set blend mode to "Alpha Clip" or whatever
- Export a gltf + bin + textures (just to easily see that the exported PNG has no alpha channel any more. Also works with glb and gltf embedded)
- Import into Blender again (or a viewer of your choice) -> no transparency, the PNG is RGB only

**Expected behavior**
Alpha channel of the PNG controls transparency / opacity

**Version**
 - OS: Linux Mint 19.1 x64
 - Blender Version: Latest own build containing the draco compression and JPG+PNG conversion

SteffenDuenner:
If I reimport the glb / gltf in Blender it sets the blend mode to "Alpha Clip", the only problem is that the alpha channel from the PNG is now gone.

julienduroure:
Can you please provide a glb file generated and the original .blend file ?
Did you test the glb in some external visualisation tool, like https://gltf-viewer.donmccurdy.com ?

SteffenDuenner:
Yes, I always test the files in other viewers (three.j/editor, donmccurdy, babylon sandbox) and none of them shows transparency.

[Transparency_v01.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3072944/Transparency_v01.zip)

Here's the Blender 2.80 file with packed texture and the resulting glb. Exported from the very latest own build of 2.80.

Thanks for having a look :)

Zool1987:
I have the same problem ... And another question on the topic is it possible to use a gradient in texture? Now I only have a blend mode without texture in my shader Principle BSDF...If you can do this, you can share the node scheme?

vitorbalbio:
Hi!
After recent updates the png/alpha export stopped work. I get exactly the same problem as reported above. It's probably caused by this commit: https://github.com/KhronosGroup/glTF-Blender-IO/pull/404

Here's a sample file:
[BugAlpha.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3095041/BugAlpha.zip)

julienduroure:
This very recent change in Blender will maybe help us : 
https://developer.blender.org/rB8f71a84496a95528303fbe0bb7c1406060353425

Cf #462

PR #454, opened 2019-05-08
Fix png image transparency issues #425

khasvictor:
Fix the image save logic so that png textures will keep their alpha channels.


emackey:
Nice!  But question, can we make this only apply to the `BaseColor` image?  It, plus someday the spec/gloss image, when that gets added to the exporter, are the only two images that can possibly use RGBA, the rest should just be RGB.

emackey:
Normal map, occlusion, rough/metal, and emissive should not include an alpha channel.

emackey:
Thanks @khasvictor, this helped inspire and has now been superceded by #506.

PR #455, opened 2019-05-08
Better bone tail calculation

MarimeGui:
Hello everyone,

This is my first ever Pull Request, I'm sorry if I am doing anything wrong !

GlTF is my favorite 3D model format because of its versatility and simplicity.
I often use Blender (with this addon) to open, view and sometimes fix the models that I get from other programs, mostly from "Game Engine models to GlTF" converters.

Because I work with Game models, I often have to deal with complicated rigging and a lot of bones. When importing such models into Blender with this Addon, the default rigging is not very usable and you mostly have to rely on the Outliner to select the correct bone you want to pose.

Here are a few examples:
_Example 1_
before1
_Example 2_
before2

To fix this issue, I re-wrote the "Armature Correction" part of the import to a simpler way of choosing the right value for the tail of every bone.

To achieve this, I first check if the bone has children.
If this is the case, then I calculate the average position of all the children heads and use this average as the new value for the tail of the bone.
If the bone does not have any children, I instead calculate the difference of position of the parent bone between its head and its tail, and add this difference to the head of the bone and use this value as the bone tail.

This method yields pretty clean results in my opinion, here are the same examples as above with my code:
_Example 1_
after1
_Example 2_
after2

This already helped me pose a few models, and I think this should be of use to everyone else.

What do you all think of this ?

Thanks for your time,
MarimeGui.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py
@@ -82,6 +82,7 @@ class BlenderGlTF():
                 bpy.context.view_layer.objects.active = armobj
             armature = armobj.data
             bpy.ops.object.mode_set(mode="EDIT")
+            no_children_bones = []
             for bone in armature.edit_bones:
                 child_bone_sum = [0, 0, 0]
                 child_count = 0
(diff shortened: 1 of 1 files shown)


MarimeGui:
Hello Again, I just committed a little fix to my code :

I added a check to make sure that in the case that there are no children bones, make sure that a parent exists before calculating the difference between the head and the tail of said parent.

I also made the calculation for the bones that have no children after the rest because the parent bone tail might change after it is used to calculate the difference, that would mean that the new tail wold not be correct.

julienduroure:
Hello,
Thanks for the PR.
I will have a look on it next week. Screenshots look promising :)

Ali-RS:
I also have the tail bone issue, I hope this PR can be integrated soon.

julienduroure:
Here are the files used for my review:
[non_symetrical.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3193683/non_symetrical.zip)

MarimeGui:
Alright, ill take a look on how I can come up with a solution that has both benefits this Week-End.

julienduroure:
I will close this PR. Don't hesitate to reopen if you found a new solution

PR #456, opened 2019-05-09
Add Captain Joe test character scene

pjoe:
I've created a full character for better testing skinning/animation.

It's rigged with the rigify add on, which does produce a somewhat complex rig (with lot's of cool features for animators).

Currently I've just added a couple of test poses to the pose library.

The file works in both 2.79 and 2.80, but you have to allow it to run the included script (`rig_ui.py`). This is generated by rigify and adds the nice extra rigify UI and functionality.

image


donmccurdy:
There are no animations included with this right, or did I miss them? Should poses export as animations? Either way the exporter isn't exactly passing this test yet haha:

Screen Shot 2019-05-13 at 10 55 37 AM

I agree that supporting the output of the Rigify addon is a good goal and test case.

pjoe:
No animations (yet), just a couple of poses in the poselib. And yeah currently exporting this is not exactly pretty 😟 

I'd love to also have tests for other popular character tools (Fuse/Mixamo, Character Creator, MakeHuman), but not too sure about licensing for models created with those. Suppose we could try contacting them, but maybe first get this one in a working state 😃

julienduroure:
This kind of rig (rigify) will not work without fixing #460.

Makehuman is also an opensource project, without any issue about sharing results.

I will also provide my test case, that is more simple than rigify, but with same issues (mostly #460 and #461)

julienduroure:
Rigify changed a lot during 2.80, 2.81 (and is changing during 2.82) developpement.
I propose to cancel this quite old PR, and a rigify 2.82 can be added back after 2.82 release.

Issue #462, opened 2019-05-13
Principled node now has alpha and emission inputs


julienduroure:
Cf https://developer.blender.org/rB8f71a84496a95528303fbe0bb7c1406060353425

We have to change importer and exporter to take into account this change (this will simply the node setup :) )

SteffenDuenner:
Nice! I was just about to post this commit here :)

donmccurdy:
Wow, yes that potentially helps quite a lot. Do "Alpha Blend" and "Alpha Clip" modes still work as expected?

Was this on a roadmap we could have seen earlier? I'd asked in the dev forums a few months ago and it sounded like the answer was "maybe someday" at the time, might have saved some trouble if we'd known this. 😅

In any case, strong +1 for using the new inputs.

julienduroure:
Was a surprise for me too ;-)

emackey:
Very cool.

Via ZanQdo on the [KhronosDev](https://www.khronos.org/news/permalink/khronos-developer-slack-5bfc62eb261764.20435008) chat:  https://devtalk.blender.org/t/13-may-2019/7076

andrewpeterprifer:
Just ran into this too, any development on this?

emackey:
There's an unfinished pull request in #476 that needs some review feedback addressed before it can proceed.

PR #466, opened 2019-05-14
✨ Allow Opacity export

m0ppers:
Creating a pull request to start some discussion:

I want to export some objects having an opacity map. So far I reused the the "standard" glTF Metallic Roughness node and added the opacity map there.

When exporting I am merging the info from there into the basecolor alpha channel.

This is working like a charm for me and the babylons.js sandbox (https://sandbox.babylonjs.com/) is displaying the result as expected.

Would this be a wothwhile addition? What is missing to get a feature like this merged? I am especially unsure about the hardcoded alpha=True in the image encode function.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
@@ -43,7 +43,6 @@ def gather_image(
 
     mime_type = __gather_mime_type(blender_shader_sockets_or_texture_slots, export_settings)
     name = __gather_name(blender_shader_sockets_or_texture_slots, export_settings)
-
     uri = __gather_uri(image_data, mime_type, name, export_settings)
     buffer_view = __gather_buffer_view(image_data, mime_type, name, export_settings)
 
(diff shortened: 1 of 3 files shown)


jjcasmar:
This should be documented as an extension, no?

julienduroure:
Hello,

Thansk for the PR.
Some remarks: 
- Some tests failed.
- See #462 : Now that principled node had emission and alpha input, we need to use it for 2.80, and we have to keep old system for 2.79 compatibility.

m0ppers:
Oh wow! that changes a lot! Having a standard way to do that will simplify things. is somebody already working on this?

`and we have to keep old system for 2.79 compatibility.`

not sure what that means. Probably one should simply skip opacity export for 2.79?

donmccurdy:
Hi @m0ppers  -- thanks for this PR. With the new properties in 2.8 on the Principled BSDF node, I do think we should focus on getting that right. Unfortunately that means probably no opacity export in 2.79, except where it's handled already (e.g. alpha channel in the base color texture).

I don't think anyone has started implementing support for the new Principled properties, PRs would be very welcome.

Screen Shot 2019-05-23 at 5 40 54 PM


Closing this PR in the meantime.

Best,
Don

Issue #426, opened 2019-04-13
Export Animation issue when only 1 item of transform is keyed


julienduroure:
When only 1 item of transform is keyed (for example only X location), other items (Y and Z) are not exported in right place.

See blend & glb file generated in attachement.
In https://gltf-viewer.donmccurdy.com/, we can see that there is 1 cube on each side of the plane, where both cubes were in same side in Blender

[bad_animation.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3075815/bad_animation.zip)

julienduroure:
Now committed :)

PR #469, opened 2019-05-15
Fix 426

julienduroure:


julienduroure:
Some very simple file I used for unit tests: 
[bugs_426.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3184086/bugs_426.zip)

@emackey @donmccurdy Ready to be tested if you want

julienduroure:
I found some non working cases. Please do not merge

julienduroure:
I removed my last comment, every should be OK

julienduroure:
@Cyp , maybe you can have a look on this one too. Thanks !

Cyp:
Using `blender_object` instead of `blender_object_if_armature` has the side effect that `@cached` no longer works well on `gather_keyframes`, `__gather_input` and `__gather_output`, resulting in a larger .gltf and .bin size in the case that multiple objects share the same animation, although the output still seems correct. Passing both `blender_object_if_armature` and a pregenerated `non_keyed_values` to the cached functions might work, I'd guess.

Apart from that, I didn't notice any obvious issues.

julienduroure:
@Cyp : This fix is not only about rigging, but can be applied on blender object too. That's why we can't use blender_object_if_armature, but blender_object directly. 

Will try to find a way to get the cache working

Cyp:
What I meant is that instead of using `blender_object` directly, it could extract the `non_keyed_values` from the `blender_object` before passing to the `@cached` functions. Assuming that's all the `blender_object` is used for, that is.

julienduroure:
@Cyp I push some fixes about cache when multiple objects share an animation

donmccurdy:
Testing against the original file linked in https://github.com/KhronosGroup/glTF-Blender-IO/issues/426, I am still seeing the cube on the far right move on multiple axes -- do you get the same result? Or does this fix require non-default settings?

julienduroure:
@donmccurdy 
With file in #426 , now both cube are on the same side of the plane, as in original blend file.

I also have the same result (the cube move on multiple axes). This is another bug linked to CUBISPLINE export.

In attachment, same blend file, but with LINEAR interpolation
[animation_issue.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3215099/animation_issue.blend.zip)

julienduroure:
OK, thanks. 
I will manage merging myself later today

Issue #461, opened 2019-05-13
Sample animation issue when using NLA


julienduroure:
Issue about animation that occurs when we have to sample animation (forced by user in option, or when needs_sample return true)

Because in gltf2_blender_gather_animation_sampler_keyframes.py file, we check value of  pose_bone_if_armature.matrix_basis.decompose()

BUT:
- this value is always the same if we stash animations, and there is no active animation. See animation_case_1 file
- this value is false, and is taken from active animation. That leads to export always the same animation, even if there are different animations stashed. See animation_case_2_file

[bugs_animations.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3173283/bugs_animations.zip)

julienduroure:
Some thoughts, to not forget:
[ ] Make each action active when exporting each action of the NLA?
[ ] When list of bones in animation are not identical ...

Issue #460, opened 2019-05-13
Animation export issues when DEF bones are not animated


julienduroure:
When Deformation bone is not animated, but is linked (directly or not) to an animated bone via a constraint. What I mean by "directly or not":
- directly: DEF bone has a constraint to an animated bone. See bug_defbone_constraint.blend file
- not directly: DEF bone is child of an animated bone, or child of a bone that have constraint. See ball file

Some thoughts of possible solution :
- try to detect if a bone is indirectly animated
- option to auto sample every DEF bone if there is at least 1 bone in armature that is animated

[bug_defbone_constraint.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3173105/bug_defbone_constraint.blend.zip)

[ball.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3173115/ball.blend.zip)

julienduroure:
It's now committed :)

Issue #451, opened 2019-04-30
Exporting multiple actions corrupts them


SirSilverCharles:
I have a rig and I stash various actions(animations), I export the glTF and I import it in a blank file, I select the rig and I select one action in the action editor, the action doesn't look right instead it looks blended with something else.

How it should look:
![image](https://user-images.githubusercontent.com/17120210/56934718-bb438c80-6ab2-11e9-9c6b-411a83a84051.png)

How it looks, is sitting as if it ignored some bones position/rotation keyframe, (same with other actions):
![image](https://user-images.githubusercontent.com/17120210/56934730-d0202000-6ab2-11e9-8841-3756978ccff0.png)

I tested many variations: "Always Sample Animations" option as well as other options, adding the actions in new Nla tracks instead of stashing them, I imported in three.js(with same result, so probably the exporter is the problem), I keyframed all the bones at the start of the action, I made sure that the active action is empty before exporting, turning on or off the "Do channels contribute to result" in the Nla track, adding all the actions in a single track instead of separated ones, etc.
If I export only one action it looks correct.

[characterWoodsmall A2.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3129600/characterWoodsmall.A2.zip)
Here is the file, just open it export and the in a new blender import it, select the rig and select an action from the "dope sheet"->"action editor"->"browse action to be linked".

This was in Windows 7, and blender-2.80.0-git.a57fec986d2d-windows64

Thanks.

julienduroure:
Multiple checks/issues here :

* Open glb on https://gltf-viewer.donmccurdy.com/ Everything seems to be OK, except Neutral pose --> Exporter issue

* Importer issue: lots of mesh without any armature modifier: Link to #331 ?

SirSilverCharles:
Thanks. Is not only the neutral pose, all the other poses have problems but are less obvious, is clear in the position of the hand mostly.

SirSilverCharles:
I found a workaround: duplicate(alt+D) the rig for each action and only use one action in each rig, the actions export correctly that way.

ERS-EIT:
I have kind of the same problem. If my model has more than one Animation and I do not sample every keyframe, the byte array in the glTF file seems to have intermediate zeroes.
So if I take a look at my keyframes (for a given joint) there are thrice as much scaling vectors in the buffer as I have timestamps. The "thrice" part might be variable. Maybe it is because I have three animations in that particular model.

[...]

PR #479, opened 2019-05-19
Fix 460: When sample animation mode is used, bake all bones

julienduroure:
When sample animation mode is used, all bones are baked, not only bones with keyframes.
This will allow to export complex rigs, where deformation bones are not animated. This is how fbx export works, for example.

Tested on some various rigs, including a bouncing ball,and a simple character with IK leg / foot roll.
[460.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3194998/460.zip)


julienduroure:
Seems ready to be tested @emackey @donmccurdy

emackey:
Unfortunately I'm not the right person to be reviewing skinning/bone/rigging code.  @donmccurdy or @pjoe or @Cyp, would you guys have any time to look at this?  Feedback from anyone else is welcome too of course.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_channels.py
```diff
@@ -35,8 +35,21 @@ def gather_animation_channels(blender_action: bpy.types.Action,
...
+        for chans in groups:
+            ranges = [channel.range() for channel in chans]
+            if bake_range_start is None:
+                bake_range_start = min([channel.range()[0] for channel in chans])
+            else:
+                bake_range_start = min(bake_range_start, min([channel.range()[0] for channel in chans]))
+            if bake_range_end is None:
+                bake_range_end = max([channel.range()[1] for channel in chans])
+            else:
+                bake_range_rend = max(bake_range_end, max([channel.range()[1] for channel in chans]))
```
Cyp: `bake_range_rend` typo?

donmccurdy:
I'll review it this week.

(and thank you @julienduroure for the PR and @Cyp for the review, much appreciated!)

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_channels.py
julienduroure: Yes, typo. Fixed now

pjoe:
From a quick read it looks pretty good to me :) I will try to do some tests with different files.

Btw. interesting with those python type annotations, though I do feel a bit weird that there is no actual type check - guess I have been doing to much TypeScript lately 😛

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_channel_target.py
```diff
@@ -24,37 +24,47 @@
 @cached
 def gather_animation_channel_target(channels: typing.Tuple[bpy.types.FCurve],
                                     blender_object: bpy.types.Object,
+                                    bake_bone: typing.Union[str, None],
+                                    bake_channel: typing.Union[str, None],
                                     export_settings
                                     ) -> gltf2_io.AnimationChannelTarget:
```
donmccurdy: Hm, it seems worrying to see a `@cached` function that isn't taking any parameters related to the current NLA track or Action name. This seems like a probable cause of https://github.com/KhronosGroup/glTF-Blender-IO/issues/451 and https://github.com/KhronosGroup/glTF-Blender-IO/issues/461?

donmccurdy:
@julienduroure what export settings should I be using to check the IK leg example? With *Always Sample, or Always Sample + Bake Skinning Constraints, I'm not getting a correct output (although it looks closer than before.

donmccurdy:
Also – None of my skinning examples export correctly on the master branch. I'm happy to do reviews, but I'm also OK if we want to merge PRs more quickly, given that the master branch skinning export is already quite broken as far as I can tell. If there are any skinning examples that do work now, let's upload them somewhere – otherwise my testing things is not very effective. 😔

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_channel_target.py
julienduroure: Yes, you're probably right.
#461 will be my next work, after #426 and #460 will be merged

julienduroure:
@donmccurdy Currently using only *Always sample* because *bake sinning* does nothing in code for now (see #485 for the discussion about that)

You have here 2 examples
[460.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3194998/460.zip)

And here is a really complex rig (generate with rigify). Re-import in blender is correct (some glitches maybe linked to only 4 skin weights export ?), but is KO when viewing it in https://gltf-viewer.donmccurdy.com/ :( 
[rigify.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3215136/rigify.zip)

julienduroure:
@donmccurdy :

In https://gltf.insimo.com/, animation is better, even if there is still a global orientation issue
And everything seems to be OK in https://sandbox.babylonjs.com/

donmccurdy:
Ok thanks! I've filed https://github.com/mrdoob/three.js/issues/16562 about the animation issue in three.js, we can look into that. We ready to merge this?

julienduroure:
@donmccurdy Before merging, I want to be sure that this behaviour is linked to the right user option in UI. 
My comments and solution proposal are in #485

julienduroure:
Thanks for your feedback. I will manage merging myself later today

PR #490, opened 2019-05-22
Add initial regression tests using golden files

pjoe:
Adding checks of content of exported files if a 'golden' expected file is present in `tests/scenes/expected`.

Added 01_cube.gltf, but so far this is failing for 2.79 as it doesn't set `doubleSided: true` on the material


pjoe:
Added initial handling of difference in accessor ordering. However I now see that accessor content seem to differ from run to run (in 2.79). Looks like vertex ordering is also changing :S

emackey:
My general question with 'golden' files is whether they will be too brittle.  For example, if we make another change to the way doubleSided is handled (recently the defaults changed from false to true, due to reasons not completely under our control, for example), will that break every golden file for other topics like UVs and morphing?

I suspect it might be more robust to have each test check specifically the values it cares about for the test, and not just compare the whole glTF file.  Otherwise, if we ever make another change that affects some tiny portion of the overall glTF structure of every export, I would expect every golden file test to break or be regenerated, and that will hide in the noise any actual collateral damage that we should be looking for.

emackey:
> Added 01_cube.gltf, but so far this is failing for 2.79 as it doesn't set `doubleSided: true` on the material

This is another issue with golden files.  The same `.blend` project currently produces different outputs when loaded into 2.79 vs 2.80, because in order to be used in both, the shared `.blend` project must be a 2.79-era project, and thus lacks control over APIs that were added or dropped in 2.80, including per-material vs per-mesh double-sidedness.  This means we can expect the projects to produce different glTF files as a result.

emackey:
Let's park this for now.

PR #492, opened 2019-05-22
Test suite should look at the Validator's issues.numErrors

### LINKED ISSUE #486: Validator results not checked carefully enough
Well, I'm sorry I didn't notice this before now.  The test suite in this project runs the glTF Validator, and it looks at some keys like `hasTextures` and `primitivesCount` to match expected values.  It also looks for the Validator's promise to be rejected, meaning that the Validator itself crashed or wasn't able to process the file being validated.

One critical piece is missing:  The test suite is not looking at `issues.numErrors` to see if the model actually passed validation with no errors.

The good-ish news is that most of the test models do pass validation in master, even without this check in place.  However, 2 of them do not:

```
scenes/01_morphed_cube.gltf
```

> Animation sampler output accessor of count 18 expected. Found 12.

```
roundtrip/03_all_animations/03_all_animations.gltf
```

> (482 errors, all normal vectors are zero length)

I discovered this while attempting to add a new test model to roundtrip.  The model is round-tripping and losing all of its normal vector information in the process, and I have not yet figured out why.  I'll file a separate issue on the normal vectors once I have more information, but it would appear that at least one of our existing roundtrip models exhibits the problem.

emackey:
Adds a check for `issues.numErrors`, and produces messages in the test suite output for the first few error messages produced by the glTF Validator.

Fixes #486.

Don't merge until the tests pass.  I still need to move one broken model out of the way, before they will pass.


emackey:
CircleCI appears to not be running on the most recent commits here...  But why??

emackey:
OK, I parked the one test that is still failing here, as it will require a separate PR to fix it.

Merging this.  I'll file a new issue for the parked broken test.

PR #502, opened 2019-05-28
Add doubleSided import (2.80 only), add stronger test for roundtrip.

### LINKED ISSUE #501: Import support for glTF's doubleSided flag
Can we add import support for doubleSided for Blender 2.80?

I think this is a one-line change, around line 51 in [imp/gltf2_blender_material.py](https://github.com/KhronosGroup/glTF-Blender-IO/blob/ecb1fc4cd5ab543b5e3e6b92d22e58f422f3862d/addons/io_scene_gltf2/blender/imp/gltf2_blender_material.py#L51)

```
        mat.use_backface_culling = (pymaterial.double_sided != True)
```

This only works in 2.80, of course.  I don't much care about adding support to 2.79, since 2.79 did not have a per-material backface culling flag.

I tried this, but apparently it's not valid Python:

```
            # Blender 2.79 did not have a per-material double-sided flag.
        else:
            mat.use_backface_culling = (pymaterial.double_sided != True)
```

Maybe we need a new construct for code that only runs on 2.80?

emackey:
This replaces an older roundtrip `01_two_sided_plane` test model with a new `01_single_vs_double_sided` test model.  The new model has one single-sided (green) and one double-sided (light blue) material.

A small change to the importer allows the new 2.80 backface culling flag to respect the glTF doubleSided flag.

A new test has been added to look at the output of the round-tripped model, to make sure the single-sided and double-sided materials came through the round trip with their flags intact.  This is only tested in 2.80 (in 2.79, we merely test that valid models came out, not that the flags were preserved).

Fixes #501.


emackey:
The new test runs in [container 3](https://circleci.com/gh/KhronosGroup/glTF-Blender-IO/625#tests/containers/3), round-trip of 2.80 models, near the bottom, "roundtrips the doubleSided flag".  It passes.

Issue #498, opened 2019-05-25
Metallic Roughness texture changes after export


Ali-RS:
This is how my metallic roughness texture looks like: 
![Warrior_marmoset_Metallic_Roughness](https://user-images.githubusercontent.com/5787647/58374596-19e71380-7f56-11e9-9e1d-a2d06fdcfb1f.png)

after exporting the model to gltf I get this :  
![Warrior_marmoset_Metallic-Warrior_marmoset_Roughness](https://user-images.githubusercontent.com/5787647/58374604-397e3c00-7f56-11e9-831b-6da9e207fd3d.png)

seems it adds color to R channel after export.

I use this example to create my material : 
![addons_io-gltf2-material-principled](https://user-images.githubusercontent.com/5787647/58374635-c0331900-7f56-11e9-8bc6-2364706ef6d1.png)

I am on blender 2.8 with gltf add-on version 0.9.11

julienduroure:
@Ali-RS Can you please share the .blend file?

Ali-RS:
Yes
[test.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3220068/test.blend.zip)

Ali-RS:
@julienduroure I found that it does work fine in v 0.9.1 but broken from v0.9.2.
So it should be one of these commits that breaks it: 
2ea96343df3e4d13b7a16ef4611fd16f6acdc2ff
or
f12b8205284d13ebcd294abc525ff54d8d4c1663

julienduroure:
@emackey : Seems to be related to your commit. Do you have time to have a look on it?

emackey:
If you're not using an occlusion channel, there is no need for this exporter to preserve the red channel.  The exporter will assemble occlusion, roughness, metallic from separate grayscale images if need be, and the final result is expected to be different from the source images.

That said, there is a separate desire to have the exporter preserve original image encoding whenever possible, and that would affect this outcome as well.  There's a long history of that feature getting implemented and then getting broken again, first in the glTF-Blender-Exporter project, and again later in this project, most recently in #404.  We're tracking that problem in #428.

julienduroure:
@emackey: Now that #506 is merged, maybe this ticket can be closed?

Ali-RS:
@julienduroure seems not resolved yet, please see the `Stuff still needs doing:` in this post: 

https://github.com/KhronosGroup/glTF-Blender-IO/pull/506#issue-283435093

[...]

PR #506, opened 2019-05-29
Image handling fixes

### LINKED ISSUE #425: Transparency issues

### LINKED ISSUE #428: Texture files get re-encoded
**Description**
I noticed that upon exporting glTF (at least separate ones) the textures get re-encoded.
This way users have no controll over compression settings and quality.

In case of my demo file, the JPG also increases in size a bit. Since jpegs are lossy, I consider this additional re-encode as quality flaw. Even lossless PNGs might get bigger in size in this step.

I can imagine good reasons for re-encoding (like limiting texture size, make the size a power of two, optimizing PNGs or changing to a better compression), but I think that's not the case here.

**To Reproduce**
Take a look at the provided blend + jpg files and the respective export in the `gltf` subfolder.
The source txt.jpg is not the same as the exported one.

**Expected behavior**
I'd expect the exported files to be precicely the source texture files (in case they are JPEG or PNG files).
If textures get re-encoded, I'd like to have control over the encoding settings (format, quality settings etc.)

**.blend file/ .gltf**
[reencode.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3076461/reencode.zip)

**Version**
 - OS: macOS
 - Blender Version 2.80 beta

### LINKED ISSUE #478: T64760 : Alpha export regression
Bug from blender repo:
https://developer.blender.org/T64760

emackey:
Some changes so far here are:

- White images and "composed" images are only done for ORM channels now.  Stuff like base color, normal, and emission will short-circuit all of the image channel composition logic, which may result in a small performance and memory improvement.
- Images that manage to take the shortcut above, including Base Color, will have their alpha channel preserved, if it exists, even if the image needs encoding.
- Images with sources on disk, that take the shortcut above and use the same MIME type, will read the encoded data directly from the source file, rather than re-encoding.  This means that user-supplied image encodings can be packed into GLBs now.
- Round-trip of `alphaMode` enum and `alphaCutoff` value now tested in 2.80.
- Bug with import of alphaCutoff discovered in above test, and fixed.
- One of the ORM tests expected an image to be re-encoded, now expects it _not_ to be re-encoded.

Fixes #425, fixes #428, fixes #478

Stuff still needs doing:

- [ ] There's still a dangerous situation where exporting a glTF to the same folder as the .blend file's source images may cause files to be overwritten.  The code to protect against this was [refactored out in #404](https://github.com/KhronosGroup/glTF-Blender-IO/pull/404/files#diff-444226537cdaaa92459dad1cd814237bL162) and is nontrivial to recover, but needs to be recovered.  Work in progress.
- [ ] There could be further optimization to the ORM hookup, specifically if the user hooked up the channels the glTF way, we could take the same short-circuit path and get all of the preserved image encoding benefits.  This would "fix" #498 (which isn't a bug, just a user misunderstanding that the red channel is unused and therefore allowed to change when the ORM is composed from channels).


donmccurdy:
For the two TODOs you mention, it sounds to me like those could be left until after this PR? Any preference?

emackey:
I'm fine either way.

donmccurdy:
This PR is a strict improvement as far as I can see, so I'd vote to merge it as-is.

PR #507, opened 2019-05-30
Fix #461 export multiple action from NLA

### LINKED ISSUE #461: Sample animation issue when using NLA

julienduroure:


donmccurdy:
Testing with https://sandbox.babylonjs.com/ and this model:

[RobotExpressive.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3238509/RobotExpressive.blend.zip)

Many of the NLA tracks are coming through correctly, but some are not, or seem to be showing animation data from other tracks. For example `Robot_Idle` looks the same as `Robot_Dance`. Perhaps that is related to the caching code?

This is also a case where merging NLA tracks with the same names into a single animation would be helpful, although that doesn't need to be in this PR.

julienduroure:
@donmccurdy You was right, it was a cache issue, when 2 actions have same length. This should now be fixed.

Merging 2 tracks when same name will be done in another PR. Baby steps :)

donmccurdy:
Tested another complex example working well too. 👍

PR #515, opened 2019-05-31
Ignore animation import if target has no node

djiamnot:
I encountered the following error when importing a .gltf file (previously exported with this addon)
```
 File "≤path_to≥/2.80/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py", line 262, in pre_compute
    if anim_idx not in gltf.data.nodes[channel.target.node].animations.keys():
TypeError: list indices must be integers or slices, not NoneType
```
It turns out that the exporter exports an `animation` block if animation export is checked (and it is by default) but there is no animation data. The resulting file cannot be imported back. This fix will not ignore offending operation if `target` has no `node`.

[book_open.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3242686/book_open.zip)

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py
@@ -279,6 +279,9 @@ class BlenderGlTF():
 
             for anim_idx, anim in enumerate(gltf.data.animations):
                 for channel_idx, channel in enumerate(anim.channels):
+                    if channel.target.node is None:
+                        continue
+
                     if anim_idx not in gltf.data.nodes[channel.target.node].animations.keys():
                         gltf.data.nodes[channel.target.node].animations[anim_idx] = []
                     gltf.data.nodes[channel.target.node].animations[anim_idx].append(channel_idx)


julienduroure:
I just checked the specification, and 'node' is not required : 
https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/animation.channel.target.schema.json

I am not sure what should happen when there is no node.

Maybe we must first check what happen in exporter step. Can you please provide the .blend ?

Thanks

donmccurdy:
Agreed with ignoring the channel if the node is not specified. It isn't required for forward-compatibility, as there may be other types of animation in the future, but there's no reason this exporter should be writing animations not targeting a node, so that part sounds like an export bug.

djiamnot:
I confirm that this must be an export bug. As I was trying to reproduce the bug, I realized that the object in question did export correctly when it was alone. However, it exports the animation block when other objects in the scene have animations.

In the attached file, exporting the (visible) book object reproduces the issue.

[open-book-gltf-issue.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3242823/open-book-gltf-issue.zip)

julienduroure:
I opened #529 to not forget the exporter bug.

PR #521, opened 2019-06-01
Fix 494 morph anim export

julienduroure:


emackey:
There's some slight strangeness in the exported (and roundtrip output) version of `01_morphed_cube.gltf` where the top vertex of the cube seems to "overshoot" and turn a bit inside-out in both the BabylonJS and ThreeJS viewers.  This isn't visible inside of Blender itself.  Not sure if this was a pre-existing or unrelated problem.  This PR is a big improvement in that the exported model is valid now, and works other than that tiny glitch.

emackey:
For what it's worth, the animation `output` data here for `01_morphed_cube` looks slightly different.  In the original file (in the test suite), the data looks like this:

```
                        [
                            0,
                            0,
                            0,
                            0,
                            0,
                            0,

                            0,
                            0,
                            1,
                            1,
                            0,
                            0,

                            0,
                            0,
                            0,
                            0,
                            0,
                            0
                        ]
```

But after roundtrip here, it looks like this:

```
                        [
                            0,
                            0,
                            0,
                            0,
                            0,
                            0,

                            0,
                            0,
                            1,
                            1,
                            1,
                            1,

                            0,
                            0,
                            0,
                            0,
                            0,
                            0
                        ]
```

These numbers represent the decoded contents of the accessor named by `animations[0].samplers[0].output` in each case.

julienduroure:
@emackey It was a typo in the code. Now this is better :)

emackey:
I missed that too.  Export looks perfect now.

Looks like there is an import bug, likely unrelated to this exporter change.  The [01_morphed_cube](https://github.com/KhronosGroup/glTF-Blender-IO/tree/d269b4ab010a972db5fce867169848ff91f10129/tests/roundtrip/01_morphed_cube) model imports with the morph key values all at zero, so the animation doesn't cause any visible changes.

emackey:
The importer doesn't handle `CUBICSPLINE` interp when importing morph weights, and doesn't multiply the index by 3.  This is a separate issue from this PR.

https://github.com/KhronosGroup/glTF-Blender-IO/blob/9fe7d9b8db4418bc6708473a4ae6760295e3bea5/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation_node.py#L184

emackey:
I'm OK with merging this, and opening a separate bug for that.

julienduroure:
@emackey I already committed a fix in master about import part : 1b34f0f7ab4e8671a99ec724f3a770991362a066

PR #527, opened 2019-06-04
Add morph target animation test

emackey:
This PR introduces a way to decode accessor data (from simple test files only), and uses it to test the round-trip of the morph target animation that was recently fixed.

/cc @donmccurdy @pjoe

### DIFF
--- a/tests/test/test.js
+++ b/tests/test/test.js
@@ -92,11 +92,71 @@ assert.equalEpsilon = function(actual, expected) {
     if (typeof actual !== 'number') {
         throw new Error("Expected " + actual + " to be a number.");
     }
-    let epsilon = Math.abs(expected * 1e-6);
+    let epsilon = 1e-6;
+    epsilon = Math.max(epsilon, Math.abs(expected * epsilon));
     if (Math.abs(actual - expected) > epsilon) {
         throw new Error("Expected " + actual + " to equal " + expected);
     }
 };
+
+assert.equalEpsilonArray = function(actual, expected) {
+    const length = expected.length;
+    assert.strictEqual(actual.length, length);
+
+    for (let i = 0; i < length; ++i) {
+        assert.equalEpsilon(actual[i], expected[i]);
+    }
+};
+
+function getAccessorData(gltfPath, asset, accessorIndex, bufferCache) {
+    // This is only for testing exporter output, it does not handle enough
+    // cases to parse arbitrary glTF files from outside this test suite.
+
+    const accessor = asset.accessors[accessorIndex];
+    const bufferView = asset.bufferViews[accessor.bufferView];
+    const bufferIndex = bufferView.buffer;
+    let bufferData;
+    if (bufferCache[bufferIndex] !== undefined) {
+        bufferData = bufferCache[bufferIndex];
+    } else {
+        const buffer = asset.buffers[bufferIndex];
+        const binPath = path.resolve(gltfPath, '../' + buffer.uri);
+        bufferData = fs.readFileSync(binPath);
+        bufferCache[bufferIndex] = bufferData;
+    }
+
+    const bufferViewByteOffset = bufferView.byteOffset || 0;
+    const bufferViewData = bufferData.slice(bufferViewByteOffset, bufferViewByteOffset + bufferView.byteLength);
+
+    let componentSize;
+    switch (accessor.componentType) {
+    case 5126:  // FLOAT
+        componentSize = 4;
+        break;
+    default:
+        throw new Error("Untested accessor componentType " + accessor.componentType);
+    }
+
+    let numElements;
+    switch (accessor.type) {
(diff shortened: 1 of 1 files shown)


tests/test/test.js
```diff
@@ -92,12 +92,72 @@ assert.equalEpsilon = function(actual, expected) {
...
+    default:
+        throw new Error("Untested accessor type " + accessor.type);
+    }
+
+    // Note that accessor.stride is not handled yet.
+    let count = accessor.count;
+    let stride = componentSize * numElements;
+    let byteOffset = accessor.byteOffset || 0;
+
+    let accessorData = [];
```
donmccurdy: nit: all of the variables above could be `const`, any reason to use `let`?

tests/test/test.js
```diff
@@ -92,12 +92,72 @@ assert.equalEpsilon = function(actual, expected) {
...
+    assert.strictEqual(actual.length, length);
+
+    for (let i = 0; i < length; ++i) {
+        assert.equalEpsilon(actual[i], expected[i]);
+    }
+};
+
+function getAccessorData(gltfPath, asset, accessorIndex, bufferCache) {
+    // This is only for testing exporter output, it does not handle enough
+    // cases to parse arbitrary glTF files from outside this test suite.
```
donmccurdy: More idle curiosity than an actual suggestion, but does glTF-Pipeline have a programmatic API we could be using here? I.e. import it from NPM, use it to parse the file, and provide some accessor data?

I've written some code of my own along these lines, which we could also use here, but glTF-Pipeline is certainly the more mature option.

```js
import { NodeIO } from '@gltf-transform/core';

const io = new NodeIO(fs, path);
const container = io.read('test.gltf');
const array = container.getAccessorArray( accessorIndex ); // TypedArray
```

tests/test/test.js
```diff
@@ -463,6 +523,37 @@ describe('Importer / Exporter (Roundtrip)', function() {
...
+                assert.strictEqual(animation.channels[0].sampler, 0);
+                assert.strictEqual(animation.channels[0].target.path, 'weights');
+                assert.strictEqual(animation.samplers[0].interpolation, 'CUBICSPLINE');
+
+                const animatedNode = asset.nodes[animation.channels[0].target.node];
+                const animatedMesh = asset.meshes[animatedNode.mesh];
+                const targetNames = animatedMesh.extras.targetNames;
+                assert.strictEqual(targetNames.length, 2);
+                assert.notStrictEqual(targetNames.indexOf('Top'), -1);
+                assert.notStrictEqual(targetNames.indexOf('Bottom'), -1);
```
donmccurdy: Does this test still pass if we expect the indices to be 0 and 1, respectively? I think we'd had a bug in the old exporter with the order of morph targets being unstable.

pjoe:
Also take a look at https://github.com/KhronosGroup/glTF-Blender-IO/pull/490/files#diff-6761cb0096988e34d5a544ab980a38de

My approach was a more general 'compare gltf' function, that would basically use deepEqual. I added de-refencing of buffers/views/accessors so their indices would be irrelevant.

However it fails for 2.79, most likely because of unstable vertex order - 2.80 seems solid though. Haven't gotten around to investigating further yet.

tests/test/test.js
emackey: I just asked someone who still part-time maintains gltf-pipeline, sounds like there's no public API for this, unfortunately.  There's internal API for it, not exposed.

But for the short term I think our tests should be able to get pretty far with a cheap decoder.  The bigger problem for us is likely to be the actual vertex order (or morph target order) being unstable.  In this particular test case, two morph targets have the same key/value data, so having the order swap makes no difference in the output.

emackey: I'm not opposed to using your gltf-transform thing, if you think that's mature enough to use here (certainly more mature than what I've written in this PR, I would imagine)

tests/test/test.js
emackey: This is just checking `indexOf(...) !== -1`, meaning testing for presence in the list, regardless of order.

tests/test/test.js
emackey: Fixed.

emackey:
I left some comments on 'golden file' tests on #490.  I have concerns about that approach but I could still be talked into including them.

I do think there's a lot of merit in testing specific values of the feature under test, so I would still like to merge this.  @pjoe @donmccurdy OK to merge this?

tests/test/test.js
donmccurdy: Right, I guess I mean that if the test _does_ pass with a strict requirement on order, we should probably strictly test the order. Maintaining consistent output order is important, because the only other indication of which morph target is which is the `extras.targetNames` array, which isn't an official spec feature.

donmccurdy:
Good to have these tests, yes – thanks!

Issue #540, opened 2019-06-10
AutoSmooth normals do not export correctly


pennomi:
**Describe the bug**
The AutoSmooth normals feature does not export correctly, even with "Apply Modifiers" selected.

**To Reproduce**
1. Download and open the attached Blend file.
2. Export as GLB.
3. Open in a GLTF viewer of your choice
4. The file's normals are all smoothed, instead of having sharp edges where the AutoSmooth feature shows they should be in the Blender viewport.

**Expected behavior**
The exported GLB should have sharp edges in the same places that the Blend file does. Old versions of the GLTF exporter had this, but I don't know what version broke it.

**Screenshots**
![a](https://user-images.githubusercontent.com/832972/59212483-d43e6380-8b6f-11e9-8991-993d62d71aba.png)
![b](https://user-images.githubusercontent.com/832972/59212485-d43e6380-8b6f-11e9-8866-7ad8dd159636.png)

**.blend file/ .gltf**
[minimal.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3273036/minimal.zip)

**Version**
 - OS: linux
 - Blender Version: 2.80 (2019-06-02)

**Additional context**
Note that this uses the features in the Object Data tab, under the Normals and Geometry Data sections. We "Clear Custom Split Normals" and then turn on "Auto Smooth". If I press the "Add Custom Split Normals Data" button before exporting, it exports correctly.

A very old version of the 2.7 plugin handled this correctly. The 2.8 version that shipped with a recent beta handled it correctly as long as the "Apply Modifiers" option was ticked. The latest master of this repo doesn't export correctly no matter what options are set.

Vinc3r:
thanks for the "Add Custom Split Normals Data" button tip, a lot more useful than applying an edge split modifier :p

Issue #546, opened 2019-06-13
"Apply modifiers" option does nothing


iirelu:
When exporting meshes with modifiers, the "Apply modifiers" option doesn't actually apply any modifiers. The original mesh is the only thing exported.

Reproduce by putting a bevel modifier on a cube and exporting to glb/gltf with "Apply modifiers" on. The resulting gltf file has the cube's original geometry instead of the modified geometry.

Blender 2.80.190609.030c7df19da9-1, glTF exporter 0.9.24

julienduroure:
Hello,
Thanks for reporting. 
This is a know issue, duplicate of first part of #525 , and #536

julienduroure:
Fixed by #547

Issue #536, opened 2019-06-08
No modifier applied


pafurijaz:
**Describe the bug**
I tried to export the file attached but time ago I had a success the exporting of intersecting planes are not supported in this particular case.

**To Reproduce**
Try to export the file attached, I have more files with problems if is useful I can share..

[Stingray.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3268960/Stingray.blend.zip)

**Version**
 - OS: Linux 
 - Blender Version version: 2.79 (sub 7), branch: blender2.7, commit date: 2019-05-23 08:19, hash: 054dbb833e15, type: Release
build date: 2019-06-05, 00:16:26
platform: Linux

julienduroure:
Fixed by #547

PR #547, opened 2019-06-13
Cache enhancement

julienduroure:


julienduroure:
@pjoe @emackey @donmccurdy Please have a look on this change on cache. Quite simple for now, let see if we need to go further.
This fix #526 regression, and fix #540 #546 #536

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_cache.py
```diff
@@ -37,8 +37,20 @@ def wrapper_cached(*args, **kwargs):
             export_settings = args[-1]
             cache_key_args = args[:-1]
 
+        __by_name = ['Object', 'Scene', 'Material', 'Action', 'Mesh']
+
         # we make a tuple from the function arguments so that they can be used as a key to the cache
-        cache_key = tuple(cache_key_args + tuple(cache_key_kwargs.values()))
+        cache_key = ()
+        for i in cache_key_args:
+            if str(type(i).__name__) in __by_name:
```
pjoe: I wonder if it isn't better to compare types directly instead of by name?along the lines of:
```
type(i) in [bpy.types.Object, bpy.types.Mesh, ... ]
```

PR #550, opened 2019-06-15
Add test to see if mesh instances have re-used data.

emackey:
This almost slipped under my radar, but @pjoe added a new test scene in #526 called `02_material_instancing.blend`.

In this PR I've added some test code to dig into the contents of the exported scene, to find out if mesh instancing has really taken place.  As with other tests, I'm using the pattern of filtering by name for the objects of interest, so that export output re-ordering won't break this test.

### DIFF
--- a/tests/test/test.js
+++ b/tests/test/test.js
@@ -202,6 +202,49 @@ describe('Exporter', function() {
             let outDirName = 'out' + blenderVersion;
             let outDirPath = path.resolve(OUT_PREFIX, 'scenes', outDirName);
 
+            it('can create instances of a mesh with different materials', function() {
+                let gltfPath = path.resolve(outDirPath, '02_material_instancing.gltf');
+                const asset = JSON.parse(fs.readFileSync(gltfPath));
+
+                assert.strictEqual(asset.meshes.length, 4);
+                assert.strictEqual(asset.materials.length, 3);
+
+                const materialRed = asset.materials.filter(m => m.name === 'MaterialRed')[0];
+                const materialGreen = asset.materials.filter(m => m.name === 'MaterialGreen')[0];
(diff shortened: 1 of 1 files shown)


donmccurdy:
Thanks!

pjoe:
Great :)

I do think though that we are getting to the point where `test.js` is ripe for being split up and re-organized. Will think a bit about this and maybe do an initial attempt.

emackey:
@pjoe That would be a good thing, of course.  I have a few more things I want to get in soon, in particular coverage of the Emissive channel, because they added a new node input for that in 2.80 that we don't support yet.

Along those lines, currently we have glTF round-trip tests, and export-only tests for 2.79 scenes.  We don't have any export tests for 2.80-specific scenes (they would fail to load in 2.79 and break the test suite).  If we're going to test newly-added things like the Emissive input on the Principled BSDF node, we'll need a way to keep those 2.80 projects separate.  I was thinking of adding a `scenes_280` folder or similar, but it feels a little messy.

pjoe:
hmm yeah those multiple blender version does make things complicated. I wonder how long it makes sense to maintain 2.79 as 2.80 final is scheduled for July.

Another thing I had been considering is how to test multiple export settings, as these can have quite an impact.

emackey:
Unfortunately I don't think we're dropping 2.79 this year.  I love 2.80 but not all users will migrate quickly, because the UI needs to be almost re-learned for some workflows.  Hopefully 2.79 can be let go next year.  We'll see.

Multiple export settings is a good idea too.  I was wondering if that can be tested by checking "Remember Export Settings" (in the glTF export panel) and then saving the Blender project.

pjoe:
Yeah we need to keep 2.79 support until the time is right :)

Having settings saved as part of .blend would work, but then again would require multiple .blends to test same content with different settings, so was thinking of finding a good way to pass in settings from js side of things.

PR #552, opened 2019-06-16
Import: fix viewport alpha for KHR_materials_unlit

scurest:
This regressed in 78e6f74. `blender_alpha` wouldn't ever get called if KHR_materials_unlit was present.


donmccurdy:
Thank you!

emackey:
Thanks @scurest.

We should add a test to prevent this regression happening again.

PR #553, opened 2019-06-17
Import: support primitive modes other than 4

### LINKED ISSUE #315: Only mode 4 is managed in primitive importer

scurest:
Fixes #315. I tested with some of [these files](https://github.com/KhronosGroup/glTF-Asset-Generator/tree/master/Output/Positive/Mesh_PrimitiveMode).

The strategy is just to first convert the indices to the equivalent LINES and TRIANGLES representation, then proceed as normal.

Questions

* Is that a good place to put the helper function?
* The `pyprimitive.faces_length` field is assigned to in this function but grep tells me it isn't ever used anywhere else. What's that for? For consistency, I assigned to a `pyprimitive.edges_length` field too.
* For unknown modes, I raised an exception, but it could also make sense to log a warning and skip the primitive.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_primitive.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_primitive.py
@@ -30,7 +30,6 @@ class BlenderPrimitive():
         """Primitive creation."""
         pyprimitive.blender_texcoord = {}
 
-        # TODO mode of primitive 4 for now.
         current_length = len(verts)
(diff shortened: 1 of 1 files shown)


julienduroure:
Thanks for the PR. I will have a look on it soon

scurest:
@julienduroure How's this going?

julienduroure:
@scurest Sorry, I am away from keyboard during part of the summer, checking bugs/PR/coding can take longer than usual.

PR #556, opened 2019-06-18
Fix transforms of children of bones

xissburg:
When an object is added as a child of a bone I was getting incorrect orientation in the output. The solution was to use the local matrix instead.

Some unnecessary code was found and removed in the process and another minor bug/crash was also fixed.


julienduroure:
Hello,
Thanks for your PR. Can you please provide 2 .blend files that show the 2 issues you are trying to solve ?

Thanks!

xissburg:
I have failed to create a simple test case for this. This bug is only concurring in my own model (which I cannot share). There's something particular to the skeleton in this mesh which causes the bug and I still can't figure out what it is after a few hours of debugging. Fortunately this change in the code allows me to export it correctly for the time being. I will close this issue for now.

PR #559, opened 2019-06-20
Check for missing textures and unassigned shader nodes

### LINKED ISSUE #345: Material export error (parrot model)
Let's take a look at this parrot model:
[parrot.zip](https://github.com/KhronosGroup/glTF-Blender-Exporter/files/2495150/parrot.zip)
load it and select the animation called "surprise" in the dope sheet. Take a look at the wings:
image
Now export it to GLTF format with the following settings enabled:
image
Place the resulting GLTF file in the Godot game engine, select the animation "surprise" in the AnimationPlayer node and take a look at the wings:
image
Tail motion isn't correct too, it seems that the skeleton does not follow the geometry well (somehow it outside the parrot body):
image
Maybe it is really a bug in the Godot game engine, but I don't know another way to test the resulting GLTF :)

### LINKED ISSUE #543: T65622: Traceback when exporting
Issue from blender tracker : https://developer.blender.org/T65622

> resolution = __get_tex_from_socket(blender_shader_sockets_or_texture_slots[0]).shader_node.image.size
> AttributeError: 'NoneType' object has no attribute 'size'

emackey:
We were already printing a warning message, but then crashed.  This should protect against the crashes.

Also, added a new test scene that has a missing PNG file and an Image Texture shader node that is not assigned an image.  No specific test code is needed, we just allow the normal test suite functionality to test whether this crashes on export (which would fail the test).

Fixes #345, fixes #543 ... like a palindrome


emackey:
@julienduroure This is ready to merge, it's just some guards against `None` values.  Any chance it can go into RC1 or RC2?

donmccurdy:
I'm assuming this will need to go into 2.81 at this point, but the changes look good to me.

PR #574, opened 2019-06-28
Export materials with alphaMode MASK if their Blend Mode is "Alpha Ha…

atteneder:
…shed".

Blender has a Blend Mode named `Alpha Hashed`, which is a more complex alpha testing/clipping (source https://www.katsbits.com/codex/alpha). Thus exports should set the alphaMode to MASK.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
@@ -26,6 +26,7 @@ from io_scene_gltf2.blender.exp import gltf2_blender_gather_materials_pbr_metall
 from io_scene_gltf2.blender.exp import gltf2_blender_generate_extras
 from io_scene_gltf2.blender.exp import gltf2_blender_get
 
+ALPHA_MASK_METHODS = ['CLIP','HASHED']
 
 @cached
 def gather_material(blender_material, mesh_double_sided, export_settings):
@@ -80,7 +81,7 @@ def __gather_alpha_cutoff(blender_material, export_settings):
     if bpy.app.version < (2, 80, 0):
         return None
     else:
-        if blender_material.blend_method == 'CLIP':
+        if blender_material.blend_method in ALPHA_MASK_METHODS:
             return blender_material.alpha_threshold
     return None
 
@@ -89,7 +90,7 @@ def __gather_alpha_mode(blender_material, export_settings):
     if bpy.app.version < (2, 80, 0):
         return None
     else:
-        if blender_material.blend_method == 'CLIP':
+        if blender_material.blend_method in ALPHA_MASK_METHODS:
             return 'MASK'
         elif blender_material.blend_method == 'BLEND':
             return 'BLEND'


atteneder:
It seems like the CI failed for different reasons (downloading/installing Blender failed). How can I re-trigger the CI?
Or can someone fix it and re-run it?

donmccurdy:
Thanks! I agree that we should do something useful with the *Alpha Hashed* setting, but I'm inclined to say that exporting it as `BLEND` would be closer to most authors' intents than `MASK`. From the article you link to, compare these results:

blend-mode

And from the [Eevee material docs](https://docs.blender.org/manual/en/dev/render/eevee/materials/settings.html),

> **Alpha Hashed**
> The previous color will be overwritten by the surface color, but only if the alpha value is above a random clip threshold. __This statistical approach is noisy but is able to approximate alpha blending without any sorting problem__. Increasing the sample count in the render settings will reduce the resulting noise.

As *Alpha Hash* is an approximation of alpha blending, it seems better to export it as such.

***

Somewhat related — does anyone know if this *Alpha Hashed* setting is the same thing as Unreal Engine's *Dither Temporal AA* translucency? See [this article](http://danieru.com/2018/07/04/dither-based-transparency-in-unreal-engines-material-editor/). If there's a common theme across engines here I'd be tempted to consider this in a glTF extension at some point...

atteneder:
Thanks for the comment and the interesting links. Admitted, I did not read the eevee material docs before.

I agree that BLEND makes sense for some cases, but it will definitely break others. The dilemma is that Blender-Eevee users will use HASHED on all kinds of alpha maps: ones that look better with MASK and ones that look better with BLEND when exported to glTF.

Proposed solution:
- Define a default fallback for HASHED (either MASK or BLEND)
- Have an glTF export operator property that lets the user change the default
- Additional per-material custom property / overrides

To have consistent output throughout sessions the latter would have to be saved with the blend file.

About picking the default fallback:
For most of the data I work with I think MASK would be the better choice.
I did not do my cause a favour by picking that particular example, since the clip threshold for the "Alpha Clip" clearly isn't properly adjusted. If it was, the edges and the text would be fully opaque and the rest fully transparent and thus the the CLIP result would be much closer to HASHED.

Let me go into details about my motives:
I'm doing a lot of automated imports of assets/materials with both CLIP and MASK alpha maps. I intend to use HASHED for all these materials, since it looks correct in all cases in Eevee renders.

For glTF exports there is no ideal setting for all cases, but I'm thinking about trying to estimate it per material by evaluating the histogram of the alpha channel. Lots of extreme values (black/white) indicates that MASK looks better while more mid-grey values probably suggest BLEND is adequate.
I presume this approach would have a fairly low failure rate, but will never be 100% correct.
Still good enough for my custom import scripts, but probably not a general purpose solution for glTF export.
But in order to achieve this, I need some sort of control over the glTF result (from script), hence the per-material override in my proposal.

Thanks for giving it a thought,

donmccurdy:
Could you share an example in which a file exported to glTF with `alphaMode=MASK` will look closer to the original than one using `BLEND`, where the original material in Blender uses *Alpha Hashed*?

scurest:
I don't think it's right to use `blender_material.alpha_threshold` as the alpha cutoff if you export it as `MASK`. That field isn't used at all in Alpha Hashing mode so it's essentially a meaningless number. Arguably it would be better to pick the expected value of the random threshold for Alpha Hashing (ie. 0.5, which is also the default cutoff in glTF, so you could just return `None`).

The lack of a natural way to pick this cutoff is another reason to favor `BLEND` IMO.

atteneder:
> Could you share an example in which a file exported to glTF with `alphaMode=MASK` will look closer to the original than one using `BLEND`, where the original material in Blender uses _Alpha Hashed_?

Sure.

This plant has alpha sorting problems (even in eevee-blend):
plant_eevee

Viewed in [your viewing tool](https://gltf-viewer.donmccurdy.com) it gets worse, because the shader writes to the depth buffer untested:
plant_3js

p.s.: Thanks for the awesome glTF viewer. I use it all the time!

atteneder:
> I don't think it's right to use `blender_material.alpha_threshold` as the alpha cutoff if you export it as `MASK`. That field isn't used at all in Alpha Hashing mode so it's essentially a meaningless number. Arguably it would be better to pick the expected value of the random threshold for Alpha Hashing (ie. 0.5, which is also the default cutoff in glTF, so you could just return `None`).
> 
> The lack of a natural way to pick this cutoff is another reason to favor `BLEND` IMO.

I get your point!

But as a user who wants to benefit from alpha-hashed within Eevee whilst wanting to fall back to glTF-MASK upon export (either default or override) I want to be able to control glTF's alphaCutoff value. The average of 0.5 that alpha-hashed uses might or might not look good on my particular alpha maps.

Solutions:
- Take Blender's Clip Threshold (even though it's not used for eevee-hashed; will probably be untouched/0.5 in most cases anyways)
- Yet another custom property that has to be set/maintained.
- Pick BLEND as convenient default and ignore my use cases entirely (as you said)

atteneder:
I acknowledge that my use cases are somewhat corner case and introduce a lot of complexity and doubtful default behaviour.

Maybe, for custom pipelines like mine, it is safer/easier to correct blender's material settings prior to glTF export (or Eevee render) via custom scripts.
For example changing alpha-hashed to whatever I want it to be before exporting glTFs (or always switch to alpha-hased before rendering in Eevee).

Still glad I kicked off the discussion. I'll be happy with either result and sure a good option will be picked.

scurest:
No, you're right, it makes perfect sense if you think of it as Alpha Hashed materials are always exported _as if_ they had been switched to Alpha Clip. That's really simple to explain and understand, and it enables a workflow where you can always just switch to Alpha Clip to see how things will look and adjust the cutoff.

I was thinking of exporting it in a way that looks like Alpha Hashing looks. `BLEND` is supposed to look correctly sorted and blended, which is the thing Alpha Hashing is approximating, but glTF implementations will also only approximate this ideal output, and in different ways. The spec [says for real-time implementations](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#alpha-coverage)

> There is no perfect and fast solution that works for all cases. Implementations should try to achieve the correct blending output for as many situations as possible. Whether depth value is written or whether to sort is up to the implementation. For example, implementations can discard pixels which have zero or close to zero alpha value to avoid sorting issues.

An implementation which sorted (or uses alpha hashing!) would render your plant correctly in `BLEND` mode of course.

donmccurdy:
> This plant has alpha sorting problems (even in eevee-blend)...

This looks, to me, like a perfect example of a case where *Alpha Clip* should be used in Blender, which would of course export to glTF's `MASK` mode. *Alpha Hash* works better than *Alpha Blend*, yes, but *Alpha Clip* is the more appropriate choice. Because Alpha Hash is effectively an approximation of Alpha Blend, without the sorting issues, it still seems more appropriate to me to export it as Alpha Blend and to leave further modifications to user pipelines.

The idea of an explicit alpha hash or dither feature in glTF would be interesting to explore, at some point.

scurest:
FWIW, user extensions were just merged and you can use them for this. Save this as a .py and install it as a normal Blender addon.

````py
bl_info = {
    'name': 'glTF 2.0 Export: Export HASHED as MASK',
    'blender': (2, 80, 0),
    'description': (
        'glTF-Blender-IO User Extension: '
        'Export materials with "Alpha Hashed" blend mode as if they used "Alpha Clip".'
    ),
    'category': 'Import-Export',
}

    def gather_material_hook(self, material, blender_material, export_settings):
        if blender_material.blend_method == 'HASHED':
            material.alpha_mode = 'MASK'
            material.alpha_cutoff = blender_material.alpha_threshold

def register():
    pass

def unregister():
    pass
````

atteneder:
I'd say this extension system is a better way of creating customized behaviour

PR #584, opened 2019-07-10
Fix missing param in recursive draco traverse_node

pjoe:
Causes error if exporting a file with parented meshes using draco


pjoe:
Gahh, just noticed this is identical to #581, closing this

Issue #583, opened 2019-07-09
Animation export is broken


ZanQdo:
**Describe the bug**
Simple animation is not exported correctly from Blender

**To Reproduce**
Export an animated cube and import back.
The action is not imported at all.

When opening the file on https://sandbox.babylonjs.com/ the action will be present but the animation count will be 0, and the end frame will be a negative value. The action will not play.

![image](https://user-images.githubusercontent.com/999476/60926685-91ea6e00-a264-11e9-9834-2d809e182491.png)

![image](https://user-images.githubusercontent.com/999476/60926672-826b2500-a264-11e9-8291-ef85542c3c2d.png)

These files where created with a recent Blender 2.8 build

.blend file
[AnimExport.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3374984/AnimExport.zip)

.gltf file
[anim.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3370872/anim.zip)

**Version**
 - Blender Version 2.80 (2019-07-08)

MiikaH:
The attached .gltf file seems to contain two duplicated versions of the node. Both share identical name and mesh, but only other one is animated. Only the non-animated version is included in the scene.

ZanQdo:
Confirmed, after manually removing duplicate nodes the files load correctly

pjoe:
Seems to be caused by #561. The node cache breaks because the animation gather now calls with different params (None for blender_scene) and then creates a new node instead of using the existing one :S

ZanQdo:
I hope this can be fixed soon before the Blender RC since animation export is completely broken even on the simplest of cases

emackey:
The RC is due out today.  I don't think this fix is going in RC1, but I would expect some more bugs & fixes to come in before the next Release Candidate comes along.

jedStevens:
I would also like to chime in about the broken animation.

Here is a gif showing what the animation looks like after it has been imported to the Godot engine. When I import it back into blender there is no animation at all, but I mostly care about getting my animations **out** of blender.

![Peek 2019-07-11 15-55](https://user-images.githubusercontent.com/6509081/61081195-a9f0f780-a3f4-11e9-9551-0c69e1202ead.gif)

This just seems wrong. I'm noticing that the bouncing up and down in the torso should instead be a rotation as I didn't animate an idle cycle this terrible.

Any ideas?

julienduroure:
[...]

PR #585, opened 2019-07-10
Workaround for broken anim export #583

pjoe:
My attempt at fixing #583.

This is not very pretty though :S

Also not sure how well this will work if animated object is linked in multiple scenes.

Maybe a more 'correct' behavior would be to pass the blender scene down through all the animation gather functions, though that also seems messy ... hmmm 🤔


julienduroure:
I will commit this workaround tonight, to be included in final 2.80 that should come in the next 2 weeks. Then we will have more time to think about a better solution if needed

PR #586, opened 2019-07-10
Add support for Quaternion and None assignments to keyframe tangent props

### LINKED ISSUE #587: Malformed tangents on first and last frames of exported animations.
**Describe the bug**
Malformed tangents on first and last frames of exported animations.

**To Reproduce**

1. Create simple animation and export
2. Playback @ https://gltf-viewer.donmccurdy.com/
3. Notice wobble about frame 1

**Expected behavior**
Clean animation without wobble

**Screenshots**
See pull request: https://github.com/KhronosGroup/glTF-Blender-IO/pull/586

**Version**
 - OS: Windows
 - Blender Version: 2.80 (2019-07-01)
 - glTF-Blender-IO: master (2019-0710)

**Additional details**
See pull request: https://github.com/KhronosGroup/glTF-Blender-IO/pull/586

Please close this report upon acceptance of the pull request.

tvkit:
Keyframe  tangent property setters assume non-Quaternion arguments resulting in malformed assignments of Quaternions.  Issue manifests during  tangent suppression/zeroing for first and last frames:

_gltf2_blender_gather_animation_sampler_keyframes.py_

```
 # Construct the in tangent
 if frame == frames[0]:
   # start in-tangent should become all zero
   key.in_tangent = key.value
```
and

```
  # Construct the out tangent
  if frame == frames[-1]:
    # end out-tangent should become all zero
    key.out_tangent = key.value
```

The problematic assignment occurs in the following Keyframe method where `value` is effectively being treated as simple vector:

```
    def __set_indexed(self, value):
	# ...
        result = [0.0] * self.get_target_len()
        for i, v in zip(self.__indices, value):
            result[i] = v
        result = gltf2_blender_math.list_to_mathutils(result, self.target)
        return result
```

Support for direct Quaternion assignments was provided in this pull request.

-----

Example export for a simple animation showing the rotation sampler data:

Before:

anim-tangent-prob

After:

anim-tangent-okay

Fixes #587.


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
```diff
@@ -62,12 +62,18 @@ def get_target_len(self):
         return length
 
     def __set_indexed(self, value):
+        if value is None:
```
julienduroure: What is the case where value come with None data?

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
```diff
@@ -62,12 +62,18 @@ def get_target_len(self):
...
+            return value
+        if "rotation" in self.target and type(value) is mathutils.Quaternion:
+            return value
         # 'value' targets don't use keyframe.array_index
         if self.target == "value":
             return value
         # Sometimes blender animations only reference a subset of components of a data target. Keyframe should always
         # contain a complete Vector/ Quaternion --> use the array_index value of the keyframe to set components in such
         # structures
+        if type(value) is mathutils.Quaternion:
```
julienduroure: Not sure to understand why there is this code: the previous check added returns out the function for rotation/Quaternion. WHat is the case where value is a Quaternion and value is not "rotation" ?

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
```diff
@@ -62,12 +62,18 @@ def get_target_len(self):
...
+            value = tuple(value.to_euler("XYZ"))
```
julienduroure: I don't understand why you convert to euler, indexes will not match anymore (if this code is really execute, see previous comment)

julienduroure:
One more comment : the issue can be found also in location, not only rotation

julienduroure:
Closing this PR as I committed a more general fix

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
tvkit: Perhaps it's desirable to reset `value` to its original state which is **None**.  I tried assigning None at one point to simply observe the keyframe results, but the plugin crashed. That shouldn't be an acceptable outcome.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
tvkit: "rotation" targets are implemented in the plugin as Quaternions; thus, it makes sense in the first test to simply return the Quaternion. But if the target is not a "rotation",  code execution continues and resolves for a non-Quaternion target; thus, the second Quaternion test which returns: `tuple(value.to_euler("XYZ"))`

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
tvkit: The existing implementation already returns an array supporting **location**, and **scale** which are effectively XYZ eulers. In the event a Quaternion is passed, the conversion satisfies the expected simple array/euler.


Keep in mind that prior to this patch, the implementation mangled the Quaternion into a euler which caused the related bug.

tvkit:
> Closing this PR as I committed a more general fix

Thank you for reviewing.

PR #591, opened 2019-07-14
Add alpha export support

m0ppers:
Ok....Another try to integrate alpha export (first attempt #466)

Now that the Principled BSDF node contains alpha this is pretty much straightforward.


m0ppers:
any update? is it be possible that this still makes it into the final release of blender 2.80? :S

donmccurdy:
I'm in favor of merging this, but I don't think we should try to get it into the initial 2.80 release, if that's even possible at this stage... better to wait and have a better chance of catching bugs. Thoughts @emackey @julienduroure ?

PR #608, opened 2019-08-04
Fix #599 custom normal export when normals are modified by modifier

### LINKED ISSUE #599: Custom normals are not exported
**Describe the bug**
Meshes lose their custom normals after exporting them from Blender 2.8 RC3

**To Reproduce**
1. Add a mesh (for example, the default monkey Suzanne)
2. Under the Mesh tab, turn on Auto Smooth
3. Add a 'Normal Edit' modifier, leave everything on default
4. Apply the modifiers
5. Export the object

**Expected behavior**
When loading the resulting glb/gltf file, custom normals should be apparent, instead they still look like the default ones.

**Screenshots**
Same object exported to Collada and gltf :
image

The custom normals as seen from Blender 2.8 :
image

Export options used :
image

**.blend file/ .gltf**
Minimal blend file :
[custom_normal_minimal.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3439107/custom_normal_minimal.zip)

Resulting glb file after export:
[exported_glb.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3439123/exported_glb.zip)

**Version**
 - OS : Windows 10 64bits
 - Blender Version : 2.8 RC3

**Additional context**
+ I previewed the exported glb file in the following applications:
  - Godot engine
  - https://sandbox.babylonjs.com/
  - https://gltf-viewer.donmccurdy.com/
None of them could display the custom normals. 
+ I tried exporting the same object to a dae file and it worked in Godot.
+ I tried both glb and gltf format, but there were no custom normals in both cases.

julienduroure:


julienduroure:
@donmccurdy @emackey @pjoe I'd like you to test non regression on it, because there is so much cases. I tried to cover them, but please have a look :)

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
```diff
@@ -248,7 +253,7 @@ def __gather_mesh(blender_object, export_settings):
             edge_split = blender_object.modifiers.new('Temporary_Auto_Smooth', 'EDGE_SPLIT')
             edge_split.split_angle = blender_object.data.auto_smooth_angle
             edge_split.use_edge_angle = not blender_object.data.has_custom_normals
-            blender_object.data.use_auto_smooth = False
+            blender_object.data.use_auto_smooth = any([m in modifier_normal_types for m in [mod.type for mod in modifiers]])
```
pjoe: This fails when there are no modifiers (see line 247), maybe use `blender_object.modifiers` instead :)

pjoe:
Code changes looks good to me (except the issue when there aren't any modifiers as commented above).

From some limited testing I am sometimes seeing some weird behavior, e.g. that culling (or guess it's polygon winding) being affected by custom normals. I've attached a blend with `Edit Normals`with a directional target that looks really weird in babylon and Don's viewer after export.

[custom_normals_599.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3467804/custom_normals_599.zip)

julienduroure:
Any other comments before I merge it?

donmccurdy:
> e.g. that culling (or guess it's polygon winding) being affected by custom normals...

Sounds like polygon winding, yes. 😕

I don't have test cases for this; maybe we should start a wishlist of regression tests we would like to get into the addon's test framework? Meanwhile I'm OK with merging this.

emackey:
> > e.g. that culling (or guess it's polygon winding) being affected by custom normals...
> 
> Sounds like polygon winding, yes. 😕

Yes winding, plus that model lacks a Blender material.  The default Blender material is double-sided, and the default glTF material is single-sided, so you're seeing the effects of backface culling on Suzanne's flipped normals.  If you assign a Blender material, they will match, with and without backface culling.

Merging.

emackey:
Thanks @julienduroure!

PR #623, opened 2019-08-19
Fix #620 shapekey anim

### LINKED ISSUE #620: Shape Keys applied to wrong target in Blender 2.8
Exporting a GLB out of Blender applies the blend shape attributes to the wrong morph target. 

In my example, the "Glee Eyes" target is being applied to the "Super Open" target. We have tried removing the "Super Open" target completely and re exporting and it just gets applied elsewhere.

Blend and GLB file: https://drive.google.com/drive/folders/14YZx1q0INs-SvKSFIdRv8dy_2QTbapA0?usp=sharing

julienduroure:


julienduroure:
This will fix #620

Issue #635, opened 2019-08-27
base64 decode expects "data:base64;..." so "data:,..." doesn't work


rorlork:
**Describe the bug**

The data URI scheme, according to RFC 2397 ( https://tools.ietf.org/html/rfc2397 ) is data:[<mediatype>][;base64],<data> but the code always tries to find ";base64," so any encoded data of the form "data:,..." will not be parsed.

**To Reproduce**
1. Export a scene with "gltf Embedded"
2. Edit the file and remove from any uri-encoded buffer the section between "data:" and ","
3. Try to load the scene

**Expected behavior**
The scene should be loaded correctly

**.blend file/ .gltf**

[gltf_base64_fail.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3545194/gltf_base64_fail.zip)

**Version**
 - OS: Windows 10
 - Blender Version 2.80

**Additional context**

I will create a pull request for a fix for this issue

lexaknyazev:
The existing implementation of Data URI loader doesn't cover the whole RFC 2397, but it is correct wrt looking for `;base64` string since only in that case the URI is encoded with base64:
> The appearance of `;base64` means that the data is encoded as base64. Without `;base64`, the data (as a sequence of octets) is represented using ASCII encoding for octets inside the range of safe URL characters and using the standard `%xx` hex encoding of URLs for octets outside that range.

Given that glTF uses URI only for binary data, there's no practical reason to support percent-encoded Data URIs in the importer.

rorlork:
You are totally right. I think the problem came when we used an asset exported from another editor which didn't use the ";base64" string so it broke the loader.
I should have read more carefully the RFC, sorry for the inconvenience.

PR #636, opened 2019-08-27
Add support for data uri without ";base64" section

rorlork:
Fix for #635


rorlork:
Yes, closing it since the issue was not valid.

PR #637, opened 2019-08-28
Merge animations

julienduroure:


julienduroure:
Not ready to be merged yet:

- [ ] Fix non regression tests
- [ ] Discuss and code how to merge extras & extensions
- [ ] Proposition to add an option to merge all active actions. That means that non-expert users (that don't use NLA), can create a file with single animation, as displayed in viewport

julienduroure:
@donmccurdy  @emackey 

Based on current PR, I propose to add an option in exporter, to be able to merge all animations of the blend file into a single animation. This action will ignore all actions in NLA, and merge only active actions. Any thoughts ?

donmccurdy:
Wonderful! 👏

> Proposition to add an option to merge all active actions...

Sounds good! I'd suggest we name the option "NLA Strips", enabled by default. When enabled, non-muted NLA strips are merged by name and exported as separate glTF animations. When disabled, a single global animation is created. This matches an option in the FBX exporter nicely.

julienduroure:
Seems this is ready to be tested for everyone! @donmccurdy

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
```diff
@@ -45,29 +48,37 @@ def gather_animations(blender_object: bpy.types.Object, export_settings) -> typi
         # Set action as active, to be able to bake if needed
         if blender_object.animation_data: # Not for shapekeys!
             if blender_object.animation_data.action is None \
-                    or (blender_object.animation_data.action.name != blender_action.name):
+                    or (blender_object.animation_data.action.name != blender_action[0].name):
```
scurest: I think it would be better here to change the loop above to

    for blender_action, track_name in blender_actions

You wouldn't need to insert all the `[0]`s and the part below with `blender_action[1]` would be clearer too.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
```diff
@@ -20,7 +20,10 @@
 from io_scene_gltf2.io.com.gltf2_io_debug import print_console
 
 
-def gather_animations(blender_object: bpy.types.Object, export_settings) -> typing.List[gltf2_io.Animation]:
+def gather_animations(blender_object: bpy.types.Object,
+                        tracks: typing.Dict[str, typing.List[int]],
+                        offset: int,
+                        export_settings) -> typing.Tuple[typing.List[gltf2_io.Animation], typing.Dict[str, typing.List[int]]]:
     """
     Gather all animations which contribute to the objects property.
```
scurest: This doc comment became out of date.

elbobo:
Hi - is it now possible to export with a global animation? I am using blender 2.81 and I don't see an option in the exporter, and it will not allow me to rename the NLA strips to be the same as someone suggests above.. Thanks for any advice.

donmccurdy:
@elbobo how is it not allowing you to rename NLA tracks? The export isn't right when you rename, or you are prevented from renaming them somehow? 

That is, at present, the only way to export a single global animation affecting multiple objects.

elbobo:
Hi @donmccurdy thanks for replying. Apologies, it does work - I was confused as to what was considered a 'track'. I'm just going to note this down for anyone having a similar issue. 

I was trying the rename either of these for example...

Screenshot 2020-02-16 at 13 10 14

And those cannot be named the same thing. I needed to 'push down' so that it ends up like the below..

Screenshot 2020-02-16 at 13 10 24

And then that newly created item can be renamed, and the animations are grouped together which is great.

I wonder if you could answer a few questions about this though when you have a moment, and apologies if I am missing something fundamental here but I don't really understand the NLA context.

a) do I need to 'push down' in order to rename these tracks? Or is there some other way to access them? Or do they not exist as tracks until they have been pushed down?
b) when pushed down, by default they all have the same name 'NlaTrack' but am I right in thinking this default same name doesn't allow them to be grouped? You actually have to change all the names manually? It doesn't work for me to simply push them all down, despite the fact that by default they do all end up with the same name.
c) do you know if there's a way to loop through these tracks and rename them programatically? There are some instances where I have a lot of animations and this process can become a bit laborious. I checked the console when I renamed them to see what the command might be but it just reads

`bpy.context.object.name = "test"`

but if I run a script saying the same thing it just renames the object in the scene collection, not the strip.

Thanks for any advice, a steer on programatically renaming these (including pushing down if thats necessary?) would be ideal.

Thanks for the exporter by the way - its great!

elbobo:
Hi, re: my above queries, I did a bit of investigation, the following script (taken from [here](https://blender.stackexchange.com/questions/88620/how-can-i-push-down-actions-to-nla-strips-with-python) and tweaked slightly) seems to do the job of renaming the strips and exports as one global animation. Adding here for others.

```python
import bpy

animationName = 'globalAnimation'

for obj in bpy.context.scene.objects:
    if obj.animation_data is not None:
        action = obj.animation_data.action
        if action is not None:
            track = obj.animation_data.nla_tracks.new()
            track.strips.new(action.name, action.frame_range[0], action)
            track.name = animationName
            obj.animation_data.action = None
```

PR #640, opened 2019-08-29
Fix #619 Set active blender scene in gltf file as main scene

### LINKED ISSUE #619: Multiple scene export : make active scene the active scene in glTF file
Multiple scene export : make active scene the active scene in glTF file

julienduroure:


emackey:
Fix #619 (repeating this here because the PR title won't autolink to an issue.)

PR #648, opened 2019-09-01
Export only bones related to action.

pgruenbacher:
exporting only bones related to the action cuts the size of my asset file by more than half and speeds it up quite a bit too.


julienduroure:
Unfortunately, this PR can't be accepted. 
Export only bones related will break most of animation export, when deformations bones are not keyed. (And not sure joint structure is handle correctly in your patch (not verified))

The solution is to export only deformation bones, the patch is WIP on my side

pgruenbacher:
no worries figured I'd illustrate

PR #651, opened 2019-09-04
Add "nla strips " and "all actions" options to the exporter ( similar to the FBX exporter)

JeckleCed:
This pull request adds two options to the exporter:

**NLA strips**:
It is on by default to keep the current default behaviour of the exporter.
it exports each non-muted NLA strip as a separated animation, if any, instead of global scene animation

**All actions**:
When on this exports each actions as individual animation.
When off and NLA strip is off, then all animations will be merged into one single with multiple channels

image

I tried to put a lot of care to avoid changing existing behaviour and code :)
This has been tested with Blender 2.8

Some previous discussion related to NLA strips and all actions can be found here:
https://github.com/KhronosGroup/glTF-Blender-Exporter/pull/166

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -261,6 +261,18 @@ class ExportGLTF2_Base:
         default=True
     )
 
+    export_nla_strips: BoolProperty(
+        name='NLA strips',
+        description='Export each non-muted NLA strip as a separated anim, if any, instead of global scene animation',
+        default=True
+    )
+
+    export_all_actions: BoolProperty(
+        name='All actions',
+        description='Export all actions as individual animations',
+        default=False
+    )
+
     export_all_influences = BoolProperty(
         name='Include All Bone Influences',
         description='Allow >4 joint vertex influences. Models may appear incorrectly in many viewers',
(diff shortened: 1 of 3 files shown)


julienduroure:
Hello, #637 is already in progress about same options. Closing this PR to avoid having two different PR about same subject, but discussion can continue on #637

JeckleCed:
@julienduroure, #637 seem to only include the "nla strips" options, while mine also add the "all actions" ones. Do you prefer to have a separate PR just for the "all actions" option when #637 completes?

julienduroure:
In #637 when nla strips is disabled, all actions are merged. It seems to be your "all actions" option.
I propose to test carefully your PR, and then check if there is something to port to #637

donmccurdy:
The FBX exporter's *All Actions* option mentions that "animated objects will get all actions compatible with them." This has the potential to duplicate animation data – it might be worth discussing how it would work, before spending too much time porting it to #637. I do agree with adding the *NLA Strips* option, which (when disabled) also lets you merge everything to one animation.

julienduroure:
@donmccurdy this "NLA strips" is already done in #637 . #637 is ready to merge after your review

PR #653, opened 2019-09-04
Fix ibm

julienduroure:


julienduroure:
Still need to be heavily tested with all files with armature / skin data

Issue #672, opened 2019-09-12
Test failures when downloading Blender


emackey:
Recent tests have all failed when trying to download Blender.

https://circleci.com/gh/KhronosGroup/glTF-Blender-IO

/cc @pjoe Any ideas?

pjoe:
I did a PR #673  to change the timeout to 20 min :S

Obviously not ideal solution - IIRC the blender devs mentioned that download.blender.org doesn't have CDN while builder.blender.org has (which is why this is only an issue for 2.79).

emackey:
Thanks @pjoe for fixing the download timeout.

There was another odd error, an unhandled promise rejection, which I've fixed in the test suite in commit 842e4b04b130.

The remaining test failures appear to be very serious:  The Blender 2.81 build isn't exporting any texturemaps from several of the round-trip test models.  Specifically, all of the numbered entries in this list have failed to export textures:

```
  Importer / Exporter (Roundtrip)
    blender28_roundtrip
      1) 01_alpha_blend
      2) 01_alpha_mask
      √ 01_color_attribute (980ms)
      √ 01_cs_morph (870ms)
      √ 01_cs_rotate (830ms)
      √ 01_cs_scale (800ms)
      √ 01_cs_translate (830ms)
      √ 01_cube (810ms)
      √ 01_cube_no_material (820ms)
      √ 01_metallic_sphere (960ms)
      √ 01_morphed_cube (1000ms)
      √ 01_morphed_cube_no_uv (890ms)
      √ 01_morphed_triangle (850ms)
      √ 01_plane (840ms)
      3) 01_principled_material
      √ 01_single_vs_double_sided (950ms)
      √ 01_sphere (940ms)
      4) 01_textured_sphere
      5) 01_textured_sphere_principled_bsdf
      √ 01_triangle (830ms)
      √ 01_unlit (830ms)
      √ 02_node_hierarchy (830ms)
      √ 02_shared_mesh (830ms)
      6) 02_suzanne
      √ 03_all_animations (880ms)
      √ 03_animated_cube (810ms)
      √ 03_skinned_cylinder (955ms)
      7) 04_common_materials
      8) 04_lenna
      √ 04_lights (830ms)
      √ 04_sphere_specular_glossiness (939ms)
      √ 05_metallic_sphere_light (961ms)
      √ 05_node_material (1110ms)
      √ 06_parent-inverse-anim (880ms)
      √ 07_nla-anim (840ms)
[...]

PR #673, opened 2019-09-13
Set CI timeout to 20m to avoid failure

pjoe:
Workaround for #672 slow blender 2.79 download that times out on CircleCI.

Looks like it takes ~12m to download blender 2.79.

Unfortunately also looks like there are other test failures, see sample build here: https://circleci.com/gh/p3d-in/glTF-Blender-IO/77


emackey:
Thanks @pjoe, merging this.

Issue #671, opened 2019-09-12
Import: multi-object animations


scurest:
I was trying out the mult-object animation stuff in the importer and assuming I understand it correctly, I found some issues.

### Track names are not unique per animations

Here's a [test file](https://github.com/KhronosGroup/glTF-Blender-IO/files/3603573/MultObjAnim.zip). It's just BoxAnimated from the sample model with three copies of the same animation (which animates two objects); two are named "A", and one is unnamed. It imports like this

![bl](https://user-images.githubusercontent.com/11024420/64750166-b15c7c00-d4dd-11e9-91f7-4502be984edd.png)

So the tracks from the two animations named "A" have the same name and there's no way to tell them apart. On the other hand, the tracks from the unnamed animation are named with the action name, so there's no way to tell that they're part of the same animation.

So I think there should be a preprocessing pass that picks a unique name for each animation in the glTF that can be used as the track name every time it's actions get pushed.

### Actions for different objects get pushed onto an object's NLA stack

Again, in the above image, every object has the actions for every _other_ object pushed onto it, not just it's own. 

### The wrong actions get restored

The importer restores the action from the bottom-most (first) track after importing. In this case, because of the above point, it means that the purple box (inner_box) plays the blue box's action (A_Node) instead of it's own (A_inner_box). It should look like this, but it currently doesn't

![](https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/BoxAnimated/screenshot/screenshot.gif)

Other sample animations suffer similarly (ex. VC).

Restoring the first track's action might also wind up playing actions from two different animations if, say, not all objects are animated by every animation.

So I think the right thing to do is to pick one animation (say the first one) and always restore the actions needed to play that one.

### No way to play animations?

Is there a way to actually _play_ an animation encoded as a set of identically-named tracks in Blender? Short of manually selecting them all the right actions? If not, could the addon add a widget somewhere that would do that automatically for you?

julienduroure:
Hello,

- [ ] 1: Track names are not unique per animations: Great enhancement ideas. Added on my TODO list
- [ ] 2: Actions for different objects get pushed onto an object's NLA stack : Clearly a bug
- [ ] 3: The wrong actions get restored: consequence of bug 2. Will be solved after solving 2., and can be enhanced by 4.
- [ ] 4: restore the first glTF animation: welcome enhancement that will clarify things 
[...]

Issue #518, opened 2019-06-01
Animation not imported correctly


mikejsavage:
**Describe the bug**
The attached model loses its animations when imported to blender but works in other viewers

(see also https://github.com/KhronosGroup/glTF-Blender-IO/issues/511)

[bigvic2.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3243735/bigvic2.zip)

mikejsavage:
Nice, thanks for the fast fix!

Loosely related, would it cause problems if the importer didn't rename things? Like the `name = animation.name + "_" + obj.name` line. I was hoping to refer to animations/materials by name in my engine and the renaming makes that difficult

julienduroure:
Hello,

Things can be done, but not 100%. Because in gltf, an action can apply on multiple nodes, than can be on different armatures / meshes, and blender action are different, so action named have to be different.

This will need a bit of refactoring. Priority is currently done on fixing bugs, but at long term, we can improve naming

PR #676, opened 2019-09-14
Import: fix most of #671

scurest:
~~This gets started on #671.~~ Okay, I actually wound up doing all of 1-4. Here's how it looks now:

bars

I tested with a [simple script](https://gist.github.com/scurest/ed3d80213fcf422bf4314eb4f9f514c6) and I can now "play" an animation by giving its name. It successfully round-trips with the exporter too.

------

Unique name were picked for each animation and stored in `animation.track_name`. It works just like picking names for other Blender stuff works, by appending .001, then .002, etc. until a name is free. The names are only unique within the current glTF file. There might be tracks already in the .blend that they collide with though.

Morph weights were spun out into their own file (did one object with both morph weight and TRS animations ever work? I think you need two actions...). I also redid it to use `foreach_set` like in #253 since it was actually easier, so there should be a nice performance boost.

Bones don't find their action by its name anymore, they look in a cache at `gltf.arma_cache`. This simplifies things a lot. This should get rid of the problems with grabbing old actions or animations having the same name, and I checked that #518 didn't regress.

Stashing was also simplified. Every time an action gets created, a note is pushed onto a work list at `gltf.needs_stash` and then it's all executed at the end. This got rid of a lot of repeated code too.

And restoring is now done by walking down from the root and and looking for strips with the right name and loading the action from those.

Hopefully you can tell me how it looks and if deleted anything I shouldn't have :)

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation.py
@@ -12,8 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import bpy
+
 from .gltf2_blender_animation_bone import BlenderBoneAnim
 from .gltf2_blender_animation_node import BlenderNodeAnim
+from .gltf2_blender_animation_utils import restore_animation_on_object
 
 
 class BlenderAnimation():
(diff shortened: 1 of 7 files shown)


julienduroure:
Thanks for the PR. I will review it this week. Target is to merge it soon, to get it in 2.81 (we are now in a stabilization step, no more new features are allowed. But this PR can go, because this is "Improve and stabilize" step now)

PS: Don't hesitate to assign yourself the ticket when working on a feature, it will avoid someone else start working on same subject in same time

PR #685, opened 2019-09-24
Importer: big performance bump in mesh creation

scurest:
This is a big change. Can wait until after stabilization step for review.

-----

This is a port of the mesh creation code I wrote for gltf-blender-importer, replacing most of the code in gltf2_blender_mesh and gltf2_blender_primitive. It gives a pretty big bump in importer performance for large models (check out Sponza!).

|Sample Model|Before|After|
|:---|:---|:---|
|VC | 3.40s | 3.24s|
|GearboxAssy | 10.68s | 5.82s|
|MetalRoughSpheres | 20.17s | 7.82s|
|BrainStem | 25.59s | 2.04s|
|Sponza | 180.32s (!!) | 4.38s |

A significant fraction of the import times now appear to be in the calculation for *Shading -> Use Normal Data*. If we use *Shading -> Smooth Shading*

|Sample Model|After|
|:---|:---|
|GearboxAssy | 2.78s|
|MetalRoughSpheres | 4.03s |
|BrainStem | 1.50s|
|Sponza | 2.45s |

Basically, the difference is this creates one BMesh, adds all primitive data into it, and then converts to a mesh.

Additional features that came along are:

* \> 4 vertex skinning weights are now imported (from `JOINTS_1` etc.)
* UV and vertex color layers are always created in order of ascending number (ex. `COLOR_0`, `COLOR_1`, etc.) with no gaps

Also it picks shapekey names per mesh instead of having one list for the whole glTF (did this work when there are multiple meshes with shapekeys?).

Needs lots of testing. The part I am most worried about is the code for normals, especially the `gltf.import_settings['import_shading'] == "NORMALS"` stuff.

Let me know what you think.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation_weight.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation_weight.py
@@ -70,11 +70,8 @@ class BlenderWeightAnim():
         values = BinaryData.get_data_from_accessor(gltf, animation.samplers[channel.sampler].output)
 
         # retrieve number of targets
-        nb_targets = 0
-        for prim in gltf.data.meshes[gltf.data.nodes[node_idx].mesh].primitives:
-            if prim.targets:
-                if len(prim.targets) > nb_targets:
-                    nb_targets = len(prim.targets)
+        pymesh = gltf.data.meshes[gltf.data.nodes[node_idx].mesh]
+        nb_targets = len(pymesh.shapekey_names)
 
         if animation.samplers[channel.sampler].interpolation == "CUBICSPLINE":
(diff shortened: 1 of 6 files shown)


julienduroure:
Hello,
Thanks for the PR.
As you said, it will probably be reviewed only after 2.81 stabilization step.

"Unfortunately", I also have a big refactor of this part of the code, not really compatible with this one. We will have to get the best part of the both PR.
With my PR, some very big files go from 41 minutes to 1.5s :)

julienduroure:
OK, I tested this PR today. Seems to be OK.
I will merge it now, because we are still in Bcon2.

I will then commit another importer perf improvement

PR #689, opened 2019-09-26
Scurest mesh creation dev

julienduroure:


julienduroure:
Oups, didn't push it in already existing PR ... Closing this one

scurest:
@julienduroure 
Saw this, caching accessors is very nice! What do you think about adding an optional `cache` parameter to `BinaryData.get_data_from_accessor` instead? So it'd look like eg.

````python
BinaryData.get_data_from_accessor(gltf, attributes['POSITION'], cache=True)
````

Would be shorter and easier to reuse I think (might be useful for caching the input accessor for samplers later on too, those often get used multiple times).

julienduroure:
Yes, this can be a great code refactoring enhancement. I already pushed the fix to master, (because I need it right now for my job), but fill free to propose a PR if you want to do it. (or create a ticket if you want me to remember)

scurest:
Ok, thanks :)

Issue #558, opened 2019-06-18
Skinning examples often require 'Always sample animations' to be enabled


donmccurdy:
This is a tracking bug for examples that export correctly with *Always sample animations* enabled, but not when it is disabled. I would consider all of these to be lower-priority than other animation issues, because the option is there to work around them.

- [ ] [ac_coa.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3303704/ac_coa.blend.zip) from https://github.com/KhronosGroup/glTF-Blender-IO/issues/347
- [ ] [game_man_sketchbookb1.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3303706/game_man_sketchbookb1.zip) from https://github.com/KhronosGroup/glTF-Blender-IO/issues/373
- [ ] [rotation-example.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3303754/rotation-example.zip) from https://github.com/KhronosGroup/glTF-Blender-IO/issues/351 (incomplete rotation)
- [ ] [Cow.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3334088/Cow.blend.zip) from https://github.com/KhronosGroup/glTF-Blender-IO/issues/571

samanthajo2:
I'd like to argue this should be a higher priority. For example that super low poly Cow file. 

Without animation: 121k
With curve based animation: 187k
With sampled animation: 615k.  

That's a 320% increase in size and that's a very simple skeleton with only 6 short animations.

if glTF is the transmission format for the web then it's actually extremely important the files be small. A solution with large files that drive users away from websites is not really a viable a solution. In other words I'm arguing the supposed workaround is not really a workaround. Sure you can test something but you can't ship if you want users to have a good experience.

PS: I know I'm not working the exporter and it's easy to say something when I'm not the one implementing. I'm only trying to add a POV that there really isn't a work around.
I appreciate all the amazing work you guys are doing. Just wanted to point out something I thought might not have been considered.

julienduroure:
@samanthajo2 The cow armature has some IK constraints. That means that the deformation bones are not directly animated (animation on FrontFoot.L location will have some influence on location and rotation on FrontLeg.L and FrontUpLeg.L and FrontLowLeg.L.
These 3 later bones are not directly animated, and I can't find any other solution than sampling to take into account these animations in glTF files.

Any good ideas are of course welcome :)

[...]

PR #706, opened 2019-10-01
Export with Always Sample Animations by default

Peach1:
Export more reliably correct animation playback for skinned meshes. #558
Users will vastly prefer that their exported content plays back correctly by default. If "Always Sample Animations" is off, there is often only a marginal difference in filesize, but high chance of animations that are completely invalid and rotating/moving in the wrong directions. The exporter should be robust and work correctly by default

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -247,7 +247,7 @@ class ExportGLTF2_Base:
     export_force_sampling = BoolProperty(
         name='Always Sample Animations',
         description='Apply sampling to all animations',
-        default=False
+        default=True
     )
 
     export_nla_strips = BoolProperty(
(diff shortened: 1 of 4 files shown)


julienduroure:
@emackey @donmccurdy Setting always sample on by default fails the roundtrip because some cubispline animations become linear.
How can we handle that?

I would like to commit this change before end of the week

emackey:
I added a new "options" file to the test suite, with only one option so far, to turn off sampled animations.  The one morph animation test file has the new option, and all the rest of the files are affected by the new default value to force-sample all their animations.

I did this because I didn't want to lose the test that deeply inspected a morph target animation output buffer, and I didn't want the test to look for a sampled buffer, so, this one test model will continue to not sample its animations, while the rest do.  In general the test suite should probably get upgraded with a lot more options to test a bunch of untested code paths.

This can be merged once the tests pass.

PR #709, opened 2019-10-02
Import extra data as custom properties

scurest:
This moves `generate_extras` for the exporter into the common files and adds a `set_extras` for the importer to use.

Extra data on the following objects is imported and can be round-tripped when the _Custom Properties_ option for the exporter is checked

* meshes
* materials
* nodes (except bones)
* lights
* cameras

(This was requested at https://github.com/ksons/gltf-blender-importer/issues/48)


julienduroure:
Next step: we need to manage bones too.

Issue #376, opened 2019-03-21
Support POINT and LINE primitives


Dok11:
Mesh with verts, edges and without faces doesnt save geometry in exported files

![image](https://user-images.githubusercontent.com/2697890/54766882-23c66080-4c0d-11e9-8559-083e17c0d779.png)

[gltf.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/2992925/gltf.zip)

emackey:
Currently this project will only export `TRIANGLES`, not `LINES` or `POINTS` in glTF format, so this is expected behavior.

Dok11:
"project" — you mean this addon or blend-file?

emackey:
Sorry, this addon.  It doesn't look like we have a feature request open to support lines & points.  They're not very common in the glTF ecosystem, most renderers are focused on triangles.

emackey:
But is that what you're asking for?  What are you trying to get from this, just preserve exact edges, or render some lines?

Dok11:
@emackey if you interest my case:
In Blender we pain spline what will be as path to movement camera, then we convert spline to mesh (vertexes like in my screenshot) and export to gltf. Then we inport scene into threejs and try attach camera to camera path, but as you now leaned it is impossible — file doesnt have required data.

Dok11:
Maybe you can prompt what place in export script may realizes this logic?
I think around this place https://github.com/KhronosGroup/glTF-Blender-IO/blob/master/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py#L521

Dok11:
@emackey how you think, what chance what this feature request will be approved?

emackey:
We're always eager to receive contributions from the community.  In this case though, exporting geometry without faces seems the wrong way to store the information in glTF.  Perhaps it would be better to assign a Blender camera to follow the path, and bake the animation to glTF?  I'm not a ThreeJS expert so I don't know if that would translate well to that library, but, I would imagine an animated camera node could be cleanly stored in the glTF format.

donmccurdy:
Merging some details from https://github.com/KhronosGroup/glTF-Blender-IO/issues/352 –

***

PRs for point and line support would be welcome. See:

- KhronosGroup/glTF-Blender-Exporter#50
- https://twitter.com/donrmccurdy/status/1104439928660746240

nasser:
@donmccurdy that tweet is a response to me! hello :wave: 

[...]

PR #717, opened 2019-10-07
Add preliminary support for loose edges and points

nasser:
This patch implements the enhancement discussed in #376.

* It adds a "Loose Vertices" and a "Loose Edges" checkbox to the Geometry drop down of exporter UI
* When "Loose Vertices" is checked, the exporter now extracts vertices that are not associated with any edge into a glTF `POINTS` primitive
* When "Loose Edges" is checked, the exporter now extracts edges that are not associated with any polygon into a glTF `LINES` primitive
* The exporter is patched to support various primitive modes (as opposed to being hard-coded for triangles)

This is a preliminary patch and some more work can be done in the future. Some things that come to mind:

* No splitting occurs, which might be an issue with large `LINES` primitives as they are indexed
* `POINTS` primitives are still indexed, which is likely wasteful. Addressing this would require a bigger patch to the exporter that I wanted to avoid without further guidance on the issue
* Materials, colors, and UV coordinates are not exported. I am not sure how much of this information Blender even allows you to associate with loose geometry, but it could be nice to have if possible.

I am open to feedback and guidance on these and any other issues.

I've [attached](https://github.com/KhronosGroup/glTF-Blender-IO/files/3697890/primitives.zip) a test scene and three files exported by this patch:

* `lines.glb` is exported with "Loose Edges" checked
* `points.glb` is exported with "Loose Vertices" checked
* `points-and-lines.glb` is exported with both checked
* `scene.blend` is the source scene


janEntikan:
It would be nice if this was resolved and merged.

nasser:
it was ready to go when i submitted it, but it didn't get merged at the time, not sure why. it will take some doing to resolve the conflicts as i am not actively working on blender stuff at the moment, but i will try and make time for it.

addons/io_scene_gltf2/__init__.py
```diff
@@ -384,6 +396,8 @@ def execute(self, context):
         export_settings['gltf_extras'] = self.export_extras
         export_settings['gltf_yup'] = self.export_yup
         export_settings['gltf_apply'] = self.export_apply
+        export_settings['export_loose_edges'] = self.export_loose_edges
+        export_settings['export_loose_vertices'] = self.export_loose_vertices
```
donmccurdy, line 400: Let's name these export options `use_mesh_edges` and `use_mesh_vertices`. The first will match the FBX addon, so we have a more consistent scripting API. The OBJ exporting uses `use_edges` instead, which so we can't be 100% consistent, but at least we avoid introducing a third name.

nasser: Sounds good, I will give it a crack sometime this week.

nasser:
Something I struggled with when I was writing this was how to develop one of the included addons while using stock blender and avoid getting conflicts. I think I ended up erasing the `io_scene_gltf2` folder that came with my blender installation and symbolically linked it to my clone instead... feels clunky. is there a better development workflow you use?

donmccurdy:
I'm using the steps in [DEBUGGING.md](https://github.com/KhronosGroup/glTF-Blender-IO/blob/master/DEBUGGING.md). It seems like you do need to delete the original `io_scene_gltf2` folder, but then I configure Blender's *Scripts* settings to point to the versioned glTF-Blender-IO/ directory in a more convenient location. This works well for me, but I'm also updating the addon folder much more frequently than I update Blender itself.

mkvvltra:
Hey, tried using this branch with 'Compression' option enabled - was met with following error:

```
  File "/Users/___/Library/Application Support/Blender/2.82/scripts/addons/io_scene_gltf2/__init__.py", line 434, in execute
    return gltf2_blender_export.save(context, export_settings)
  File "/Users/___/Library/Application Support/Blender/2.82/scripts/addons/io_scene_gltf2/blender/exp/gltf2_blender_export.py", line 40, in save
    json, buffer = __export(export_settings)
  File "/Users/___/Library/Application Support/Blender/2.82/scripts/addons/io_scene_gltf2/blender/exp/gltf2_blender_export.py", line 54, in __export
    __gather_gltf(exporter, export_settings)
  File "/Users/___/Library/Application Support/Blender/2.82/scripts/addons/io_scene_gltf2/blender/exp/gltf2_blender_export.py", line 72, in __gather_gltf
    gltf2_io_draco_compression_extension.compress_scene_primitives(scenes, export_settings)
  File "/Users/___/Library/Application Support/Blender/2.82/scripts/addons/io_scene_gltf2/io/exp/gltf2_io_draco_compression_extension.py", line 65, in compress_scene_primitives
    dll.createCompressor.restype = c_void_p
  File "/Applications/Blender.app/Contents/Resources/2.82/python/lib/python3.7/ctypes/__init__.py", line 377, in __getattr__
    func = self.__getitem__(name)
  File "/Applications/Blender.app/Contents/Resources/2.82/python/lib/python3.7/ctypes/__init__.py", line 382, in __getitem__
    func = self._FuncPtr((name_or_ordinal, self))
AttributeError: dlsym(0x6000002c1e00, createCompressor): symbol not found

location: <unknown location>:-1
```

It seems to occur even if 'Loose Vertices' and 'Loose Edges' are disabled.

donmccurdy:
Note that the compression methods implemented in this addon do only support triangle meshes. But we should make sure that it safely ignores loose vertices and edges whether compression is enabled or not, yes.

donmccurdy:
We'd be glad to support this feature, but for clarity (since it currently needs updates) I'll mark the PR as a draft. Please feel free to mark it as ready for review, or open another PR, whenever you like. :)

nasser:
Will do. Sorry, life situation changed since I made the original PR and I'm trying to make time to update it!

donmccurdy:
Completely understand, no worries! When you have the time let us know, or any other users interested in this feature are welcome to build on your work and open a new PR too. Thanks for getting this started. 🙂

Snouf-ID:
Exporting points and lines would be very useful. I had to do an obj-wavefront to gltf converter in my code, which is annoying if you want to have anything other than simple primitives with no great material.
This great work is highly anticipated !

dohkoxiaozu:
I'm looking forward to exporting lines, because if I can't export lines, I'll have to manually find the right place in the world to calculate the right dot matrix to draw a line (which is a disaster).

scurest:
[Here](https://github.com/scurest/glTF-Blender-IO/tree/loose-geometry) is an up-to-date branch with loose geometry support in case it helps anyone. Look for the "Loose Geometry" options under "Geometry" on the export screen. I have not tested it extensively.

nasser:
Updated the patch to the changes that have happened in the meantime. I tracked my original implementation fairly closely, so I expect its not as idiomatic as it should be, specifically w.r.t #1023 which I haven't looked at. It works locally, and the CI tests pass, so that's something. Still a few things to do though:

* Look at #1023 and make sure my implementation is in line with it
* Make sure the patch is safely ignored by compression as brought up by @mkvvltra 

Once I get those taken care of I will start a review. Happy for any guidance/feedback as always.

@scurest I ended up updating my original code as that was the path of least resistance -- thanks though!

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
```diff
@@ -283,10 +284,90 @@ def extract_primitives(glTF, blender_mesh, library, blender_object, blender_vert
...
+            'indices': np.empty(vert_count, dtype=np.uint32),
+            'material': 0,
+            'attributes': attributes
+        }
+        
+        i = 0
+        for edge in valid_edges:
+            for j in edge.vertices:
+                vertex = blender_mesh.vertices[j]
+                attributes['POSITION'][i, :] = swizzle_yup_location(vertex.co) if export_settings[gltf2_blender_export_keys.YUP] else vertex.co
```
scurest, line 332: Please slice the correct locations out of the `locs` array, which holds the glTF vertex position for each blender vert idx. This includes transforms for skinning in addition to the Yup thing.

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
```diff
@@ -283,10 +284,90 @@ def extract_primitives(glTF, blender_mesh, library, blender_object, blender_vert
...
+                attributes['NORMAL'][i, :] = vertex.normal
```
scurest, line 333: The normal is still in Blender space; it would have to be swizzled and skinned. But not sure normals make sense for loose geometry...

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
```diff
@@ -283,10 +284,90 @@ def extract_primitives(glTF, blender_mesh, library, blender_object, blender_vert
...
+    
+    else:
+        vert_count = 0
+        for e in valid_edges:
+            vert_count += len(e.vertices)
+        
+        attributes = {
+            'POSITION': np.empty((vert_count, 3), dtype=np.float32),
+            'NORMAL': np.empty((vert_count, 3), dtype=np.float32)
+        }
```
scurest, line 319: You need to do morph targets (`MORPH_POSITION_%d`) too. Spec requires that all primitives have the same number of targets, so if you don't do it the file will be invalid.

You need WEIGHTS and JOINTS too so skinning will work.

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
```diff
@@ -283,10 +284,90 @@ def extract_primitives(glTF, blender_mesh, library, blender_object, blender_vert
...
+    if len(valid_edges) == 0:
+        return []
+    
+    else:
+        vert_count = 0
+        for e in valid_edges:
+            vert_count += len(e.vertices)
+        
+        attributes = {
+            'POSITION': np.empty((vert_count, 3), dtype=np.float32),
```
scurest, line 317: You're generating one glTF vert for each end of a Blender edge. You should generate one per Blender vert. Eg for this

star

You're generating 12 verts; you should only generate 7.

scurest (changes requested):
Think you should consider using my branch as a base instead. In addition to being in line with #1023, it fixes several bugs, uses the FBX-compatible option names discussed above, doesn't export indices for POINT primitives, and adds a test.

Pointing out some issues anyway. I'll only comment on edges, but some of this applies to points too.

nasser:
@scurest thank you for the detailed review. I think you're right, I will take a look at your branch as the base instead next!

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
```diff
@@ -283,10 +284,90 @@ def extract_primitives(glTF, blender_mesh, library, blender_object, blender_vert
...
+        primitives.extend(point_primitives)
+
     print_console('INFO', 'Primitives created: %d' % len(primitives))
 
     return primitives
 
+def extract_line_primitives(blender_mesh, excluded_indices, export_settings):
+    valid_edges = []
+    for blender_edge in blender_mesh.edges:
+        if blender_edge.vertices[0] not in excluded_indices or blender_edge.vertices[1] not in excluded_indices:
```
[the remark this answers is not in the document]
nasser: I didn't know about that property, super useful!

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
```diff
@@ -283,10 +284,90 @@ def extract_primitives(glTF, blender_mesh, library, blender_object, blender_vert
...
+            # edge has at least one edge not in a polygon, add to primitive
+            valid_edges.append(blender_edge)
+    
+    if len(valid_edges) == 0:
+        return []
+    
+    else:
+        vert_count = 0
+        for e in valid_edges:
+            vert_count += len(e.vertices)
```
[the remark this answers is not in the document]
nasser: I had this originally but I somehow convinced myself that edges could contain more than one vertex, I think I was thinking about loops

scurest:
About Draco: I think it's only for mode=4? So this?

```diff
diff --git a/addons/io_scene_gltf2/io/exp/gltf2_io_draco_compression_extension.py b/addons/io_scene_gltf2/io/exp/gltf2_io_draco_compression_extension.py
index 2bb79aa0..6a661334 100644
--- a/addons/io_scene_gltf2/io/exp/gltf2_io_draco_compression_extension.py
+++ b/addons/io_scene_gltf2/io/exp/gltf2_io_draco_compression_extension.py
@@ -200,6 +200,10 @@ def __compress_primitive(primitive, dll, export_settings):
         'UnsignedInt': 4,
     }
 
+    # Only compress triangle mode
+    if primitive.mode not in [None, 4]:
+        return
+
     # Positions are the only attribute type required to be present.
     if 'POSITION' not in attributes:
         print_console('WARNING', 'Draco exporter: Primitive without positions encountered. Skipping.')
```

Maybe @UX3D-eckerlein could confirm? (This PR will add primitives with mode=0 and mode=1.)

donmccurdy:
That's correct, the `KHR_draco_mesh_compression` extension only supports triangles.

nasser:
backed up and used @scurest's changes. they work great and meet all my own use cases. all i did was tweak a few internal naming conventions. this is more or less ready to go but a few things come to mind. let me know if these are out of scope/make more sense in a separate PR.

1) this is a UI quibble, but currently the "Loose Geometry" label is clipped by default.

screenshot-00039

alternatives that come to mind are

```
Loose [ ] Edges
      [ ] Points
```

or 

```
[ ] Loose Edges
[ ] Loose Points
```

any preferences? other ideas? not an issue?

2) currently all loose geometry is exported without a material meaning, in threejs  at least, they get a solid white material. it would be great if there was some way to affect this from the blender side, though if i understand correctly materials can only be assigned to faces, not edges or points. i am not sure what the best way to "pick" a material for these primitives is. simply using the material at index zero of the associated mesh covers a lot of the use cases i have in mind, though.

screenshot-00040

it means loose geometry gets a reasonable default, and if it's not what the user wanted they can break the loose geometry out into its own mesh and assign a different material. admittedly, the only thing that gets communicated is color, but it makes a difference in some cases. the image above is from a mock-up i have not pushed out, rendered in three.js (intentionally downsampled to highlight the effect).

does this make sense? maybe behind another export option? is there a better way to deal with materials on loose geometry?

addons/io_scene_gltf2/__init__.py
```diff
@@ -431,10 +443,18 @@ def invoke(self, context, event):
         return ExportHelper.invoke(self, context, event)
 
     def save_settings(self, context):
-        # find all export_ props
+        # find all props to save
+        exceptional = [
+            # options that don't start with 'export_'
+            'use_selection',
+            'export_loose_edges',
+            'export_loose_points',
```
scurest, line 451: I thought these were going to be consistent with the FBX exporter, like [the comment above](https://github.com/KhronosGroup/glTF-Blender-IO/pull/717#discussion_r430054075) said.

Also like the comment says, this is only for option names that don't start with `export_` so no need to put them here if they're named like this.

scurest:
1. Up to the maintainers but I'd do the second one

    ```
    [ ] Loose Edges
    [ ] Loose Points
    ```

    because of consistency with the FBX exporter.

    fbx

2. Using the 0th material slot seems like a good idea, +1 from me. It matches how Blender uses the 0th slot when it needs a default (eg for the faces of a Surface, or a newly created Mesh). And like you said, otherwise there's no way to set a material at all.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitives.py
```diff
@@ -54,7 +54,7 @@ def gather_primitives(
         material_idx = internal_primitive['material']
         material = None
 
-        if export_settings['gltf_materials'] == "EXPORT":
+        if export_settings['gltf_materials'] == "EXPORT" and  material_idx is not None:
```
scurest, line 57: There are two spaces after `and`.

donmccurdy:
^Agreed on @scurest's suggested solutions to (1) and (2) above.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitives.py
expenses, line 57: 😉 
```suggestion
        if export_settings['gltf_materials'] == "EXPORT" and material_idx is not None:
```

expenses:
Looking forwards to this! ❤️

scurest:
I've updated my [branch](https://github.com/scurest/glTF-Blender-IO/tree/loose-geometry) with everything discussed so far.

nasser:
@scurest this looks great to me, thanks for taking care of this!

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
```diff
@@ -283,6 +283,90 @@ def extract_primitives(glTF, blender_mesh, library, blender_object, blender_vert
...
+                    attributes['JOINTS_%d' % i] = js
+                    attributes['WEIGHTS_%d' % i] = ws
+
+            primitives.append({
+                'attributes': attributes,
+                'indices': indices,
+                'mode': 1,  # LINES
+                'material': 0,
+            })
+
```
donmccurdy, line 327: I don't quite understand how `material: 0` works, but it seems to work anyway ... is the "0" index here getting resolved to the (possibly non-zero) eventual index of the first material in the current mesh?

addons/io_scene_gltf2/io/exp/gltf2_io_draco_compression_extension.py
```diff
@@ -85,6 +85,10 @@ def __encode_primitive(primitive, dll, export_settings):
     attributes = primitive.attributes
     indices = primitive.indices
 
+    # Only do TRIANGLES primitives
+    if primitive.mode not in [None, 4]:
+        return
```
donmccurdy, line 90: Despite this check, I'm still getting an error when I enable Draco export. Here's a test case:

[Uploading LooseTest.blend.zip…]()

donmccurdy (commented):
This is an awesome PR, thanks for the updates @nasser and @scurest.

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
scurest: Yes, the `material` field on an "internal primitive" is the material slot index. It's resolved into the actual `gltf2_io.Material` (or `None`) [here](https://github.com/KhronosGroup/glTF-Blender-IO/blob/4f4f6fb09cc588d938994b97c780fb20d8e42f1b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitives.py#L54-L68).

addons/io_scene_gltf2/io/exp/gltf2_io_draco_compression_extension.py
scurest: I get that anytime I try to export with "Compresssion" on, even with master.

donmccurdy: Hm, I'm definitely willing to believe it's a version mismatch or some detail of the native Draco integration with local Blender, if it's working for someone else. I typically don't do compression until after exporting, with an offline tool.

That said ... the sample above has no triangles at all, so ideally it wouldn't be trying to create a Draco encoder.

This doesn't seem like a blocking issue, then, in any case.

donmccurdy (approved):
Excited to have this functionality, and glad the implementation ended up being pretty concise. 

@emackey any concerns with the material handling?

Has someone tested this with skinned points or lines? I don't think three.js even supports that, I'm not sure about Babylon.js or other engines...

If there are no concerns or other issues, I'll merge this in a few days.

scurest:
> Has someone tested this with skinned points or lines?

Yes, it works reimporting back into Blender and in Babylon, but not in Three.

addons/io_scene_gltf2/__init__.py
```diff
@@ -242,6 +242,24 @@ def __init__(self):
         default=True
     )
 
+    use_mesh_edges: BoolProperty(
+        name='Loose Edges',
+        description=(
+            'Export loose edges too. Loose geometry will use the material from the '
+            'first material slot'
```
emackey, line 249:
```suggestion
            'Export loose edges as lines, using the material from the first material slot'
```

addons/io_scene_gltf2/__init__.py
```diff
@@ -242,6 +242,24 @@ def __init__(self):
...
+        ),
+        default=False,
+    )
+
+    use_mesh_vertices: BoolProperty(
+        name='Loose Points',
+        description=(
+            'Export loose points too. Loose geometry will use the material from the '
+            'first material slot'
```
emackey, line 258:
```suggestion
            'Export loose points as glTF points, using the material from the first material slot'
```

emackey (commented):
Some minor wording suggestions for the export options.  Are these OK?

emackey:
> @emackey any concerns with the material handling?

We actually had a concern about how materials are [barely] specified for points & lines crop up in the Khronos PBR TSG just this week.  But that didn't have an easy answer, so it shouldn't hold up this PR.

I have no concerns about the mechanics of using the first material slot, so this should be good to go.

scurest:
So what happened?

donmccurdy:
@scurest I think @emackey is saying that the material handling here is fine and we should proceed.

We are down to just unresolved comments on tooltips at this point.

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
donmccurdy: Resolving — normal attributes have been removed from the PR. Willing to discuss adding them again but let's consider that in a new issue if someone has a need for it.

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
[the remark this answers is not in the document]
donmccurdy: I can't find this code in the PR anymore, and am assuming this was resolved.

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
donmccurdy: Nice to see this can be supported too. 👍🏻

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
donmccurdy: This code isn't present in the PR anymore; resolving.

donmccurdy:
Thanks @nasser and @scurest for implementing and testing this, a really helpful addition. 🎊

nasser:
it was all @scurest in the end :pray: very happy to see this merged and i look forward to using this feature in blender!

mogya2 (approved):

PR #726, opened 2019-10-11
Export: don't split primitives

### LINKED ISSUE #233: gltf exporter: unwanted split of a large mesh
In Blender 2.8 Beta, I've imported Stanford dragon and exported it in ".gltf + bin" format. In the generated gltf file, geometry is split in multiple "primitive objects" such that indices are **unsigned short**, instead of keeping only one "primitive object" using **unsigned int** indices.

scurest:
Removes primitive splitting in the exporter.

Should fix #233.


emackey:
Added new tests for UNSIGNED_SHORT and UNSIGNED_INT.  Behavior appears to be correct.

emackey:
Thanks @scurest!

Issue #730, opened 2019-10-12
Export Custom Properties for EnumProperty Type


Ziflin:
It appears that currently "EnumProperty()" types are not exported as Custom Properties? Would this be difficult to add to avoid needing to wrap an EnumProperty with a StringProperty every time?

Attached is an example Blender file with a simple script that adds an "Exported Properties" panel to object properties (just above the Custom Properties panel). This script also adds a "Collider Type" EnumProperty that is not getting exported although the other custom BooleanProperty  for "Collidable" is being exported into "extras."

[EnumProperties.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3720891/EnumProperties.zip)

scurest:
Although it looks like it does from the UI, the object doesn't have a `colliderShape` custom property just from running the script. You have to click the drop-down and select something for it to be set.

You can test this by typing `D.objects[0]['colliderShape']` into the Python console. It doesn't exist until you select something with the drop-down.

julienduroure:
Yes, confirm that blender maybe come from Blender or from your script, not from exporter, because data can't be retrieve from python until you change it in UI.

Ziflin:
Sorry! I swear I tested changing it thinking that was the problem. There's not a setting to have it export as the enum's string id vs the numeric value I assume? I can just use a getter/setter I think if we really need it to export a string value.

scurest:
So currently the exporter uses the dict-like properties of an object to use for exporting, ie. what you get with `obj['prop']` and what shows up in the "Custom Properties" widget.

You're using custom `bpy.props` properties ie. what you get with `obj.prop`. These appear to store their data in dict-like properties with the same name, which is the only reason they even work at all. When a `bpy.props` property is added, the dict-like property with the same name no longer shows up in the "Custom Properties" widget, and it will re-appear if the `bpy.props` is removed. Also, the dict-like property doesn't get set until you change the custom property (ex. with the drop-down or by using `obj.prop = val`), which is why `colliderShape` wasn't being exported here.

I don't know if we want to do this, but [here's](https://github.com/KhronosGroup/glTF-Blender-IO/compare/master...scurest:export_bpy_props_extras) a patch that will try to handle these kind of `bpy.props` custom properties when exporting. It will pick up the default value of `colliderShape` even if it hasn't been set, and it uses the value of `obj.colliderShape` which is a string, instead of `obj['colliderShape']`, which is a number, so you'd get:

````json
[...]

Issue #542, opened 2019-06-12
Option to export vertex colors as non-color data


jitspoe:
I'm currently using vertex colors as data, so I want to preserve the raw values as I enter them in blender.  For example, I might use one color channel to determine how much sway foliage has for a vertex shader.  However, if I enter 0.5, for 50% sway, after the sRGB->linear conversion happens, I end up with a value of ~0.22, meaning I have to compensate for this when I color the verts, or I have to do a linear->srgb conversion in the shader, and I'd prefer not to add more calculations.  Also, there's data loss if I convert the values when I try to set them in Blender, since it appears that the colors are stored as 8bits per channel.

Note: Original issue filed was incorrect (I had colorspaces backward in my head), and this became repurposed, so I edited this.  Leaving the original post below:

The glTF exporter runs vertex colors through a function called color_srgb_to_scene_linear().  There are a couple issues with this:
1) I'm using vertex colors as data, not colors, so I want to preserve the raw values.
2) color_srgb_to_scene_linear() seems to be backward:
```
def color_srgb_to_scene_linear(c):
    if c < 0.04045:
        return 0.0 if c < 0.0 else c * (1.0 / 12.92)
    else:
        return pow((c + 0.055) * (1.0 / 1.055), 2.4)
```
srgb->linear conversion should be using 1/2.4.  Basically seems like these function names are reversed.

**To Reproduce**
1. Export using the glTF format.
2. Note the vertex color values are LOWER than the values in blender

**Expected behavior**
If values stored in the glTF format are supposed to be linear, and Blender is using sRGB, the sRGB -> linear conversion should make the values HIGHER/BRIGHTER.  If Blender is already using linear values (not sure what the Blender vertex colorspace is), no conversion should take place. Would be nice to have a checkbox for export saying "I'm already using Linear vertex color values" to avoid conversion.

**Version**
 - Blender Version [e.g. 2.80 (2019-05-17)]

**Additional context**
Looks like this is where the color conversion was introduced: https://github.com/KhronosGroup/glTF-Blender-IO/issues/261

donmccurdy:
Let's start with the question about whether `color_srgb_to_scene_linear` is correct. Vertex colors in glTF are supposed to be linear, and I'm fairly confident Blender's Python API is returning sRGB values (even though it's _hopefully_ using linear values internally). I haven't found that documented anywhere, but testing seems to confirm below.

[...]

Issue #333, opened 2019-02-27
Copy startup improvements


emackey:
Possibly related to #330 

We should copy the following startup changes from Blender Diffusion to this repo:

https://developer.blender.org/rBAc94604993b3e0bfbc733861e890aff18513e02b4

julienduroure:
I saw that this morning, and I will merge __init__.py files tonight

Issue #7, opened 2018-08-07
Exporter: KHR extensions, add hooks for Vendor extensions


UX3D-nopper:
- [x] KHR_lights_punctual
- [x] KHR_materials_unlit
- [x] KHR_texture_transform
- [x] Refactor to have hooks/callbacks

donmccurdy:
Which new `KHR_` extensions do we think should be included here? +1 for KHR_texture_transform and KHR_lights_punctual from me. Others (draco, blending, technique_webgl) might be nice to have (if finished in time) but would require more discussion first.

UX3D-nopper:
KHR_texture_transform and KHR_lights_punctual will be in. Also, we want to refactor the code, that it is easier to "hook" into the code, that future extensions can be easier implemented. It has to be avoided, that one has to change stabel code.

emackey:
Will vendors be able to add their own Blender plugins, that somehow communicate data to export into the glTF's extensions objects?  Or would every supported glTF extension need to be included here in this repo?

UX3D-nopper:
It has to be added to this repo, but they should be in their own packag and/or file.

UX3D-becher:
Extensions can easily be added to the exporter by implementing the __gather_extensions function in the appropriate module. 
For example the KHR_unlit extensions for materials is implemented in __gather_extensions of the gather_material module as follows
```python
def __gather_extensions(blender_material, export_settings):
    extensions = {}
    if blender_material.use_shadeless:
        extensions["KHR_materials_unlit"] = {}
    return extensions if extensions else None
```
Return ``None`` instead of an empty dict to avoid empty dicts in the exported glTF file.

PR #732, opened 2019-10-14
User defined extensions

jjcasmar:
This patch allows to create user defined extensions. This works by loading dynamically python files in the extensions directory. This way, a user can define his own extensions without the need to fork the exporter

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -67,6 +67,7 @@ from bpy_extras.io_utils import ImportHelper, ExportHelper
 #  Functions / Classes.
 #
 
+extension_panel_unregister_functors = []
 
 class ExportGLTF2_Base:
     # TODO: refactor to avoid boilerplate
@@ -334,6 +335,7 @@ class ExportGLTF2_Base:
     def invoke(self, context, event):
         settings = context.scene.get(self.scene_key)
         self.will_save_settings = False
+        self.has_active_extenions = False
         if settings:
             try:
                 for (k, v) in settings.items():
@@ -344,6 +346,18 @@ class ExportGLTF2_Base:
                 self.report({"ERROR"}, "Loading export settings failed. Removed corrupted settings")
                 del context.scene[self.scene_key]
 
+        if bpy.app.version < (2,80,0):
+            pass
+        else:
+            import sys
+            for addon_name in bpy.context.preferences.addons.keys():
+                try:
+                    if hasattr(sys.modules[addon_name], 'glTF2ExportUserExtension'):
+                        extension_panel_unregister_functors.append(sys.modules[addon_name].register_panel())
+                        self.has_active_extenions = True
+                except Exception:
+                    pass
+        
         return ExportHelper.invoke(self, context, event)
 
     def save_settings(self, context):
@@ -437,6 +451,18 @@ class ExportGLTF2_Base:
         export_settings['gltf_binary'] = bytearray()
         export_settings['gltf_binaryfilename'] = os.path.splitext(os.path.basename(
             bpy.path.ensure_ext(self.filepath,self.filename_ext)))[0] + '.bin'
+
+        user_extensions = []
+
+        if bpy.app.version < (2,80,0):
+            pass
+        else:
+            import sys
+            for addon_name in bpy.context.preferences.addons.keys():
+                if hasattr(sys.modules[addon_name], 'glTF2ExportUserExtension'):
+                    extension_ctor = sys.modules[addon_name].glTF2ExportUserExtension
+                    user_extensions.append(extension_ctor())
+        export_settings['gltf_user_extensions'] = user_extensions
 
         return gltf2_blender_export.save(context, export_settings)
     if bpy.app.version < (2, 80, 0):
@@ -799,6 +825,25 @@ class GLTF_PT_export_animation_skinning(bpy.types.Panel):
         layout.active = operator.export_skins
         layout.prop(operator, 'export_all_influences')
 
+class GLTF_PT_export_user_extensions(bpy.types.Panel):
+    bl_space_type = 'FILE_BROWSER'
+    bl_region_type = 'TOOL_PROPS'
+    bl_label = "Extensions"
+    bl_parent_id = "FILE_PT_operator"
+    bl_options = {'DEFAULT_CLOSED'}
+
+    @classmethod
+    def poll(cls, context):
+        sfile = context.space_data
+        operator = sfile.active_operator
+
+        return operator.bl_idname == "EXPORT_SCENE_OT_gltf" and operator.has_active_extenions
+
+    def draw(self, context):
+        layout = self.layout
+        layout.use_property_split = True
+        layout.use_property_decorate = False  # No animation.
+
 
 class ExportGLTF2(bpy.types.Operator, ExportGLTF2_Base, ExportHelper):
     """Export scene as glTF 2.0 file"""
@@ -930,6 +975,7 @@ else:
         GLTF_PT_export_animation_export,
         GLTF_PT_export_animation_shapekeys,
         GLTF_PT_export_animation_skinning,
+        GLTF_PT_export_user_extensions,
         ImportGLTF2
     )
 
@@ -951,6 +997,10 @@ def register():
 def unregister():
     for c in classes:
         bpy.utils.unregister_class(c)
+    for f in extension_panel_unregister_functors:
+        f()
+    extension_panel_unregister_functors.clear()
+        
     # bpy.utils.unregister_module(__name__)
 
     # remove from the export / import menu
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_channel_target.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_channel_target.py
@@ -20,6 +20,7 @@ from io_scene_gltf2.blender.exp.gltf2_blender_gather_cache import cached
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_nodes
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_joints
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_skins
+from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 
 @cached
 def gather_animation_channel_target(channels: typing.Tuple[bpy.types.FCurve],
@@ -29,12 +30,22 @@ def gather_animation_channel_target(channels: typing.Tuple[bpy.types.FCurve],
                                     export_settings
                                     ) -> gltf2_io.AnimationChannelTarget:
 
-        return gltf2_io.AnimationChannelTarget(
+        animation_channel_target = gltf2_io.AnimationChannelTarget(
             extensions=__gather_extensions(channels, blender_object, export_settings, bake_bone),
             extras=__gather_extras(channels, blender_object, export_settings, bake_bone),
             node=__gather_node(channels, blender_object, export_settings, bake_bone),
             path=__gather_path(channels, blender_object, export_settings, bake_bone, bake_channel)
         )
+
+        export_user_extensions('gather_animation_channel_target_hook', 
+                               export_settings, 
+                               animation_channel_target, 
+                               channels, 
+                               blender_object, 
+                               bake_bone, 
+                               bake_channel)
+        
+        return animation_channel_target
 
 def __gather_extensions(channels: typing.Tuple[bpy.types.FCurve],
                         blender_object: bpy.types.Object,
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_material_normal_texture_info_class.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_material_normal_texture_info_class.py
@@ -21,6 +21,7 @@ from io_scene_gltf2.blender.exp import gltf2_blender_search_node_tree
 from io_scene_gltf2.blender.exp import gltf2_blender_export_keys
 from io_scene_gltf2.blender.exp import gltf2_blender_get
 from io_scene_gltf2.io.com.gltf2_io_extensions import Extension
+from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 
 
 @cached
@@ -40,6 +41,11 @@ def gather_material_normal_texture_info_class(blender_shader_sockets_or_texture_
 
     if texture_info.index is None:
         return None
+
+    export_user_extensions('gather_material_normal_texture_info_class_hook', 
+                           export_settings, 
+                           texture_info, 
+                           blender_shader_sockets_or_texture_slots)
 
     return texture_info
 
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
@@ -25,6 +25,7 @@ from io_scene_gltf2.blender.exp import gltf2_blender_search_node_tree
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_materials_pbr_metallic_roughness
 from ..com.gltf2_blender_extras import generate_extras
 from io_scene_gltf2.blender.exp import gltf2_blender_get
+from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 
 
 @cached
@@ -54,6 +55,8 @@ def gather_material(blender_material, mesh_double_sided, export_settings):
         occlusion_texture=__gather_occlusion_texture(blender_material, orm_texture, export_settings),
         pbr_metallic_roughness=__gather_pbr_metallic_roughness(blender_material, orm_texture, export_settings)
     )
+
+    export_user_extensions('gather_material_hook', export_settings, material, blender_material)
 
     return material
     # material = blender_primitive['material']
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_skins.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_skins.py
@@ -21,6 +21,7 @@ from io_scene_gltf2.io.com import gltf2_io_constants
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_accessors
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_joints
 from io_scene_gltf2.blender.com import gltf2_blender_math
+from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 
 
 @cached
@@ -35,7 +36,7 @@ def gather_skin(blender_object, export_settings):
     if not __filter_skin(blender_object, export_settings):
         return None
 
-    return gltf2_io.Skin(
+    skin = gltf2_io.Skin(
         extensions=__gather_extensions(blender_object, export_settings),
         extras=__gather_extras(blender_object, export_settings),
         inverse_bind_matrices=__gather_inverse_bind_matrices(blender_object, export_settings),
@@ -43,6 +44,10 @@ def gather_skin(blender_object, export_settings):
         name=__gather_name(blender_object, export_settings),
         skeleton=__gather_skeleton(blender_object, export_settings)
     )
+
+    export_user_extensions('gather_skin_hook', export_settings, skin, blender_object)
+
+    return skin
 
 
 def __filter_skin(blender_object, export_settings):
--- a/addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
+++ b/addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
@@ -1,0 +1,22 @@
+# Copyright 2019 The glTF-Blender-IO authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+def export_user_extensions(hook_name, export_settings, gltf2_object, *args):
+    if gltf2_object.extensions is None:
+        gltf2_object.extensions = {}
+
+    for extension in export_settings['gltf_user_extensions']:
+        hook = getattr(extension, hook_name, None)
+        if hook is not None:
+            hook(gltf2_object, *args, export_settings)
--- a/misc/ExampleExtension.py
+++ b/misc/ExampleExtension.py
@@ -1,0 +1,106 @@
+import bpy
+
+bl_info = {
+    "name": "KDAB_example_extension",
+    "category": "Generic",
+    "blender": (2,80,0),
+    'isDraft': False,
+    'developer': "KDAB",
+    'url': 'https://www.kdab.com',
+}
+
+class ExampleExtensionProperties(bpy.types.PropertyGroup):
+    enabled: bpy.props.BoolProperty(
+        name='ExampleExtension_enabled',
+        description='This is an example of a BoolProperty used by a UserExtension.',
+        default=True
+        )
+    float_property: bpy.props.FloatProperty(
+        name='ExampleExtension_enabled',
+        description='This is an example of a FloatProperty used by a UserExtension.',
+        default=1.0
+        )
+
+def register():
+    bpy.utils.register_class(ExampleExtensionProperties)
+    bpy.types.Scene.ExampleExtensionProperties = bpy.props.PointerProperty(type=ExampleExtensionProperties)
+
+def register_panel():
+    # Register the panel on demand, we need to be sure to only register it once
+    # This is necessary because the panel is a child of the extensions panel, 
+    # which may not be registered when we try to register this extension
+    try:
+        bpy.utils.register_class(GLTF_PT_UserExtensionPanel)
+    except Exception:
+        pass
+
+    # If the glTF exporter is disabled, we need to unregister the extension panel
+    # Just return a function to the exporter so it can unregister the panel
+    return unregister_panel
+
+
+def unregister_panel():
+    # Since panel is registered on demand, it is possible it is not registered
+    try:
+        bpy.utils.unregister_class(GLTF_PT_UserExtensionPanel)
+    except Exception:
+        pass
+        
+
+def unregister():
+    unregister_panel()
+    bpy.utils.unregister_class(ExampleExtensionProperties)
+    del bpy.types.Scene.ExampleExtensionProperties
+
+class GLTF_PT_UserExtensionPanel(bpy.types.Panel):
+
+    bl_space_type = 'FILE_BROWSER'
+    bl_region_type = 'TOOL_PROPS'
+    bl_label = "Example Extension"
+    bl_parent_id = "GLTF_PT_export_user_extensions"
+    bl_options = {'DEFAULT_CLOSED'}
+
+    @classmethod
+    def poll(cls, context):
+        sfile = context.space_data
+        operator = sfile.active_operator
+        return operator.bl_idname == "EXPORT_SCENE_OT_gltf"
+
+    def draw_header(self, context):
+        props = bpy.context.scene.ExampleExtensionProperties
+        self.layout.prop(props, 'enabled')
+
+    def draw(self, context):
+        layout = self.layout
+        layout.use_property_split = True
+        layout.use_property_decorate = False  # No animation.
+
+
+        box = layout.box()
+        box.label(text="Is draft: " + str(bl_info['isDraft']))
+        box.label(text="Developer: " + str(bl_info['developer']))
+        box.label(text="url: " + str(bl_info['url']))
+
+        props = bpy.context.scene.ExampleExtensionProperties
+        layout.prop(props, 'float_property', text="Some float value")
+
+
+class glTF2ExportUserExtension:
+
+    def __init__(self):
+        # We need to wait until we create the gltf2UserExtension to import the gltf2 modules
+        # Otherwise, it may fail because the gltf2 may not be loaded yet
+        from io_scene_gltf2.io.com.gltf2_io_extensions import Extension
+        self.Extension = Extension 
+        self.properties = bpy.context.scene.ExampleExtensionProperties
+
+    def gather_node_hook(self, gltf2_object, blender_object, export_settings):
+        if self.properties.enabled:
+            if gltf2_object.extensions is None:
+                gltf2_object.extensions = {}
+            gltf2_object.extensions[bl_info['name']] = self.Extension(
+                name= bl_info['name'], 
+                extension={"float": self.properties.float_property},
+                required=False
+            )
+
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_channels.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_channels.py
@@ -23,6 +23,7 @@ from io_scene_gltf2.blender.exp import gltf2_blender_gather_animation_samplers
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_animation_channel_target
 from io_scene_gltf2.blender.exp import gltf2_blender_get
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_skins
+from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 
 
 @cached
@@ -149,12 +150,25 @@ def __gather_animation_channel(channels: typing.Tuple[bpy.types.FCurve],
     if not __filter_animation_channel(channels, blender_object, export_settings):
         return None
 
-    return gltf2_io.AnimationChannel(
+    animation_channel = gltf2_io.AnimationChannel(
         extensions=__gather_extensions(channels, blender_object, export_settings, bake_bone),
         extras=__gather_extras(channels, blender_object, export_settings, bake_bone),
         sampler=__gather_sampler(channels, blender_object, export_settings, bake_bone, bake_channel, bake_range_start, bake_range_end, action_name),
         target=__gather_target(channels, blender_object, export_settings, bake_bone, bake_channel)
     )
+
+    export_user_extensions('gather_animation_channel_hook', 
+                           export_settings, 
+                           animation_channel, 
+                           channels, 
+                           blender_object, 
+                           bake_bone, 
+                           bake_channel, 
+                           bake_range_start, 
+                           bake_range_end, 
+                           action_name)
+
+    return animation_channel
 
 
 def __filter_animation_channel(channels: typing.Tuple[bpy.types.FCurve],
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_samplers.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_samplers.py
@@ -27,6 +27,7 @@ from io_scene_gltf2.io.com import gltf2_io
 from io_scene_gltf2.io.com import gltf2_io_constants
 from io_scene_gltf2.io.exp import gltf2_io_binary_data
 from . import gltf2_blender_export_keys
+from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 
 
 @cached
@@ -55,7 +56,7 @@ def gather_animation_sampler(channels: typing.Tuple[bpy.types.FCurve],
                                                  export_settings)
 
 
-    return gltf2_io.AnimationSampler(
+    sampler = gltf2_io.AnimationSampler(
         extensions=__gather_extensions(channels, blender_object_if_armature, export_settings, bake_bone, bake_channel),
         extras=__gather_extras(channels, blender_object_if_armature, export_settings, bake_bone, bake_channel),
         input=__gather_input(channels, blender_object_if_armature, non_keyed_values,
@@ -71,6 +72,19 @@ def gather_animation_sampler(channels: typing.Tuple[bpy.types.FCurve],
                                action_name,
                                export_settings)
     )
+
+    export_user_extensions('gather_animation_sampler_hook', 
+                            export_settings, 
+                            sampler, 
+                            channels, 
+                            blender_object, 
+                            bake_bone, 
+                            bake_channel, 
+                            bake_range_start, 
+                            bake_range_end, 
+                            action_name)
+
+    return sampler
 
 def __gather_non_keyed_values(channels: typing.Tuple[bpy.types.FCurve],
                               blender_object: bpy.types.Object,
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
@@ -19,6 +19,7 @@ from io_scene_gltf2.io.com import gltf2_io
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_animation_channels
 from io_scene_gltf2.io.com.gltf2_io_debug import print_console
 from ..com.gltf2_blender_extras import generate_extras
+from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 
 
 def gather_animations(blender_object: bpy.types.Object,
@@ -107,6 +108,8 @@ def __gather_animation(blender_action: bpy.types.Action,
 
     if not animation.channels:
         return None
+
+    export_user_extensions('gather_animation_hook', export_settings, animation, blender_action, blender_object)
 
     return animation
 
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_cameras.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_cameras.py
@@ -16,6 +16,7 @@ from . import gltf2_blender_export_keys
 from io_scene_gltf2.blender.exp.gltf2_blender_gather_cache import cached
 from ..com.gltf2_blender_extras import generate_extras
 from io_scene_gltf2.io.com import gltf2_io
+from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 
 import bpy
 import math
@@ -26,7 +27,7 @@ def gather_camera(blender_camera, export_settings):
     if not __filter_camera(blender_camera, export_settings):
         return None
 
-    return gltf2_io.Camera(
+    camera = gltf2_io.Camera(
         extensions=__gather_extensions(blender_camera, export_settings),
         extras=__gather_extras(blender_camera, export_settings),
         name=__gather_name(blender_camera, export_settings),
@@ -34,6 +35,10 @@ def gather_camera(blender_camera, export_settings):
         perspective=__gather_perspective(blender_camera, export_settings),
         type=__gather_type(blender_camera, export_settings)
     )
+
+    export_user_extensions('gather_camera_hook', export_settings, camera, blender_camera)
+
+    return camera
 
 
 def __filter_camera(blender_camera, export_settings):
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
@@ -26,6 +26,7 @@ from io_scene_gltf2.io.exp import gltf2_io_image_data
 from io_scene_gltf2.io.com import gltf2_io_debug
 from io_scene_gltf2.blender.exp import gltf2_blender_image
 from io_scene_gltf2.blender.exp.gltf2_blender_gather_cache import cached
+from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 
 
 @cached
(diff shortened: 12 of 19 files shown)


jjcasmar:
@julienduroure

julienduroure:
@donmccurdy @emackey @scurest This PR is quite interesting, and will need lots of testing / code review. Any help is welcome :)

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_material_normal_texture_info_class.py
```diff
@@ -60,16 +62,24 @@ def __gather_extensions(blender_shader_sockets_or_texture_slots, export_settings
...
+                return {"KHR_texture_transform": extension}
+            else:
+                extensions["KHR_texture_transform"] = extension
+                return extensions
+    return None
 
-    extension = Extension("KHR_texture_transform", texture_transform)
-    return {"KHR_texture_transform": extension}
+                
+
```
scurest: Weird whitespace.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_skins.py
```diff
@@ -55,11 +57,21 @@ def __filter_skin(blender_object, export_settings):
 
 
 def __gather_extensions(blender_object, export_settings):
-    return None
+    extension_wrapper = GLTF2ExtensionSkins(blender_object)
+    extensions = export_user_extensions(extension_wrapper, export_settings)
+    return extensions
 
 
 def __gather_extras(blender_object, export_settings):
```
scurest: Why does __gather_extras get changed only here?

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_skins.py
```diff
@@ -55,11 +57,21 @@ def __filter_skin(blender_object, export_settings):
...
-    return None
+    extensions = {}
+
+    # Handle custom user defined extensions
+    for user_extension_name, user_extension in export_settings['gltf_user_extensions'].items():
+        extension = user_extension.export(blender_object, export_settings)
+        if extension:
+            extensions[user_extension_name] = extension
+
+    return extensions
```
scurest: Should be `return extensions or None` (not that I think it matters since empty dicts get erased at some point anyway, but for consistency...)

addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
```diff
@@ -0,0 +1,133 @@
...
+        self.blender_object_if_armature = blender_object_if_armature,
+        self.bake_bone = bake_bone
+        self.bake_channel = bake_channel
+
+class GLTF2ExtensionAnimations:
+    def __init__(self, blender_action, blender_object):
+        self.blender_action = blender_action,
+        self.blender_object = blender_object
+
+class GLTF2ExtensionCameras:
```
scurest: Grammatical number seems to be inconsistent here. Some are singular (`GLTF2ExtensionMesh`, `GLTF2ExtensionNode`, etc.) and some are plural (`GLTF2ExtensionCameras`, `GLTF2ExtensionMaterials`, etc.)

addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
```diff
@@ -0,0 +1,133 @@
...
+
+class GLTF2ExtensionAnimationSamplers:
+    def __init__(self, channels, blender_object_if_armature, bake_bone, bake_channel):
+        self.channels = channels,
+        self.blender_object_if_armature = blender_object_if_armature,
+        self.bake_bone = bake_bone
+        self.bake_channel = bake_channel
+
+class GLTF2ExtensionAnimations:
+    def __init__(self, blender_action, blender_object):
```
scurest: How does this work? What action/object is it getting? Because generally an animation will correspond to a collection of Blender actions.

scurest:
The big thing I guess is backwards compatibility. This creates a public API the exporter would presumably be beholden to maintain, which seems like a big commitment.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_skins.py
jjcasmar: This is an error. gether_extras should be unchanged.

addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
jjcasmar: It works exactly the same way as in other exported element. The `__gather_extensions` in `gltf2_blender_gather_animations` receives a `blender_action` and a `blender_object`. I simply redirect those to the extension exporters.

addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
jjcasmar: I using the same notation as the gltf2_blender_gather_<stuff>.py files. There are some that are `cameras` and others that are `mesh`.

jjcasmar:
What can we do about the backward compatibility? Maybe use the Blender version number or the exporter version as some kind of extension API version? Not sure about this, but if there are no commitments at all about API stability, is hard (imposible) to expose this to the public.

addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
scurest: Node is the only difference then (`GLTF2ExtensionNode` but `gltf2_blender_gather_nodes.py`). I still think they should still have a consistent number though.

addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
scurest: Yeah, I see that you just use whatever the `__gather_extensions` function take, but is that okay? Most of these are just returning None right now anyway so there's no guarantee their arguments are actually useful or sensible enough to become a public API.

For example, I just tested this and this gets called once for each (action,blender_object) in the animation, but only the value returned by the first call for each animation actually makes it into the `extensions` dict; the rest get dropped on the floor. I don't think that's how it should work.

scurest:
Well, what's the use case for this? It's already pretty limited because you can only add data to the `extensions` object without changing the rest of the glTF. 

I'm wondering if we could design something more like how `extras` already works. Ziflin showed something in #730 where they used a regular Blender add-on to "extend" the exporter with new custom properties. Could we do something like that?

addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
jjcasmar: Well, if the __gather_extensions should receive another thing, the it's obvious it will be incorrect, but I'm not sure if its on the scope of this PR to fix what the __gather_extensions function should receive. 
If you think its on the scope, we can try to fix it, but probably we will have to review all the other __gather_extensions methods.

jjcasmar:
I dont understand what you mean when you say you can only add data to the `extensions` object. I have been able to reimplement `KHR_materials_unlit` and `KHR_punctual_lights` with this mechanism. You can add new extensions to the root extensions object and refer to them from other places in the glTF file, as proved by the KHR_punctual_lights reimplementation. Ill push it to this PR so you can see how it works

scurest:
For example, you wouldn't be able to set the [fallback behavior](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_unlit#extension-compatibility-and-fallback-materials) on a `KHR_materials_unlit` material with this would you? And doesn't knowledge about `KHR_lights_punctual` need to be integrated into the exporter anyway so it can emit the correction node? There'd be no way to do that from a user extension.

That's what I mean about only being able to add data to `extensions` dicts.

jjcasmar:
The default fallback material is already exported. The current code exports the PBR information if the shader graph has a Principled BSDF node, no matter what the extension does. Then the extension add the extension data if the shader graph has a Background node. 

For the lights you are right, its not possible to add the correction node from the extension exporter, as we can only add data to the extensions dictionaries.
How do you suggest to improve this? Im not sure if an extension should have impact outside of the extension dictionaries anyway. That means that an importer that doesn't support the extension may be importing something that is not useful and could potentially break the scene

A bit offopic, I'm looking at the current code for the exporter and wouldn't be better to add a global correction node at the beggining of the hierarchy instead of per camera/light?

scurest:
> The current code exports the PBR information if the shader graph has a Principled BSDF node, no matter what the extension does

But the PBR information depends on whether it should be unlit. For example, unlit materials should have a metallicFactor of 0 for fallback purposes, but the default metallicFactor is 1.

Right now if I import a `KHR_materials_unlit` material (in 2.8 where an Emission node is used for this because there is no Background node in 2.8) I get a material like this

```
{
    "emissiveFactor": [1, 1, 1],
    "emissiveTexture": { "index": xxx },
    "pbrMetallicRoughness": {}
}
```

A correct KHR_materials_unlit implementation should give something like this

```
{
    "pbrMetallicRoughness": {
        "baseColorTexture": { "index": xxx },
        "roughnessFactor": 0.9,
        "metallicFactor": 0.0
    },
    "extensions": { "KHR_materials_unlit": {} }
}
```

That requires more that just adding `extensions` data.

------

(The correction nodes are there to rotate the camera/light when you have a situation where glTF says it should point along one axis but in Blender it points along another axis. You can't do that with one rotation at the root.)

jjcasmar:
I understand what you mean. Maybe we can do a two step mechanism? From, what I have seen, the gltf2 object is created using something like this:

```
    node = gltf2_io.Node(
        camera=__gather_camera(blender_object, export_settings),
        children=__gather_children(blender_object, blender_scene, export_settings),
        extensions=__gather_extensions(blender_object, export_settings),
        extras=__gather_extras(blender_object, export_settings),
        matrix=__gather_matrix(blender_object, export_settings),
        mesh=__gather_mesh(blender_object, export_settings),
        name=__gather_name(blender_object, export_settings),
        rotation=None,
        scale=None,
        skin=__gather_skin(blender_object, export_settings),
        translation=None,
        weights=__gather_weights(blender_object, export_settings)
    )
```

Maybe we can keep the first step as it is right now, returning an `Extension` on the `__gather_extensions` and then do a second step in which we pass the glTF2 object to the extension plugin to modify something if its necessary. 

What do you think?

scurest:
The first step seems unnecessary. If you're going to be given the glTF object to manipulate as you please you can just add anything you want to the `extensions` property then. At that point this would no longer be limited to glTF extensions since you could also do manipulations that don't touch the `extensions` prop. You'd basically have a bunch of hooks to modify the output of the `__gather_XXX` functions just before they finish?

It's difficult to see how well this would work in practice. I think any user extension would have to track changes to the exporter pretty heavily so that in practice they would end up highly coupled anyway. The two examples we just had show that it's not so easy to write a correct user extension.

Do you have an example of a user extension you actually want to use this for?

jjcasmar:
Probably the user extension developers would need to keep track on exporter changes but I think there are some benefits:
First of all, if the developer needs to patch the exporter to add extensions, he is going to need to keep track on the changes anyway.
Second, patching the exporter and deploying a new exporter forces the users to disable/uninstall the official exporter and install the patched exporter. Not very user friendly in my opinion. 

You are right that passing around the glTF object to the extension exporter opens the door to modifying the whole node. Patching the exporter doesn't fix this, a developer can add whatever he wants. Also, if to properly implement an extension you need to modify other properties than the `extensions` property, then its clear the developer needs access to the whole glTF object. 

As use cases, at KDAB we have developed an internal extension to add tags to nodes. Basically its a new glTF array `layers` and a `layer` property on the node. I'm algo trying to draft something for supporting animating any floating point value property using fcurves. Of course its possible to implement these extensions, or any extension, just patching the exporter, but I think this has some inconvenients as already mentioned at the beginning.

scurest:
> Second, patching the exporter and deploying a new exporter forces the users to disable/uninstall the official exporter and install the patched exporter. Not very user friendly in my opinion.

Yes, and it would also let the user mix and match the user-extensions they want (if there are n user-extensions, you'd need 2^n - 1 patched exporters to represent every combination of installed user-extensions). Like I said above, I think it would be even better if we could find a way to use regular Blender addons for this. Then Blender would handle managing their lifecycles and their installation would be even more user friendly.

> You are right that passing around the glTF object to the extension exporter opens the door to modifying the whole node

I wasn't saying this was bad, just that user-extensions aren't about glTF extensions anymore; they become general plugins for modifying bits of the glTF.

> add tags to nodes

Tagging sounds like something you basically just use custom properties for though.

jjcasmar:
How do you propose to let Blender managed this? If we add the user extensions as a new Blender addon, we wont be able to call them as hooks from the exporter, no? 
Something that was in my roadmap is to let the user specify the folder where the user extensions python files are, but they will be still managed by the exporter itself, as they are with the current approach. 

> Tagging sounds like something you could basically use custom properties for though.

They are not just tags. The layers have some properties and can be shared between different nodes. They represent the layers in Qt3D.

scurest:
I dunno, does something like this work?

```py
# At beginning of each export
import sys
user_exts = {}
for addon_name in bpy.context.preferences.addons.keys():
    try:
        # Every user-extension would have a top-level class called Gltf2UserExt
        user_exts[addon_name] = sys.modules[addon_name].Gltf2UserExt()
    except Exception:
        pass

# Example hook
for user_ext in user_exts.values():
    try:
        user_ext.gather_node_hook(gltf_node, blender_object, export_settings)
    except Exception:
        pass
```

The user-extensions would need to make sure they wait until the `Gltf2UserExt` class is called into to import `io_scene_gltf2` I think.

jjcasmar:
Right now, if the user wants to enable/disable one extension, he just need to check a checkbox in the `Extensions` section of the exporter panel. Probably easy to do anyway with the approach you suggested.
You approach will allow to simplify code also, as I dont need all the code to load for the py files. 

If I understand correctly, at each place where a gltf2 object is created, I will need to call the user hooks, similar to what Im doing now, right?

Thanks for the suggestion

jjcasmar:
I have tried what you suggested, but there is an issue. The gltf2 addon is imported before the user extensions addons, so in `sys.modules[addon_name]`, the module for the user_extension doesn't exist yet. Do you know how can I force the extensions to be imported before the gltf2 addon?

scurest:
When are you searching for it? It should be okay as long as you search for it at the beginning of each export, because all the addon modules will have been loaded by then.

jjcasmar:
I was doing it at the definition of the `ExportGLTF2_Base` class. I have move it to the `invoke` method and it works, but the problem with this approach is that I am not sure how I can add properties to the extension panel in the exporter options. I have tried the following:

In the gltf2 addon:

(in the `invoke` method)
```
self.user_extensions = {}
        for addon_name in bpy.context.preferences.addons.keys():
            try:
                # Every user-extension would have a top-level class called Gltf2UserExt
                extension_ctor = getattr(sys.modules[addon_name], "glTF2UserExtension")
                extension_bl_info = getattr(sys.modules[addon_name], "bl_info")
                print(extension_bl_info)
                self.user_extensions[extension_bl_info['name']] = extension_ctor

                # The extension should return a list of properties
                self.user_extension_test_property =  bpy.props.BoolProperty(
                    name='user_extension_test_property',
                    description='This is an example of a BoolProperty used by a UserExtension.',
                    default=True
                    )
            except Exception:
                pass
        print(self.user_extensions)
```

in the extension addon:
```
(inside a panel definition)
    def draw_header(self, context):
        sfile = context.space_data
        operator = sfile.active_operator
        self.layout.prop(operator, "user_extension_test_property", text="")

    def draw(self, context):
        layout = self.layout
        layout.use_property_split = True
        layout.use_property_decorate = False  # No animation.

        sfile = context.space_data
        operator = sfile.active_operator

        sfile = context.space_data
        operator = sfile.active_operator
        extension = operator.user_extensions[bl_info['name']]

        box = layout.box()
        box.label(text="Is draft: " + str(extension.meta['isDraft']))
        box.label(text="Developer: " + str(extension.meta['developer']))
        box.label(text="url: " + str(extension.meta['url'])
```
I get this error
```
rna_uiItemR: property not found: EXPORT_SCENE_OT_gltf.user_extension_test_property
/home/kdab/.config/blender/2.81/scripts/addons/example_extension.py:44
```

image

scurest:
> I was doing it at the definition of the `ExportGLTF2_Base` class

You'd want to do it in `invoke` anyway since you want to pick up only the currently enabled addons. The user might enable, disable, or install addons in between the definition of `ExportGLTF2_Base` and when they call the exporter.

----

I would just let the user-extension addon handle that stuff on its own. Any addon can already add a panel to the "Export glTF" screen without needing any cooperation from us.

exam


<p><details><summary>Example Code</summary>

````py
class FooProps(bpy.types.PropertyGroup):
    foo_prop: bpy.props.BoolProperty(
        name='user_extension_test_property',
        description='This is an example of a BoolProperty used by a UserExtension.',
        default=True,
    )


class Foo_PT_export_foo(bpy.types.Panel):
    bl_space_type = 'FILE_BROWSER'
    bl_region_type = 'TOOL_PROPS'
    bl_label = "Foo Config"
    bl_parent_id = "FILE_PT_operator"
    bl_options = {'DEFAULT_CLOSED'}

    @classmethod
    def poll(cls, context):
        sfile = context.space_data
        operator = sfile.active_operator

        return operator.bl_idname == "EXPORT_SCENE_OT_gltf"

    def draw(self, context):
        layout = self.layout
        layout.use_property_split = True
        layout.use_property_decorate = False  # No animation.

        props = bpy.context.scene.FooProps

        layout.prop(props, 'foo_prop')


classes = (
    FooProps,
    Foo_PT_export_foo,
)


def register():
    from bpy.utils import register_class
    for cls in classes:
        register_class(cls)
    bpy.types.Scene.FooProps = bpy.props.PointerProperty(type=FooProps)


def unregister():
    from bpy.utils import unregister_class
    del(bpy.types.Scene.FooProps)
    for cls in classes:
        unregister_class(cls)
````

</details></p>

**Comment by jjcasmar** [comment]
Reviews addressed. Now extensions are loaded as addons. I have done a simple ExampleExtension. Once we avlidate this is the way to go, I will call the hooks for the rest of gltf2 nodes. Right now I'm only calling them for gltf2.Node and gltf2.Material

File: addons/io_scene_gltf2/__init__.py [hunk 6]
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
```
scurest: This should go inside of the function where you need it, not at the top-level (see #333).

addons/io_scene_gltf2/__init__.py
```diff
@@ -336,6 +334,16 @@ def invoke(self, context, event):
                 self.report({"ERROR"}, "Loading export settings failed. Removed corrupted settings")
                 del context.scene[self.scene_key]
 
+        self.user_extensions = {}
+        for addon_name in bpy.context.preferences.addons.keys():
+            try:
+                # Every user-extension would have a top-level class called Gltf2UserExt
+                extension_ctor = getattr(sys.modules[addon_name], "glTF2UserExtension")
```
scurest: You don't need the `getattr` stuff. The except block will catch any exception if it isn't there.

addons/io_scene_gltf2/__init__.py
```diff
@@ -429,6 +438,11 @@ def execute(self, context):
         export_settings['gltf_binaryfilename'] = os.path.splitext(os.path.basename(
             bpy.path.ensure_ext(self.filepath,self.filename_ext)))[0] + '.bin'
 
+        export_settings['gltf_user_extensions'] = {
```
scurest: Can't you just do `export_settings['gltf_user_extensions'] = self.user_extensions` here?

addons/io_scene_gltf2/__init__.py
```diff
@@ -927,7 +965,7 @@ def register():
     for c in classes:
         bpy.utils.register_class(c)
     # bpy.utils.register_module(__name__)
-
+
```
scurest: Trailing whitespace.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_channel_target.py
```diff
@@ -40,7 +41,7 @@ def __gather_extensions(channels: typing.Tuple[bpy.types.FCurve],
                         blender_object: bpy.types.Object,
                         export_settings,
                         bake_bone: typing.Union[str, None]
-                        ) -> typing.Any:
+                        ):
```
scurest: Why did the typing go away? (Some other places too, I won't comment on them all.)

addons/io_scene_gltf2/__init__.py
jjcasmar: indeed. Previously I was doing some checks so no all extensions were added. Now we can leave it as you said

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
```diff
@@ -55,6 +56,14 @@ def gather_material(blender_material, mesh_double_sided, export_settings):
         pbr_metallic_roughness=__gather_pbr_metallic_roughness(blender_material, orm_texture, export_settings)
     )
 
+    user_extensions = export_user_extensions(material, blender_material, export_settings)
+    if user_extensions is not None:
```
scurest: I don't see why you need to have `export_user_extensions` return anything. Just let the user-extension code add anything it wants to `material.extensions`.

misc/ExampleExtension.py
```diff
@@ -0,0 +1,76 @@
+import bpy
+import sys
+from io_scene_gltf2.io.com.gltf2_io_extensions import Extension
```
scurest: You can't import from `io_scene_gltf2` at the top level because you can't control in what order the addons get loaded and it might not exist yet. You'll need to do something like this

````py
class glTF2UserExtension:
    def __init__(self):
        import io_scene_gltf2
        self.io_scene_gltf2 = io_scene_gltf2

    def export(...):
        ext = self.io_scene_gltf2.Extension(...)
````

That's what I meant about the user-extensions needing to wait until the `Gltf2UserExt` class is called into to import `io_scene_gltf2`.

(If you want to test it, disable the glTF2 addon, restart, and then enable it to make sure it gets loaded last.)

File: misc/ExampleExtension.py [hunk 13]
```diff
@@ -0,0 +1,76 @@
...
+    bpy.utils.unregister_class(GLTF_PT_export_example_extension)
+    bpy.utils.unregister_class(ExampleExtensionProperties)
+    del bpy.types.Scene.ExampleExtensionProperties
+
+class GLTF_PT_export_example_extension(bpy.types.Panel):
+
+    bl_space_type = 'FILE_BROWSER'
+    bl_region_type = 'TOOL_PROPS'
+    bl_label = "Example Extension"
+    bl_parent_id = "GLTF_PT_export_user_extensions"
```
scurest: For the same reason that the glTF addon might not have loaded yet, you can't put the panel inside anything it will create. If you don't want it inside just `FILE_PT_operator`, gltf-blender-io and the user-extension will need to "talk" to each other somehow. Or else go back to loading them itself I guess.

jjcasmar: During the invoke method in gltf2 exporter, can I ask for the extensions to return the panels and add there th bl_parent_id?

scurest: I think you can set it in the user extension but you can't register it until after GLTF_PT_export_user_extensions has been registered.

I'm not quite sure how unregistering would need to work though. I can't tell if gltf-blender-io would need to unregister any panels that had been parented to GLTF_PT_export_user_extensions in its unregister function or not.

addons/io_scene_gltf2/__init__.py
```diff
@@ -336,6 +332,17 @@ def invoke(self, context, event):
                 self.report({"ERROR"}, "Loading export settings failed. Removed corrupted settings")
                 del context.scene[self.scene_key]
 
+        self.user_extensions = {}
+        import sys
+        for addon_name in bpy.context.preferences.addons.keys():
+            try:
+                # Every user-extension would have a top-level class called Gltf2UserExt
```
scurest: The comment says `Gltf2UserExt` but the code uses `glTF2UserExtension`. (You can just throw away my comment here if you want.)

addons/io_scene_gltf2/__init__.py
```diff
@@ -336,6 +332,17 @@ def invoke(self, context, event):
...
+                extension_ctor = sys.modules[addon_name].glTF2UserExtension
+                extension_bl_info = getattr(sys.modules[addon_name], "bl_info")
```
scurest: This `getattr` is also unnecessary.

addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2019 The glTF-Blender-IO authors.
```
jjcasmar: Is this copyright correct or should I put another thing?

misc/ExampleExtension.py
jjcasmar: Should be ready now. When the exporter looks for the extension addons, it also registers the panels. This is done with a try block, because it may try to register two times a panel if you export two or more times. The unregister function also has a try block, because if you unregister the extension without exporting, then the panel is not registered and it fails.

misc/ExampleExtension.py
```diff
@@ -0,0 +1,101 @@
...
+        box.label(text="Developer: " + str(bl_info['developer']))
+        box.label(text="url: " + str(bl_info['url']))
+
+        props = bpy.context.scene.ExampleExtensionProperties
+        layout.prop(props, 'float_property', text="Some float value")
+
+
+class glTF2UserExtension:
+
+    def __init__(self):
```
scurest: You should also read your options from the current scene in `__init__` too, since the current scene might change during the export, eg.

    self.my_props = bpy.context.scene.ExampleExtensionProperties

misc/ExampleExtension.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def export(self, gltf2_object, blender_object):
+        if isinstance(gltf2_object, self.Material):
+            if bpy.context.scene.ExampleExtensionProperties.enabled:
+                if gltf2_object.extensions is None:
+                    gltf2_object.extensions = {}
+                gltf2_object.extensions[bl_info['name']] = self.Extension(
+                    name= bl_info['name'], 
+                    extension={"float": bpy.context.scene.ExampleExtensionProperties.float_property}
+                )
```
scurest: No new line at end of file.

addons/io_scene_gltf2/__init__.py
```diff
@@ -322,6 +322,8 @@ def __init__(self):
     scene_key = "glTF2ExportSettings"
 
     #
+    # Add a user extension list. It will be filled during the operator invocation
+    user_extensions = []
```
scurest: The way this was before (on `self.user_extensions`) was correct. If you put it at the class level you're pushing new copies of stuff to it every time `invoke` gets called and you're also holding onto stuff that might have gotten unloaded because an addon was disabled etc.

jjcasmar: Would be valid if I add it as a class variable but I empty it in invoke before pushing the new extensions? I need it to be a class variable to pass CI for 279.

scurest: Is the problem with storing it on `self`? Can you just collect the user extensions in `execute` instead of `invoke`?

misc/ExampleExtension.py
```diff
@@ -0,0 +1,102 @@
+import bpy
+import sys
```
scurest: This is unused.

addons/io_scene_gltf2/__init__.py
jjcasmar: Fixed checking if the user_extensions attribute exist and assigning in that case

misc/ExampleExtension.py
scurest: I'm not sure it's right, but if it works I guess it's fine.

jjcasmar:
So if you consider like this is ready, I will add the rest of the calls to the hooks so we can merge it.

scurest:
You're still waiting to hear from someone who can merge it, so might as well wait.
@julienduroure @emackey @donmccurdy?
This has become shorter and simpler. Can you take a look?

scurest:
By the way, what do you think of a hook design where it calls `ext.hook_name(args...)` instead of `ext.export(x, y)` and then has to indentify which hook it is by the type of `x`? Like this

````py
# Identify hooks by a name instead of by the type of their first argument.
# This lets hooks have more flexible signatures (each hook can pick its
# own signature).
#
# Also export_settings will probably need to be exposed to user-extensions
# (ex: the correct node example we talked about requires knowing if the Y-UP
# conversion is enabled).


# In gltf2_io_user_extensions.py
def export_user_extensions(hook_name, export_settings, *args):
    for extension in export_settings['gltf_user_extensions']:
        hook = getattr(extension, hook_name, None)
        if hook is not None:
            hook(*args, export_settings)


# Example hooks
export_user_extensions('gather_material_hook', export_settings,
    material, blender_material)
export_user_extensions('gather_node_hook', export_settings,
    node, blender_object)
# etc


# In user-extension
class Gltf2UserExtension():

    def gather_material_hook(self, material, blender_material, export_settings):
        if not self.properties.enabled: return

        if gltf2_object.extensions is None:
            gltf2_object.extensions = {}
        gltf2_object.extensions[bl_info['name']] = self.Extension(
            name= bl_info['name'], 
            extension={"float": self.properties.float_property}
        )
````

**Comment by jjcasmar** [comment]
I like your idea! Ill implement it

**Comment by jjcasmar** [comment]
@scurest is everything correct with the patch? Can we push it a bit more so it be merged?

File: addons/io_scene_gltf2/__init__.py [hunk 21]
```diff
@@ -336,6 +336,18 @@ def invoke(self, context, event):
                 self.report({"ERROR"}, "Loading export settings failed. Removed corrupted settings")
                 del context.scene[self.scene_key]
 
+        import sys
+        self.user_extensions = []
+        for addon_name in bpy.context.preferences.addons.keys():
+            try:
+                extension_ctor = sys.modules[addon_name].glTF2UserExtension
```
scurest: Maybe put `Export` in the name, like `glTF2ExportUserExtension`? If that's kinda wordy, maybe `glTF2ExportExtension`?

Just for consistency with a hypothetical `glTF2ImportExtension` at some point in the future (julienduroure mentioned that somewhere).

addons/io_scene_gltf2/__init__.py
```diff
@@ -429,6 +441,9 @@ def execute(self, context):
         export_settings['gltf_binaryfilename'] = os.path.splitext(os.path.basename(
             bpy.path.ensure_ext(self.filepath,self.filename_ext)))[0] + '.bin'
 
+        user_extensions = getattr(self, 'user_extensions', [])
```
scurest: Okay, okay! I figured out why you're doing it this way :) `invoke` only gets called when a human uses the operator; if it's done from a script, like the testing script for the CI, only `execute` gets called, right?

Obviously we still want to use user-extensions if the exporter is called from a script though, so I think this should be like this...

In invoke, do the panel stuff:

````py
# Register extension panels
import sys
for addon_name in bpy.context.preferences.addons.keys():
    try:
        if not hasattr(sys.modules[addon_name], 'glTF2UserExtension'):
            continue
        sys.modules[addon_name].register_panel()
    except Exception:
        pass
````

Here in execute, collect the extension objects:

````py
import sys
user_extensions = []
for addon_name in bpy.context.preferences.addons.keys():
    if not hasattr(sys.modules[addon_name], 'glTF2UserExtension'):
        continue
    user_extensions.append(sys.modules[addon_name].glTF2UserExtension())
export_settings['gltf_user_extensions'] = user_extensions
````

File: addons/io_scene_gltf2/__init__.py [hunk 23]
```diff
@@ -788,6 +803,27 @@ def draw(self, context):
...
+
+        return operator.bl_idname == "EXPORT_SCENE_OT_gltf"
+
+    def draw(self, context):
+        layout = self.layout
+        layout.use_property_split = True
+        layout.use_property_decorate = False  # No animation.
+
+        sfile = context.space_data
+        operator = sfile.active_operator
```
scurest: These lines aren't doing anything are they?

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
```diff
@@ -55,6 +56,8 @@ def gather_material(blender_material, mesh_double_sided, export_settings):
         pbr_metallic_roughness=__gather_pbr_metallic_roughness(blender_material, orm_texture, export_settings)
     )
 
+    export_user_extensions('gather_materials_hook', export_settings, material, blender_material)
```
scurest: `gather_material_hook`? (With no `s`.) Since the function is called `gather_material`?

addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
```diff
@@ -0,0 +1,22 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+def export_user_extensions(hook_name, export_settings, gltf2_object, *args):
+    if gltf2_object.extensions is None:
+        gltf2_object.extensions = {}
```
scurest: I think this should be removed (user-extensions can do it themselves) and `gltf2_object` should become part of `args`. Since (a) not all user-extensions are going to do anything to the extensions property, and (b) it makes hooks more flexible, since their first argument doesn't have to be an object with an extensions property.

misc/ExampleExtension.py
```diff
@@ -0,0 +1,98 @@
...
+
+class glTF2UserExtension:
+
+    def __init__(self):
+        # We need to wait until we create the gltf2UserExtension to import the gltf2 modules
+        # Otherwise, it may fail because the gltf2 may not be loaded yet
+        from io_scene_gltf2.io.com.gltf2_io_extensions import Extension
+        from io_scene_gltf2.io.com.gltf2_io import Material
+        self.Extension = Extension 
+        self.Material = Material
```
scurest: This and the corresponding import of `Material` above are no longer used.

julienduroure:
Using this new system, is it possible to enable custom _extras_ fields, non only custom _extensions__ fields?
Thinking about it reading #542 request

jjcasmar:
> Using this new system, is it possible to enable custom _extras_ fields, non only custom _extensions__ fields?
> Thinking about it reading #542 request

Yes, the sure has access to the whole glTF node, so he can add whatever he wants there, including extensions and extras

addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
jjcasmar: Sure the user can do that himshelf, but since I have focused on user extensions, I think it was good to just provide the empty dictionary so he dont need to do the check. 

I think I dont get the second part. How does it makes the hooks more flexible? If I add the gltf node to the args, the user still needs to receive those args and parse them.

addons/io_scene_gltf2/__init__.py
jjcasmar: You need them to have access to the operator, which is used to test the name in the poll

addons/io_scene_gltf2/__init__.py
jjcasmar: Done

addons/io_scene_gltf2/__init__.py
jjcasmar: Done

addons/io_scene_gltf2/__init__.py
jjcasmar: Done

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
jjcasmar: Done

misc/ExampleExtension.py
jjcasmar: Done

addons/io_scene_gltf2/__init__.py
```diff
@@ -429,6 +443,15 @@ def execute(self, context):
         export_settings['gltf_binaryfilename'] = os.path.splitext(os.path.basename(
             bpy.path.ensure_ext(self.filepath,self.filename_ext)))[0] + '.bin'
 
+        import sys
+        user_extensions = []
+        for addon_name in bpy.context.preferences.addons.keys():
```
scurest: In 2.79, [it is apparently called `bpy.context.user_preferences`](https://docs.blender.org/api/2.79/bpy.types.Context.html#bpy.types.Context.user_preferences). So see if this fixes the CI

````py
    prefs = bpy.context.user_preferences
else:
    prefs = bpy.context.preferences
for addon_name in prefs.addons.keys():
````

File: misc/ExampleExtension.py (hunk 26 above)
Comment (scurest) [reply misc/ExampleExtension.py]: Thanks! Btw, you should have a button you can click called "Mark Conversation as Resolved" or something on these code comments that will collapse these discussions so they don't clutter up the thread as much.

File: addons/io_scene_gltf2/__init__.py (hunk 22 above)
Comment (scurest) [reply addons/io_scene_gltf2/__init__.py]: It looks like `invoke` didn't change though.

File: addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py (hunk 25 above)
Comment (scurest) [reply addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py]: It allows hooks that do not pass a glTF object as their first argument. 

I was going to say if you want to have it here, you can delete the check from the example user-extension, but I think robustness demands it be there anyway, since another user-extension may have set it to `None`.

Comment (jjcasmar) [reply addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py]: Do we want to support hooks that doesnt pass a gltf object? What would they do?

**Comment by scurest** [comment]
You'll need to rebase onto master to fix the CI.

File: addons/io_scene_gltf2/__init__.py [hunk 28]
```diff
@@ -336,6 +336,20 @@ def invoke(self, context, event):
...
+                if not hasattr(sys.modules[addon_name], 'glTF2ExportUserExtension'):
+                    continue
+                extension_ctor = sys.modules[addon_name].glTF2ExportUserExtension
+                self.user_extensions.append(extension_ctor)
+
+                # Register extension panels
+                sys.modules[addon_name].register_panel()
+            except Exception:
+                pass
+
```
scurest: You didn't update invoke.

addons/io_scene_gltf2/__init__.py
```diff
@@ -429,6 +443,19 @@ def execute(self, context):
         export_settings['gltf_binaryfilename'] = os.path.splitext(os.path.basename(
             bpy.path.ensure_ext(self.filepath,self.filename_ext)))[0] + '.bin'
 
+        user_extensions = []
+
+        if bpy.app.version < (2,80,0):
+            pass
```
scurest: Reason not to do 2.79?

addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
scurest: Modify the export settings before exporting begins.

Modify an internal primitive (which is a dict).

Modify the "schedule" for making animations from a collection of object/action pairs.

Anyway it seems better to preserve flexibility rather than try to predict what all possible desirable hooks will look like.

addons/io_scene_gltf2/__init__.py
jjcasmar: Need to check how to do the UI stuff. If posible, I prefer to get it merged and later we can work on the 2.79 version if needed

scurest: Ok. You still need to update invoke for it though.

scurest:
Alright, I don't see anything else wrong :+1:

jjcasmar:
I have added the call to the hooks for the rest of gltf objects that have a `__gather_extension` function

julienduroure:
So ... Being quite far from this development. Can you please confirm that this dev is stable and can be merged?

emackey:
@jjcasmar Thanks so much for this contribution.  I've wanted this feature for [some time](https://github.com/KhronosGroup/glTF-Blender-IO/issues/7#issuecomment-411102601), but I haven't really been able to sit and review this thoroughly until this morning.

My initial impression is this appears very stable and works well.  I do have a small concern on the UI.  When there are no custom extensions enabled, Blender still puts an extra `Extensions` dropdown on the glTF export file dialog.  Most users will see this as empty, so expanding the dropdown doesn't do anything but spin a triangle.  Is there any way to remove that?  (I'm testing with Blender 2.81a. No need for compatibility with 2.80).

I also tried going into user preferences and toggling the glTF addon itself off, and after a moment, back on.  This ensured that the glTF addon "woke up" with a custom extension already previously active.  When I did this, the custom settings appeared below (outside of) the `Extensions` dropdown on the UI.  This also resulted in an expandable dropdown being empty.

I wonder if it would be better to just get rid of the `Extensions` dropdown on the UI, and always have custom extension options simply appear underneath, separately.  It seems like that could save some concern over which addon is activated first.  Just a suggestion.

Great work on this.

jjcasmar:
@emackey thanks for your comments. 

I will review the UI stuff and submit any needed changes.

jjcasmar:
@emackey The latest version should fix the issues you have found. Now I only show the extension panel if there is some extension to be shown and I unregister the extension panels when the exporter is disabled. That way, when the exporter is enabled again, there are no panels preregistered.

addons/io_scene_gltf2/__init__.py
```diff
@@ -940,6 +986,9 @@ def register():
 def unregister():
     for c in classes:
         bpy.utils.unregister_class(c)
+    for f in extension_panel_unregister_functors:
+        f()
```
emackey: Should the `extension_panel_unregister_functors` list be cleared after this?  Or Blender always makes a fresh copy of the whole addon when re-registering?

emackey:
I tried to manually re-run the CI but I think it only ran 1 of the 4 containers.  Can you push a new commit to trigger the whole thing?  I think it should be OK now...

julienduroure:
These is in #829 a discussion about required / used extensions: how this PR manage it? (Sorry, don't have the time to test for now)

emackey:
> These is in #829 a discussion about required / used extensions: how this PR manage it? (Sorry, don't have the time to test for now)

This PR doesn't need to do anything about #829.

misc/ExampleExtension.py
```diff
@@ -0,0 +1,105 @@
...
+
+class ExampleExtensionProperties(bpy.types.PropertyGroup):
+    enabled: bpy.props.BoolProperty(
+        name='ExampleExtension_enabled',
+        description='This is an example of a BoolProperty used by a UserExtension.',
+        default=True
+        )
+    float_property: bpy.props.FloatProperty(
+        name='ExampleExtension_enabled',
+        description='This is an example of a BoolProperty used by a UserExtension.',
```
emackey: `FloatProperty`

julienduroure:
My question was (sorry, was not really clear): 
With this new way of extension creation at export, is there a way to define these new extensions as required or used? 
Can't see any info about that in misc/ExampleExtension.py file

emackey:
@julienduroure Yes, good question, I just tested this and it works.

in `misc/ExampleExtension.py` around line 101, set the `required` parameter:

```
            gltf2_object.extensions[bl_info['name']] = self.Extension(
                name= bl_info['name'], 
                extension={"float": self.properties.float_property},
                required=False
            )
```

That might be a good thing to add to this example, now that the bug is fixed in master.

The default is `True` (required), and the flag is only respected if #829 or master is merged into this branch.

julienduroure:
Thanks for your explanation.
Agree that this should be added to the example. This will not replace the documentation (that still need to be written ?), but it can lead to understand/find this parameter easier.

addons/io_scene_gltf2/__init__.py
jjcasmar: I can clear the list, but anyway, since the function runs in a try/catch scope, there is no problem is there is a dangling function trying to remove an already removed panel.

jjcasmar:
I will add that flag in the ExampleExtension.

@julienduroure where/how should I write the documentation?

julienduroure:
@jjcasmar The documentation is here: docs/blender_docs/io_scene_gltf2.rst
This file is committed in official blender documentation here : https://docs.blender.org/manual/en/latest/

addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
donmccurdy: All files should have the same header, I think – so just 2018-2019 here.

donmccurdy:
It would be a good to see an example or two of real vendor extensions implemented on this API before we consider the API stable — I expect that if we merge it now, there could be breaking changes before we get to that point. If that's OK and matches everyone's expectations, I'm fine with merging the current PR any time.

> The documentation is here: docs/blender_docs/io_scene_gltf2.rst
This file is committed in official blender documentation here : https://docs.blender.org/manual/en/latest/ ...

Perhaps just put a brief section mentioning the existence of this API in the public Blender documentation, and link to a more detailed markdown API doc for the feature in this repository?  Both can be a separate PR. I don't think we'd want to spend a lot of words on Python APIs in the otherwise UI-focused official addon documentation.

jjcasmar:
I can provide a KDAB extension if necesary. I will do this last changes ASAP

julienduroure:
Hello,
I am going to merge it, because bcon2 will start on Monday, so this is the last days to merge new big feature into Blender repository. (see #801 for more info about that). 
Polishing can be done on another PR

emackey:
Thanks again @jjcasmar, this is an important one.

jjcasmar:
Thanks @emackey . This patch is very useful for KDAB.

PR #738, opened 2019-10-17
Avoid bone.name due to apparent memory corruption effect

emackey:
I've seen random test failures (locally and online) that [look like this](https://circleci.com/gh/KhronosGroup/glTF-Blender-IO/1007#tests/containers/2).

These are intermittent and difficult to reproduce reliably, even by following an exact series of steps.  Ultimately what I think I'm seeing is some kind of memory corruption around the value of `bone.name` on [this line](https://github.com/KhronosGroup/glTF-Blender-IO/blob/1f350035f1b219e27038547062ad3f06175f6f9c/addons/io_scene_gltf2/blender/imp/gltf2_blender_skin.py#L164) just after `set_bone_transforms` is called.

This PR just swaps `bone.name` with `pynode.blender_bone_name` which should have the same value.  This is used inside of `set_bone_transforms` which makes me think it's the safer choice.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_skin.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_skin.py
@@ -161,8 +161,8 @@ class BlenderSkin():
         BlenderSkin.set_bone_transforms(gltf, skin_id, bone, node_id, parent)
         bpy.ops.object.mode_set(mode="OBJECT")
         # Custom prop on pose bone
-        if bone.name in obj.pose.bones:
-            set_extras(obj.pose.bones[bone.name], pynode.extras)
+        if pynode.blender_bone_name in obj.pose.bones:
+            set_extras(obj.pose.bones[pynode.blender_bone_name], pynode.extras)
 
     @staticmethod
     def create_vertex_groups(gltf, skin_id):


emackey:
One way to test master (without this change) is to launch Blender (2.79 or 2.8x), press "a" to select all three default objects (cube, camera, light), press "x" to delete all three, and then file -> import from here: [03_skinned_cylinder](https://github.com/KhronosGroup/glTF-Blender-IO/tree/master/tests/roundtrip/03_skinned_cylinder)

Sometimes, the import will fail with a UTF-8 exception on `bone.name`.  Other times, the import will appear to go through, but not all 4 joints will be added to Blender.  With my change, all 4 joints should import.

scurest:
I think this is because you're not supposed to access an EditBone outside of Edit mode. I can't find documentation to that effect, but I'd also get random segfaults when I did that (just an exception is pretty lucky!).

emackey:
@scurest Could be.  So switching to `pynode.blender_bone_name` would be the correct solution then?

emackey:
Or delay the switch back to object mode?

scurest:
Switching to `pynode.blender_bone_name` is good :+1: (under that theory anyway)

emackey:
Agreed, looks like we just don't touch `bone` after the switch to object mode and it's OK.

emackey:
@julienduroure Can this be included as a critical bugfix for 2.81?

julienduroure:
Setting extras for bones is not included in 2.81, so no need to include it. 

Can be merged directly in master

PR #476, opened 2019-05-17
Fix emission factor (with MULTIPLY node) on export (and use new Principled BSDF Emission slot)

runouw:
This is my first pull request, so sorry if I messed up something.

For a project I was working on, I needed support for emission factor.  I managed to get the feature working, however I am not sure if I implemented it correctly. I re-used most of the code that was used for extracting the baseColorFactor, so perhaps a refactor could remove this code duplication. Additionally, I included support for the new Principled BSDF emission slot, but to do so I wrote in a slightly hacky looking implementation (because the old code assumed emission would be using an Emission node).

I'm making this pull request in the hopes that I can speed up development of the plugin. If it's not helpful, I'm sorry!

PR #740, opened 2019-10-17
Add export support for Emission socket of Principled BSDF node.

### LINKED ISSUE #462: Principled node now has alpha and emission inputs

emackey:
This can wait for 2.82 since 2.81 is closed to new features.

Fixes #462.

Supersedes #476.


julienduroure:
@emackey For completeness of #462 fixes, we should also doing the same for alpha, not only emissive.
This can be done on another PR

PR #741, opened 2019-10-17
Export: fix #737

### LINKED ISSUE #737: Winding direction is wrong
glTF 2.0 spec defines the front face using a CCW winding direction. 
The exporter fails to export correctly a simple cube. Normals computed using the CCW winding direction point towards the inside of the cube. 

As by default materials are marked as doubleSided, this normally doesn't matter as both faces are going to be rendered, but in some cases it can give weird results. For example, exporting the default scene enabling the lights will show artifacts and a very different lighting setup in babylon.js

image

scurest:
Fixes reversed winding order when exporting n-gons in Blender versions older than 2.81.15.

Fixes #737.

edit: 03_skinned_cylinder has it out for me.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
@@ -617,6 +617,11 @@ def extract_primitives(glTF, blender_mesh, blender_object, blender_vertex_groups
             triangles = tessellate_polygon((polyline,))
 
             for triangle in triangles:
+                if bpy.app.version < (2, 81, 15):
+                    # tessellate_polygon winding-order is reversed in old versions of Blender
+                    # See https://developer.blender.org/T70594
+                    triangle = (triangle[0], triangle[2], triangle[1])
+
                 for triangle_index in triangle:
                     loop_index_list.append(blender_polygon.loop_indices[triangle_index])
         else:


emackey:
The 03_skinned_cylinder test problem will be fixed separately in #738.

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
```diff
@@ -617,6 +617,11 @@ def extract_primitives(glTF, blender_mesh, blender_object, blender_vertex_groups
             triangles = tessellate_polygon((polyline,))
 
             for triangle in triangles:
+                if bpy.app.version < (2, 81, 15):
+                    # tessellate_polygon winding-order is reversed in old versions of Blender
+                    # See https://developer.blender.org/T70594
+                    triangle = (triangle[1], triangle[0], triangle[2])
```
emackey: The [previous code here](https://github.com/KhronosGroup/glTF-Blender-IO/commit/2fb64552718638d8d4a1b80ebd6f82c9d4b293fc) was `0, 2, 1`, which I realize is the same order as `1, 0, 2`.  But could we use the original just for consistency?

Also I think you need an `else` clause around the code below.

scurest: Yeah.

emackey:
@julienduroure What's the branching strategy for 2.81 vs 2.82 for this repo?  Should master be 2.82 already?  Do we need a separate branch to track 2.81 to completion?

emackey:
Anyway this particular change doesn't affect "strict 280" output at all, it's a 2.79-only change, so it's probably safe to merge anytime.

julienduroure:
@emackey 

In blender repo : 
* master is 2.82
* 2.81 is in branch blender-v2.81-release

Here, this is the same:
* master is 2.82
* 2.81 is in branch blender-v2.81-release

PR #758, opened 2019-10-31
Import: fix TODOs in accessor decoding

scurest:
* Accessors without a buffer view are now zero initialized
* The matrix formats in the section about alignment in the spec are handled. This doesn't do anything though because nothing uses MAT2 or MAT3 accessors but I guess if they ever do get used they'll hopefully be correct now lol.

The duplication between decoding accessors and decoding their sparse data was eliminated by creating a new function `get_data_from_accessor_obj` (that way only one place needed to be fixed). While `get_data_from_accessor` takes an accessor index, `get_data_from_accessor_obj` takes an `Accessor` object and recursively calls itself to decode the sparse indices/values.

### DIFF
--- a/addons/io_scene_gltf2/io/imp/gltf2_io_binary.py
+++ b/addons/io_scene_gltf2/io/imp/gltf2_io_binary.py
@@ -15,6 +15,8 @@
 import struct
 import base64
 
+from ..com.gltf2_io import Accessor
+
 
 class BinaryData():
     """Binary reader."""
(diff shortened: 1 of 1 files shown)


julienduroure:
Hello,
@scurest 
Seems OK to me for Blender part, but at Airbus, we are using io files as API for manipulating glb file outsite of blender. We need a fonction to get binary data of an accessor, without unpack data. So I still need _get_binary_from_accessor_ function to manipulate data on binary format.

scurest:
Ok. I left `get_binary_from_accessor` in, but it's not called by anything in this repo anymore.

I changed it to return `None` when there is no buffer view on the accessor (in master it raises an exception in that case).

julienduroure:
Ok, thanks.
Maybe we can add a comment to explain that this function is not used by Blender itself, but is kept for manipulating accessor data in python

Issue #578, opened 2019-07-04
Not all animations are exported


samanthajo2:
**Describe the bug**

There are 6 animations in this file but only 5 get exported

[Pig.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3081630/Pig.blend.zip)

Same file from https://github.com/KhronosGroup/glTF-Blender-IO/issues/433

**To Reproduce**
1. Load into blender
2. File->Export->gltf2.0
3. Check to sample animations
4. Export

**Expected behavior**

6 Animations are exported

<img width="319" alt="Screen Shot 2019-07-05 at 1 06 36" src="https://user-images.githubusercontent.com/49654280/60679212-d1f6cd00-9ec1-11e9-8c59-81b33c39cf9a.png">

In particular of those 6, "Run" is missing

**Version**
 - OS: macOS 10.14.5
 - Blender Version 2.79

julienduroure:
We are currently exporting, for each object:
- active action
- action from NLA tracks

You're "run" action is not the active one, and is not in any NLA track.

We are not exporting all actions, because some actions are not linked to the object. Maybe in the future will can add a check on each action to check if this action is linked to current object.

samanthajo2:
I see. 

Just as a reference I imported the file into Unity and Unreal and both import all 6 animations so maybe importing all 6 is what users are more used to?

donmccurdy:
> Just as a reference I imported the file into Unity and Unreal and both import all 6 animations so maybe importing all 6 is what users are more used to?

This isn't really possible. Unused Actions in Blender are not associated with any particular object, and can't be exported to glTF in isolation... The requirement to export only active Actions, NLA Tracks, and stashed Actions allows the exporter to identify which Actions are associated with which objects.

SoapSpangledGames:
I stashed all of my actions, and the ones that were missing from the export before I stashed them are still missing from the export after I stashed them.

julienduroure:
@SoapSpangledGames please attach a blend file.  We can't investigate without your test case

SoapSpangledGames:
Thank you very much for your quick response. I appreciate it. I've attached my Blender file with all of the animations.

[alpha dragon stashed animations.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/5361195/alpha.dragon.stashed.animations.zip)

julienduroure:
In your file:

[...]

PR #764, opened 2019-11-07
Allow exporting non-active Actions outside NLA strip

Peach1:
Let users export all animations seen in Blender's Action list; Sort the animations so they alphabetically match the Action list in Blender.

All Action Export is the default behavior of Blender's FBX and Collada exporter, so users will think glTF animation export is broken when they see only [*some* actions export](https://github.com/KhronosGroup/glTF-Blender-IO/issues/578). Typically a user will have one blend for one character, so 'Export All Actions' is an expected default for the majority of use cases. Users can still uncheck 'Export All Actions' if they prefer to only export NLA-strip animations.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
@@ -198,6 +198,16 @@ def __get_blender_actions(blender_object: bpy.types.Object,
         if blender_object.animation_data.action is not None:
             blender_actions.append(blender_object.animation_data.action)
             blender_tracks[blender_object.animation_data.action.name] = None
+        
+        #Export all actions outside of NLA track
+        if export_settings['gltf_all_actions'] is True:
+            for action in bpy.data.actions:
+                if action is None:
+                    continue
+            
+                blender_actions.append(action)
+                blender_tracks[action.name] = None
+
 
         # Collect associated strips from NLA tracks.
         if export_settings['gltf_nla_strips'] is True:
@@ -231,5 +241,7 @@ def __get_blender_actions(blender_object: bpy.types.Object,
 
     # Remove duplicate actions.
     blender_actions = list(set(blender_actions))
+    # sort animations alphabetically (case insensitive) so they have a defined order and match Blender's Action list
+    blender_actions.sort(key = lambda a: a.name.lower())
 
     return [(blender_action, blender_tracks[blender_action.name]) for blender_action in blender_actions]
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
@@ -202,9 +202,6 @@ def __get_blender_actions(blender_object: bpy.types.Object,
         #Export all actions outside of NLA track
         if export_settings['gltf_all_actions'] is True:
             for action in bpy.data.actions:
-                if action is None:
-                    continue
-            
                 blender_actions.append(action)
                 blender_tracks[action.name] = None
 
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -259,8 +259,8 @@ class ExportGLTF2_Base:
     
     export_all_actions = BoolProperty(
         name='Export All Actions',
-        description='Include actions outside of NLA strips',
-        default=True
+        description='Include all actions outside of NLA strips',
+        default=False
     )
 
     export_def_bones = BoolProperty(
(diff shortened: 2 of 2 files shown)


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
```diff
@@ -198,6 +198,16 @@ def __get_blender_actions(blender_object: bpy.types.Object,
         if blender_object.animation_data.action is not None:
             blender_actions.append(blender_object.animation_data.action)
             blender_tracks[blender_object.animation_data.action.name] = None
+        
+        #Export all actions outside of NLA track
+        if export_settings['gltf_all_actions'] is True:
+            for action in bpy.data.actions:
+                if action is None:
```
scurest: Is this possible?

Peach1: Ah, the `None` check is just overly cautious, probably doesn't need to be there, don't think it would cause issues whether it's kept or edited out, would it?

scurest:
AFAICT this is generating one new animation for each (blender_object, action) pair? I think the name and description should be more clear about how it works.

I don't really think this is a good idea. We really do need some piece of data that associates an object with an action. I guess it works if there's one object, but otherwise you rapidly multiply confusion. For example, if you import the [VC](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/VC) sample model and then export, you'll get a glTF with ~1600 animations! Importing this back in takes over 2 minutes and I don't even want to think what would happen if I exported it again....

> All Action Export is the default behavior of Blender's FBX and Collada exporter

At least for the Collada exporter, this isn't true. It appears to only exports the active action (`object.animation_data.action` I guess) on each object. We do that too.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
scurest: No, but no reason to keep it if it never runs.

Peach1:
`All Action Export is the default behavior of Blender's FBX and Collada exporter`
...
> At least for the Collada exporter, this isn't true

Default_Export_All

For the FBX exporter, I can confirm that it exports non-active Actions that aren't in the NLA, whereas glTF exports only NLA and the active action.

Yes this should be changed for large scenes, is there any better solution? Outside of NLA, is there some way to check if an action is linked to an object?

The goal is just to have an option to 'export all' of Blender's Action list [like this user expected](https://github.com/KhronosGroup/glTF-Blender-IO/issues/578)

scurest:
I'm not sure what _Include all Actions_ does in the Collada exporter, but it doesn't appear to be this. If I import VC and export as a .dae I get only 41 animations. I'll take your word about the FBX one.

> is there any better solution?

The current one seems fine to me. Push the action to an NLA strip on the object you want it to play on. At least one other exporter does things the same way (.exm).

You have to learn about how this works but that's true of everything. You would also have to learn about how it worked if this was turned on you wanted to know why you got 1600 animations.

I think one way to make the NLA Strip animation stuff more discoverable though is to improve the UI. The current option for controlling this is _Animation > Animation > NLA Strips_ (tooltip: "Export NLA Strip animations.") This is clear as mud and anyway I'd never even check there if I was looking for why my animations weren't exporting ("I'm not using NLA strips so I can ignore that..."). Maybe something like this might give people a chance to figure out what's going on?

ffy

julienduroure:
Hello,

Agreed that the current situation is maybe not the best, but we have to be very carefull about what we are doing. Exporting all actions for each object (because action is linked to an object) is a very bad idea, I agreed with scurest.
There is no easy way, for an action not in a track, to retrieve on which object this action should be applied.
This could work if there is only 1 character is the .blend file, but not working for large files.

donmccurdy:
There's a lot of prior discussion in https://github.com/KhronosGroup/glTF-Blender-Exporter/pull/166, which provides some context on how we arrived at this solution. The current solution still seems like the best available, for now. I do like the UI/wording changes that @scurest suggests though.

The alternative is a workflow that really allows only a single character to be exported at a time, and will have bad results (very possibly affecting performance and crashing) if users don't know this.

I do wish there were a way to surface warnings (without cancelling the export) from this exporter. E.g. to warn users, "Your model has animations that weren't exported because they weren't in NLA tracks." That would avoid much of the confusion, and creating tracks or stashing actions is not all that much work.

Peach1:
Gave a better description of how animations export, and made it more discoverable like @scurest suggested. The 'All Actions' functionality is off by default since it can cause duplicate actions in multi object scenes. All Actions is useful for single character blend files

julienduroure:
Well, even with only 1 character, how does it work when the character has armature animation and some shapekeys animations ?
Will current PR, bone animations will be try to be added to mesh, and shapekeys animation will be try to be added to armature animation. This can't happen that way

julienduroure:
Same discussion occurs here #578, so closing this PR for same reason.

PR #769, opened 2019-11-12
Fix #761 wrong normal export with normal modifiers

### LINKED ISSUE #761: T71313 Wrong custom normals that depend on autosmooth angle
See https://developer.blender.org/T71313

julienduroure:


julienduroure:
Can you please have a look on this @donmccurdy ?

Issue #257, opened 2019-01-27
Import: distorted animation on joint with node scale of -1


scurest:
![robots](https://user-images.githubusercontent.com/11024420/51795367-8c620580-21a7-11e9-8683-488a459f7ae9.png)

Right-side shows how it should look. The arm looks like it's flipped, so I'm guessing this is connected to the fact that the node for the joint on that arm has a scale property of `[-1, -1, -1]`.

[Test file](https://github.com/KhronosGroup/glTF-Blender-IO/files/2799685/robot.zip)

julienduroure:
Problem identified. I will merge speedup PR #253 before working on it, it will simplify merging

julienduroure:
First step of fix is done on branch https://github.com/KhronosGroup/glTF-Blender-IO/tree/import_joint_scaled

This fix defaut bone poses, but not animation yet

PR #771, opened 2019-11-13
Fix default bone pose

julienduroure:


julienduroure:
This is first step of #257

PR #775, opened 2019-11-14
Import: import data from texture samplers

### LINKED ISSUE #687: Import: sampler.minFilter/magFilter is ignored

scurest:
Imports the texture filtering (eg. linear or nearest) and the wrap mode (eg. repeat) for textures. The following will now roundtrip successfully

example

Of the nine possible wrap modes in glTF, only two (REPEATxREPEAT and CLAMPxCLAMP) are possible in Blender (using the `extension` property at least). Blender doesn't support different modes in the s and t directions and it doesn't support MIRRORED_REPEAT. So in the end only CLAMPxCLAMP (imported as EXTEND) had been added. A warning is now printed for the wrap modes we can't handle.

The exporter was incorrectly exporting the CLIP mode as CLAMPxCLAMP instead of the EXTEND mode, so that's also been fixed too.

Closes #687.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_material_utils.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_material_utils.py
@@ -15,6 +15,8 @@
 import bpy
 from .gltf2_blender_image import BlenderImage
 from ..com.gltf2_blender_conversion import texture_transform_gltf_to_blender
+from io_scene_gltf2.io.com.gltf2_io import Sampler
+from io_scene_gltf2.io.com.gltf2_io_debug import print_console
 
 def make_texture_block(gltf, node_tree, tex_info, location, label, name=None, colorspace=None):
     """Creates a block of Shader Nodes for the given TextureInfo.
@@ -48,7 +50,12 @@ def make_texture_block(gltf, node_tree, tex_info, location, label, name=None, co
             if tex_img.image:
                 tex_img.image.colorspace_settings.is_data = True
 
-    # TODO do sampler
+    if pytexture.sampler is not None:
+        pysampler = gltf.data.samplers[pytexture.sampler]
+    else:
+        pysampler = Sampler.from_dict({})
+    set_filtering(tex_img, pysampler)
+    set_wrap_mode(tex_img, pysampler)
 
     # Mapping (transforms UVs for KHR_texture_transform)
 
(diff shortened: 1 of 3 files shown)


addons/io_scene_gltf2/blender/imp/gltf2_blender_material_utils.py
```diff
@@ -93,3 +100,65 @@ def make_texture_block(gltf, node_tree, tex_info, location, label, name=None, co
     node_tree.links.new(tex_img.inputs[0], mapping.outputs[0])
 
     return tex_img
+
+NEAREST = 9728
+LINEAR = 9729
+NEAREST_MIPMAP_NEAREST = 9984
+LINEAR_MIPMAP_NEAREST = 9985
+NEAREST_MIPMAP_LINEAR = 9986
+LINEAR_MIPMAP_LINEAR = 9987
```
julienduroure: We should centralized all the constants in a same place, for example here, where some are already listed:
io/com/gltf2_io_constants.py

addons/io_scene_gltf2/blender/imp/gltf2_blender_material_utils.py
```diff
@@ -93,3 +100,65 @@ def make_texture_block(gltf, node_tree, tex_info, location, label, name=None, co
...
+
+    # If both are nearest or the only specified one was nearest, use nearest.
+    if (minf, magf) in [(NEAREST, NEAREST), (NEAREST, None), (None, NEAREST)]:
+        tex_img.interpolation = 'Closest'
+    else:
+        tex_img.interpolation = 'Linear'
+
+CLAMP_TO_EDGE = 33071
+MIRRORED_REPEAT = 33648
+REPEAT = 10497
```
julienduroure: Constant centralization here ? io/com/gltf2_io_constants.py

addons/io_scene_gltf2/blender/imp/gltf2_blender_material_utils.py
scurest: I added a commit. Is that what you had in mind?

scurest:
Here's a test file.

[Test.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3852384/Test.zip)


out

addons/io_scene_gltf2/blender/imp/gltf2_blender_material_utils.py
julienduroure: Yes, it is :)

Issue #290, opened 2019-02-10
Import: all instances of the same mesh share the same morph weights


scurest:
For example, only one cube should be animated in this example: [MorphCube.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/2848792/MorphCube.zip)

This is caused by the fact that morph weights are per mesh instance in glTF, but there is one shape key per mesh (not per mesh instance) in Blender. I solved this by duplicating the mesh.

julienduroure:
I confirm.
We have to detect if some nodes are animated in order to know if we have to instance the mesh or create a new one

scurest:
IMO you should always create a new one, because the user can always change the weights by hand, so they would still experience "false sharing" of unanimated instances.

julienduroure:
I'm not agreed with that.
I know that there is no perfect import in that case because animation is at node level in glTF, and at mesh level in Blender, but it's really important to keep mesh instancing for performance. 

If user want to change weights by hand in a specific object, he can easily make a single-user copy of the mesh.
I think we have to be as close as possible to the glTF file, so keeping mesh instancing everytime we can

scurest:
Okay. Also of course, even unanimated instances can have different weights just because their nodes have different `"weights"` properties.

julienduroure:
Yes, of course in this case we need to create a new mesh

julienduroure:
"weights" at node level is not implemented in importer for now. In this issue, I will solve animation issue only.
I will create a new issue to manage weights at node level

PR #776, opened 2019-11-14
Import: import UV scrolling animations using EXT_property_animation

scurest:
This implements importing animations that scroll UVs by animating a `KHR_texture_transform/offset` property with the `EXT_property_animaton` extension.

out

[UVScrollExample.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3844916/UVScrollExample.zip)

Only animations that ONLY target the offset property are handled; if the rotation or scale is also animated, it gets ignored. The reason is that the curves will need to be resampled at a higher rate to handle the change in UV space between glTF and Blender (for example, a constant offset and simple linear ramp for rotation in glTF would require a sinusoidal offset in Blender). Even if the more general case were implemented, it's still desirable to have a separate codepath for handling only-offset animations because you can do it without resampling.

Also UV scrolling animations are much more common than animating the scale or rotation.

Outline of the implementation

1. call `BlenderMaterialAnim.anim` for each animation in the file
2. the data in `EXT_property_animation` gets rearranged so we can lookup a sampler from what it targets
3. one action is created per glTF material as necessary (if there are two Blender materials for a glTF material (to handle vertex color) they will use the same action)
4. any `KHR_texture_transform` that has only its offset property animated gets processed
5. the `KHR_texture_transform` object needs to be looked up because we need the constant rotation and scale values to do the glTF<->Blender UV change
6. the Mapping node is targeted by its name (eg. `baseColorTexture.mapping`) which is set in the `make_texture_block` function from #696
7. at the end, the first animation in the glTF file is restored from the NLA tracks, just like for regular animations

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation_material.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation_material.py
@@ -1,0 +1,228 @@
+# Copyright 2018-2019 The glTF-Blender-IO authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+import math
+import re
+
+import bpy
+from mathutils import Matrix
+
+from ...io.imp.gltf2_io_binary import BinaryData
+from .gltf2_blender_animation_utils import simulate_stash, restore_animation_on_object, make_fcurve
+
+
+class BlenderMaterialAnim():
+    """Blender Material Animation."""
+    def __new__(cls, *args, **kwargs):
+        raise RuntimeError("%s should not be instantiated" % cls)
+
+    @staticmethod
+    def first_match(patterns, s):
+        for pattern in patterns:
+            match = re.match(pattern, s)
+            if match:
+                return match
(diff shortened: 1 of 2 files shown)


julienduroure:
Not sure about the status of this extension.
Is it this: https://github.com/KhronosGroup/glTF/pull/1301 ?

Not sure if we have to implement non official extension. @donmccurdy  ?
Maybe #732 can be used for importing too?

scurest:
Yes, that. https://github.com/KhronosGroup/glTF/pull/1301#issuecomment-468011457 indicated an interest in implementing it here.

This case is unintrusive and could probably be done in a importer user-extension with right hooks, but animating other properties can't be so easily split from the importer. For material stuff, whether a property is animated needs to inform material creation (ex. even if a material doesn't have a `baseColorFactor`, if it's animated the node for it needs to be created) and then material creation needs to inform action creation (ex. you need to know where in the node tree `baseColorFactor` was inserted so you can target it for animation; it might have ended up directly in the PBR node, in an emission node (for unlit materials), in a multiply node (if there was also a `baseColorTexture`), etc). Also there are already places in the importer where it needs to know if something is animated (for Y-Up removal, for meshes with morph targets (#290)). So it'd be difficult to split it off like that.

donmccurdy:
> Not sure if we have to implement non official extension. @donmccurdy ?

We certainly don't have to, but we can decide. My preference would be:

- Try to support all KHR_ extensions unless there's some reason not to.
- Decide on a case-by-case basis whether to support EXT_ extensions.
- Do not implement vendor extensions (all other prefixes), leave to #732.

My only concern about this PR is that https://github.com/KhronosGroup/glTF/pull/1301 is still a draft, and most working group effort is on other extensions right now. The specification could change if/when 1301 is picked back up again, although certainly having implementations (even unmerged) helps us to make progress. So I think it's better to (a) keep this PR open for a while, (b) implement via #732, or (c) put the setting behind some kind of "experimental" option. My vote is probably (a).

PR #798, opened 2019-11-24
Export: better performance by avoiding pointless image->python data->image roundtrips

### LINKED ISSUE #790: High res textures as height for bump node cause massive freeze
**Description**
When having a material with a bump node, which takes a high resolution texture as input (in my tests they were 8k textures) and exporting as gltf 2.0, the (or at least my) PC slowly hangs up and almost crashes. I thought it might be my CPU. which couldn't keep up with the exporter, so i tried to let it sit for some time, with no results (at one point i have waited 90 minutes), but it might also be that the RAM isn't sufficient (when looking at blender in the task manager, its RAM use raises from ~500mb to almost 5gb, where it stops and freezes the whole computer, and even closes other programs).
This problem can be prevented by disabling "Material" under "Geometry" in the export setting, but it would still be great if this problem didnt exist in the first place.

**How to reproduce**
1. Create any type of 3D mesh object, and add a material
2. Create a bump node, and connect it to the normal input of the principled node
3. Create a new texture in the image panel with a high resolution (i tested with 8192x)
4. create an image texture node, select the newly created image, and connect it with the 'height' slot of the bump node
5. export the scene as gltf (with materials)
**CAUTION: This might lead your PC to freeze for half an hour or longer!**

**Expected behavior**
The Exporter creates the primitives of the mesh with the material (see console output) and then freezes. The RAM usage of blender raises tremendously, until the whole computer freezes.

**Version**
 - OS: Windows 10, Version 1903
 - Blender Version 2.80 / 2.81

My CPU is an Intel Ceon E3-1231 v3, and i have 8GB RAM, if those stats matter.

scurest:
This PR creates a "happy path" for exporting images where we don't have to round-trip through Python data, which is much faster and uses much less memory.

Fixes #790 and helps with export times generally.

## Problem

The glTF spec requires that eg. the metalllicRoughness texture have the metalness in B and the roughness in G. The exporter will assemble the appropriate image for this even if the material graph reads the metalness from, say, the R channel of one image and the roughness from the G channel of another.

To do this, *all* images are decoded to python data. The code sequence for this looks roughly like this:

    creating export image:
        pixel_data = np.array(blender_image.pixels[:])              (★1)

    pixel data may be modified to insert channels

    encoding export image:
        if created from blender image that's on disk:
            read back data from the disk

        tmp_image = bpy.data.images.new(...)
        tmp_image.pixels = pixel_data.img.flatten().tolist()        (★2)
        save tmp_image to disk and read it back

The lines marked with a ★ are slow and eat up loooots of memory. Eg. `blender_image.pixels[:]` is going to allocate a tuple of width×height×channels floats. It's really bad.

As you can see, sometimes it was possible to skip ★2 but everything did ★1.

However, for many images, like all color textures, normals, or metal/rough/occlusion textures that are already packed the way the exporter likes, this is totally wasted effort: the image data is converted to python data, nothing gets changed, and the python data is converted back to a new temp image, which is then encoded to disk and read back.

In this case it would have been better to just encode the original image!

## Solution

Change `ExportImage` to store some declarative data saying how it's channels should be filled in (ex. it says its R channel should be filled by the G channel of such-and-such Blender image) instead of storing a numpy array. I tried to write a good docstring here, so see it for details.

Then when it comes time to encode, we examine that data and determine whether all we need to do is save some existing Blender image (happy path), or whether we need to do the pixel manipulation to repack the channels (unhappy path).

The unhappy path is basically the same as before, but the happy path can be significantly better :)

## Results

Results seem pretty good. Obviously they depend on how much you can stay on the happy path but since that should include the most common texture types (base color, normal, etc.) it should be pretty often.

Export times:

| | Before | After|
|:---|---|---|
| #790 (8192x8192 normalmap) | OOM | 3s |
| DamagedHelmet | 15s | 7s |
| VC | 17s | 6s |
| Sponza | 91s | 38s |

Reviews and testing are appreciated!

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
@@ -14,121 +14,199 @@
 
 import bpy
 import os
-import typing
+from typing import Optional
 import numpy as np
 import tempfile
+import enum
+
+
+class Channel(enum.IntEnum):
+    R = 0
+    G = 1
+    B = 2
+    A = 3
+
+# These describe how an ExportImage's channels should be filled.
+
+class FillImage:
+    """Fills a channel with the channel src_chan from a Blender image."""
+    def __init__(self, image: bpy.types.Image, src_chan: Channel):
+        self.image = image
+        self.src_chan = src_chan
+
+class FillWhite:
+    """Fills a channel with all ones (1.0)."""
+    pass
 
 
 class ExportImage:
-    """Custom image class that allows manipulation and encoding of images"""
-    # FUTURE_WORK: as a method to allow the node graph to be better supported, we could model some of
-    # the node graph elements with numpy functions
+    """Custom image class.
 
-    def __init__(self, img: typing.Union[np.ndarray, typing.List[np.ndarray]], max_channels: int = 4,\
-            blender_image: bpy.types.Image = None, has_alpha: bool = False):
-        if isinstance(img, list):
-            np.stack(img, axis=2)
+    An image is represented by giving a description of how to fill its red,
+    green, blue, and alpha channels. For example:
 
-        if len(img.shape) == 2:
-            # images must always have a channels dimension
-            img = np.expand_dims(img, axis=2)
+        self.fills = {
+            Channel.R: FillImage(image=bpy.data.images['Im1'], src_chan=Channel.B),
+            Channel.G: FillWhite(),
+        }
 
-        if not len(img.shape) == 3 or img.shape[2] > 4:
-            raise RuntimeError("Cannot construct an export image from an array of shape {}".format(img.shape))
+    This says that the ExportImage's R channel should be filled with the B
+    channel of the Blender image 'Im1', and the ExportImage's G channel
+    should be filled with all 1.0s. Undefined channels mean we don't care
+    what values that channel has.
 
-        self._img = img
-        self._max_channels = max_channels
-        self._blender_image = blender_image
-        self._has_alpha = has_alpha
+    This is flexible enough to handle the case where eg. the user used the R
+    channel of one image as the metallic value and the G channel of another
+    image as the roughness, and we need to synthesize an ExportImage that
+    packs those into the B and G channels for glTF.
 
-    def set_alpha(self, alpha: bool):
-        self._has_alpha = alpha
+    Storing this description (instead of raw pixels) lets us make more
+    intelligent decisions about how to encode the image.
+    """
 
-    @classmethod
-    def from_blender_image(cls, blender_image: bpy.types.Image):
-        img = np.array(blender_image.pixels[:])
-        img = img.reshape((blender_image.size[0], blender_image.size[1], blender_image.channels))
-        has_alpha = blender_image.depth == 32
-        return ExportImage(img=img, blender_image=blender_image, has_alpha=has_alpha)
+    def __init__(self):
+        self.fills = {}
 
-    @classmethod
-    def white_image(cls, width, height, num_channels: int = 4):
-        img = np.ones((width, height, num_channels))
-        return ExportImage(img=img)
+    @staticmethod
+    def from_blender_image(image: bpy.types.Image):
+        export_image = ExportImage()
+        for chan in range(image.channels):
+            export_image.fill_image(image, dst_chan=chan, src_chan=chan)
+        return export_image
 
-    def split_channels(self):
-        """return a list of numpy arrays where each list element corresponds to one image channel (r,g?,b?,a?)"""
-        return np.split(self._img, self._img.shape[2], axis=2)
+    def fill_image(self, image: bpy.types.Image, dst_chan: Channel, src_chan: Channel):
+        self.fills[dst_chan] = FillImage(image, src_chan)
 
-    @property
-    def img(self) -> np.ndarray:
-        return self._img
+    def fill_white(self, dst_chan: Channel):
+        self.fills[dst_chan] = FillWhite()
 
-    @property
-    def shape(self):
-        return self._img.shape
+    def is_filled(self, chan: Channel) -> bool:
+        return chan in self.fills
 
-    @property
-    def width(self):
-        return self.shape[0]
+    def __on_happy_path(self) -> bool:
+        # Whether there is an existing Blender image we can use for this
+        # ExportImage because all the channels come from the matching
+        # channel of that image, eg.
+        #
+        #     self.fills = {
+        #         Channel.R: FillImage(image=im, src_chan=Channel.R),
+        #         Channel.G: FillImage(image=im, src_chan=Channel.G),
+        #     }
+        return (
+            all(isinstance(fill, FillImage) for fill in self.fills.values()) and
+            all(dst_chan == fill.src_chan for dst_chan, fill in self.fills.items()) and
+            len(set(fill.image.name for fill in self.fills.values())) == 1
+        )
 
-    @property
-    def height(self):
-        return self.shape[1]
+    def encode(self, mime_type: Optional[str]) -> bytes:
+        self.mime_type = mime_type
 
-    @property
-    def channels(self):
-        return self.shape[2]
+        # All channels are white or undefined
+        # (I don't think this should happen...)
+        if all(isinstance(fill, FillWhite) for fill in self.fills.values()):
+            return self.__encode_blank()
 
-    def __getitem__(self, key):
-        """returns a new ExportImage with only the selected channels"""
-        return ExportImage(self._img[:, :, key])
+        # Happy path = we can just use an existing Blender image
+        if self.__on_happy_path():
+            return self.__encode_happy()
 
-    def __setitem__(self, key, value):
-        """set the selected channels to a new value"""
-        if isinstance(key, slice):
-            self._img[:, :, key] = value.img
-        else:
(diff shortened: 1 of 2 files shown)


julienduroure:
I will have a look, but because material/texture is not my main domain of expertise, I hope @emackey or @donmccurdy can have a look

addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
```diff
@@ -17,118 +17,195 @@
...
-    def append(self, other):
-        if self.channels + other.channels > self._max_channels:
-            raise RuntimeError("Cannot append image data to this image "
-                               "because the maximum number of channels is exceeded.")
+    def attach(self, chan: Channel, attachment: Attachment):
+        self.fills[chan] = attachment
 
-        self._img = np.concatenate([self.img, other.img], axis=2)
+    def fill_white(self, chan: Channel):
+        self.fills[chan] = White()
```
donmccurdy: This is minor, but the naming of "attach" and "Attachment" don't feel very clear to me. On the other hand, `fill_white(Channel.B)` is quite easy to understand. Since the Attachment class is only used once outside this class, maybe replace it with a similar function that accepts an image?

```
image.fill_image(image, Channel.R)
image.fill_white(Channel.G)
```

If the container class is useful regardless, maybe names like ImageFill or SourceImage would be more obvious?

scurest: I'd like to improve the names too. `Attachment` was loosely based on color attachments for a render pass.

How about this then?

````py
# Outside this file
export_image.fill_image(image, dst_chan=dst_chan, src_chan=src_chan)
export_image.fill_white(dst_chan=dst_chan)

# Only used inside this file
class FillImage # fills with an image channel
class FillWhite # fills with all 1.0s
````

donmccurdy: Sounds good to me!

scurest: Alright, done. I also renamed `has_channel`->`is_filled`.

scurest:
I wrote a [bash script](https://gist.github.com/scurest/c10a60d362a3b9fe52eef0d55eea9471) for testing if anyone else wants to try. Make sure you have the faster-export-image branch, cd in the glTF-Blender-IO folder, and run `test.sh FILE`. FILE will be loaded and then exported with both master and the faster-export-image branch, and the .pngs will be compared with imagemagick.

So far the only difference I have found is desirable IMO: this branch will sometimes figure out it can use an existing Blender image without change, where master would have changed it.

For example, if you import MetalRoughSpheres.glb, you get this image

blender

On master, if you export, the red channel will have been filled with ones

out

whereas on this branch it exports exactly as it was.

scurest:
Oh, here's a difference: when Blender textures are missing, it was possible to have an empty ExportImage which would export as a single pixel image. In master, the image is just thrown away in this case. I'll add a commit to fix it.

scurest:
So I tested it with dirty images (ie. images that have been painted on and not saved; in the Image Editor there will be a `*` next to the Image menu if the current image is dirty) and I found another bug. After export, any image that goes down the happy path will have its dirty flag cleared since it was saved to a temp file.

This can be fixed by just doing `image.pixels[0] = image.pixels[0]` on images that were dirty to set their dirty flags again. This is can be pretty slow for some reason. Luckily we only pay this cost if an image was dirty. I added a commit to do this.

Btw, you might have noticed that the first commit also fixed a bug in master where it would use an image from disk even if the image was dirty.

scurest:
Saving a copy of an image turns out to be surprisingly hard :(

Besides clearing the dirty flag, saving images would also turn generated images into from-files-on-disk images, so I've given up on using `image.save()` to save images.

Instead I'm now using `image.save_render()` which doesn't modify the image at all so it won't have any of those problems. It does use the render settings from a scene though, so this now creates a temp scene, sets all the settings we need, and then uses `save_render()`. See the commit message on 685e9ec for details.

(Doing it this way is slightly slower unfortunately; Sponza has gone to about 40s vs 38s before; the others I tested are about the same.) 

The reason the tests are failing on 2.79 is that the test expects an RGB image and somewhere between 2.79 and 2.8 Blender learned to do

    if you call img.save_render()
        if the render color mode is RGBA
            if img is an RGB image
                save the image as an RGB file instead of an RGBA file

which makes 2.8 pass, but since 2.79 doesn't know how to do this, the PNG files it saves when it saves a copy will be RGBA images instead of RGB. As far as I could tell, there is no field that exposes if the image data is RGB or RGBA on a Blender image, so I don't know how I could fix this. (Other than having an unnecessary A channel though, I don't think there's anything actually wrong with 2.79).

I'd really love any help writing `_save_copy_of_image`. Does anyone else have any ideas? Is there some way to ask what the correct way to save a copy of an image from Python is?

julienduroure:
>  saving images would also turn generated images into from-files-on-disk images

You can get back to generative using bpy.data.images['Untitled'].source = 'GENERATED'

> clearing the dirty flag

This is dirty too, but we can change pixels[0] to any value and back to original one, it will get back the  dirty flag

scurest:
> You can get back to generative using bpy.data.images['Untitled'].source = 'GENERATED'

You can't if the image was dirty though. When you set `image.source = 'GENERATED'` the image gets recreated from the generation parameters, which means any changes you painted on top of it would be lost.

> This is dirty too, but we can change pixels[0] to any value and back to original one, it will get back the dirty flag

Yeah, just `image.pixels[0] = image.pixels[0]` works (though like I said before, it is pretty slow; seems anything touching `pixels` is).

The other problem is I don't know if this is an exhaustive list of things that are messed up by `image.save()`.

emackey:
@scurest Is `image.save_render()` only blocked by the 2.79 behavior described in your [comment](https://github.com/KhronosGroup/glTF-Blender-IO/pull/798#issuecomment-562442020) above?  We'll be dropping 2.79 soon, as you know.

The other thing I thought of is that the UI actually has something similar to what we want, saving an image as a copy from a menu item.  There's a matching Python operator:

> bpy.ops.image.save_as(save_as_render=False, copy=True, filepath="/path/file.png", relative_path=True, show_multiview=False, use_multiview=False)

I know that addons are discouraged from using `ops` for stuff like this, but it means there's code in there somewhere ([here](https://developer.blender.org/diffusion/B/browse/master/source/blender/editors/space_image/image_ops.c$1919)? [here](https://developer.blender.org/diffusion/B/browse/master/source/blender/blenkernel/intern/image_save.c$161)?) for this, even if there doesn't seem to be a non-ops Python binding for it.

Let's not let version 2.79 hold us back if that's the only stumbling block on `save_render` though.

scurest:
>  There's a matching Python operator

Yeah, I'd love to use it but I can't figure out how. Do you know?

> Is image.save_render() only blocked by the 2.79 behavior described in your comment above?

AFAIK yes.

I'm not sure if the 2.79 behaviour should really be considered blocking. It's unfortunate that an alpha channel is added when it isn't needed, but it doesn't really hurt anything except the file size. This also only happens in what should be the uncommon case, when we can't find an existing file to use for the image (eg. it was packed from channels, it was dirty, it was generated).

If another way can't be found... If the test could be relaxed to only check if the images are pixel-for-pixel equal instead of checking if the files are byte-for-byte equal, it should pass on 2.79 too. (I think that would be a good idea anyway; checking if the PNGs files are byte-for-byte equal is too restrictive. For example, the tests will break even if all we do is change the compression setting for PNGs since the file will get encoded differently even though obviously the pixels don't change.)

julienduroure:
> Yeah, I'd love to use it but I can't figure out how. Do you know?

You have to override the context, see https://docs.blender.org/api/current/bpy.ops.html#overriding-context

emackey:
> > There's a matching Python operator
> 
> Yeah, I'd love to use it but I can't figure out how. Do you know?

I haven't used these much but my understanding is that `bpy.ops.*` is tied specifically to user UI operations, such as saving an image from an Image Editor or UV Editor UI panel.  This may be a dead-end.

> checking if the PNGs files are byte-for-byte equal is too restrictive

Yeah, I put in the byte-for-byte check when I was looking to test a "happy path" where the file was simply copied from the source, and then I probably reused it and never wrote a more detailed pixel-for-pixel check.  Some sort of relaxing of tests would be fine with me, we could even isolate this test to only Blender 2.8x and above, and skip it for 2.79.

scurest:
@julienduroure Can you post the code for it?

julienduroure:
@scurest Have a look here: 
https://gist.github.com/julienduroure/83661b26fa5e1e442dcdd7873936a5a6

I didn't test in the context of addon exporting

scurest:
@julienduroure Thanks a ton! Do you know how to control the file format?

scurest:
Actually, you know what, sorry, I've dragged this PR off into the weeds. I didn't expect saving a copy of an image to be so hairy, but this case is not that important. We can just do

    tmp_img = image.copy()
    tmp_img.filepath/file_format = whatever
    tmp_img.save()

which still avoids going through Python data like master has to.

The important thing here is the restructuring of ExportImage to hold a description of how to build the image instead of holding raw pixel data so that we can avoid decoding every single image to a Python tuple. That's plenty for one PR.

I'm going to close this and open a fresh PR with just that. Trying to figure out how to save a copy of an image can always be done separately. Sorry for the noise.

Issue #554, opened 2019-06-17
Unlit materials do not round-trip


emackey:
There was much debate over "Background" vs "Emission" nodes for `KHR_materials_unlit` in #83 and #220.  But it looks like currently we're in a place where the importer uses Emission and the exporter expects Background.

As a result, `KHR_materials_unlit` is lost during round-trip from/to glTF.

In related news, `KHR_materials_unlit` is completely untested by the test suite.

donmccurdy:
My preferred solution would be https://github.com/KhronosGroup/glTF-Blender-IO/issues/220#issuecomment-455811913, with the exception that i'm fine with having the importer use Emission too. So long as it doesn't also create a Principled node on the same material, the exporter should be able to infer that it's unlit.

Issue #220, opened 2019-01-08
Add KHR_materials_unlit support to importer


emackey:
Following #211, the importer should set up a "background" node when it sees the `KHR_materials_unlit` extension.

![unlitmaterial](https://user-images.githubusercontent.com/1235080/50860041-eae35300-1362-11e9-8f02-3207a4600213.png)

emackey:
New sample model in https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/UnlitTest

emackey:
Implementation node:  If there's a BaseColor texture, it should get attached to the Color input of the Background node, with the BaseColorFactor multiplied in if needed.  ALL other textures and factors (O/R/M, Emissive, Normal) will be ignored (NOT imported) when `KHR_materials_unlit` is set.

emackey:
Is it wrong to connect an Image Texture node or RGB node directly to material output?  I realize it's a type mismatch, but it looks like Blender doesn't mind.

julienduroure:
@emackey This works on 2.80, but doesn't seem to work on 2.79
@donmccurdy 2.79 has shadeless checkbox only blender internal, not for Cycles.

Seems the LightPath trick is the better way for 2.79, and Background node for 2.80

donmccurdy:
My suggestion would be:

Importer:

- Creates a 'shadeless' (Principled BSDF?) material on import in 2.79
- Creates Background node for 2.80+

Exporter:

- Writes unlit material for Background node
- Writes unlit material for Emissive node _if_ material does not also have a Principled BSDF node
- Writes unlit material when shadeless checkbox is selected

In that case the LightPath trick will export correctly (the emissive node alone implies unlit) but it isn't required that users do so. I realize this means the importer is creating a material that only appears unlit in Blender Render mode in 2.79, but it seems like a shame to not use the shadeless checkbox when it's there...

julienduroure:
I updated my branch with nodetree with 2.79, using emission node + lightpath "is camera ray"

donmccurdy:
Thanks!

julienduroure:
Seems we have to reopen this discussion, because of #271 
Seems with our Background node setup, we can't handle alpha in texture.
#271 is already fixed for 2.79

donmccurdy:
Ah that's frustrating. 😞

[...]

PR #800, opened 2019-11-25
Emission materials should export as Unlit.

donmccurdy:
Fixes https://github.com/KhronosGroup/glTF-Blender-IO/issues/554.


donmccurdy:
^Ah, noticed I've broken a test.  Let me fix that and perhaps add another test case for this example.

donmccurdy:
Still not entirely sure this is the right way to go, see thread in https://twitter.com/jimmygunawanapp/status/1199455335066390528. But given that Blender seems to discourage use of the Background node this way, we need _something_, and an isolated Emissive node seems like the closest reasonable thing.

scurest:
Shouldn't you also set

        "roughnessFactor": 0.9,
        "metallicFactor": 0.0

for [fallback purposes](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_unlit#extension-compatibility-and-fallback-materials)?

donmccurdy:
Arguably, but it complicates an otherwise simple implementation (`__gather_roughness_factor` needs to check whether it's an emissive material), and the fallback seems unnecessary for me in what is arguably the simplest glTF extension. And if there were a fallback to PBR for the Emission node, a pure emissive material might make more sense? I guess I'd just rather not change that at all in this PR...

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
```diff
@@ -109,6 +109,10 @@ def __gather_double_sided(blender_material, mesh_double_sided, export_settings):
 
 
 def __gather_emissive_factor(blender_material, export_settings):
+    # An Emission node, alone, should be considered an Unlit material.
+    if not __has_node_of_type(blender_material, bpy.types.ShaderNodeBsdfPrincipled):
```
emackey: If the legacy custom group node is in use, this test will give the wrong answer.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
```diff
@@ -123,6 +127,10 @@ def __gather_emissive_factor(blender_material, export_settings):
 
 
 def __gather_emissive_texture(blender_material, export_settings):
+    # An Emission node, alone, should be considered an Unlit material.
+    if not __has_node_of_type(blender_material, bpy.types.ShaderNodeBsdfPrincipled):
```
emackey: Likewise here, I think the legacy group node won't be able to produce any emission.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
```diff
@@ -136,10 +144,15 @@ def __gather_extensions(blender_material, export_settings):
...
-
-    # TODO specular glossiness extension
+        if blender_material.node_tree and blender_material.use_nodes:
+            # Materials using at least one Background or Emissive node, without a
+            # Principled BSDF node, should be considered Unlit.
+            # See: https://github.com/KhronosGroup/glTF-Blender-IO/issues/554
+            has_background = __has_node_of_type(blender_material, bpy.types.ShaderNodeBackground)
+            has_emission = __has_node_of_type(blender_material, bpy.types.ShaderNodeEmission)
+            has_principled = __has_node_of_type(blender_material, bpy.types.ShaderNodeBsdfPrincipled)
+            if not has_principled and (has_background or has_emission):
```
emackey: This seems a more robust check.  Perhaps this could be encapsulated somewhere common?

tests/test/test.js
```diff
@@ -88,23 +88,23 @@ function validateGltf(gltfPath, done) {
 var assert = require('assert');
 
 // This tests floating-point numbers for equality, ignoring roundoff errors.
-assert.equalEpsilon = function(actual, expected) {
+assert.equalEpsilon = function(actual, expected, message) {
```
emackey: I like the new `message`.

tests/test/test.js
```diff
@@ -642,6 +642,34 @@ describe('Exporter', function() {
                 };
                 assert.deepStrictEqual(customNormalHash, expectedCustomNormalHash);
             });
+
+            it('exports unlit materials', function() {
```
emackey: This test needs to be protected by `if (blenderVersion !== 'blender279b')`, similar to [line 745](https://github.com/KhronosGroup/glTF-Blender-IO/blob/59f59b721ecdb278730c41f1a7cd7933c53b11b6/tests/test/test.js#L745) for example.

tests/test/test.js
```diff
@@ -88,23 +88,23 @@ function validateGltf(gltfPath, done) {
 var assert = require('assert');
 
 // This tests floating-point numbers for equality, ignoring roundoff errors.
-assert.equalEpsilon = function(actual, expected) {
+assert.equalEpsilon = function(actual, expected, message) {
     if (typeof actual !== 'number') {
         throw new Error("Expected " + actual + " to be a number.");
```
emackey: Probably need to add `${message}` to the front of this error too.

donmccurdy:
> Has your thinking evolved since #211? I'm curious what changed.

Yes, what complicated that idea was @julienduroure's observation (https://github.com/KhronosGroup/glTF-Blender-IO/issues/220#issuecomment-453760813) that Blender's documentation says *Background* nodes "should only be used for the world surface output." If you use the `Add --> Shader` menu in Blender >2.80's node-based shader window, *Background* isn't even listed as an option, although you can still select it in the *Material* sidebar.

Since an *Emission* node is created by the importer, I'm inclined to go ahead and support in on the export side too. Not the most intuitive, but at least roundtrips will work consistently.

PR #804, opened 2019-11-27
Remove unusued image code

scurest:
Removes some image stuff that isn't being used by anything (that I could find with grep)?


emackey:
@julienduroure This PR includes two file deletions that I don't see in [rBA a6008be7622c](https://developer.blender.org/rBAa6008be7622c4acd85759fbb184fc803e0d26658).  Can you delete these from the Blender repo as well?  Thanks.

julienduroure:
@emackey done :)
I should probably change my tools/convert_to_strict_2.80.py code to manage deleted files, but not sure we are going to remove other files before we drop 2.79 and stop using this script.

scurest:
Btw, for probably the same reason, the Blender repo still has io_scene_gltf2/blender/imp/gltf2_blender_texture.py, which was deleted here.

PR #807, opened 2019-11-30
Export named vertex attribute sets as custom attributes.

donmccurdy:
Proposed fix for https://github.com/KhronosGroup/glTF-Blender-IO/issues/542.

Not ready to merge – there's one TODO in the code, and I'd like to get more feedback on whether this is the right approach and whether it's useful.

Steps to test:

1. Create one or more vertex color sets.
2. Assign a custom (non-"Col") name to at least one vertex color set.
3. Export with the "Meshes -> Custom Attributes" option _enabled_.
4. Result should be something like this, given a vertex color set named "Flux":

Screen Shot 2019-11-29 at 10 49 32 PM

Screen Shot 2019-11-29 at 10 45 06 PM

This will import to three.js with the name `mesh.geometry.attributes._flux`, for example.


julienduroure:
I fixed tests (download daily build was KO), so you should merge master to re-run tests

donmccurdy:
Maybe this feature should use [Weight Painting](https://easyblend.org/html/sculpt_paint/painting/weight_paint/introduction.html) instead of vertex colors... 🤔

mgsx-dev:
i'm wondering why limit exported color layers count. Master branch [limits to 2](https://github.com/KhronosGroup/glTF-Blender-IO/blob/master/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py#L49) and current PR limits to 4.

Is there any reason to have a limit? and have it hard coded?

If so maybe we could at least log a warning during export.

donmccurdy:
Yeah, I'm not sure why there's a limit at all... I'd be fine with removing that.

Ziflin:
@donmccurdy I'm guessing this has changed as it appears that all Vertex Color data (regardless of name) is now exported as COLOR_1..N. We have a bit of a problem with this as several Blender plug-ins these days (uvpackmaster, etc.) store useful data in Vertex Colors, but of course we don't want to export that, though we do want to export our vertex colors. Is there some plan to filter vertex colors during export in a better way? Maybe a regex or wildcard to include or exclude?

donmccurdy:
> We have a bit of a problem with this as several Blender plug-ins these days (uvpackmaster, etc.) store useful data in Vertex Colors, but of course we don't want to export that ...

Hm, I'm sure some users do want to export that data as well. Two thoughts:

1. Is there a better way that we could encourage other extensions to store this data, like generic vertex attributes? I don't totally understand those but I think they are part of the geometry nodes development maybe. It would be ideal if this exporter could distinguish vertex colors from data-only attributes automatically.
2. Perhaps we could add a sub-checkbox like "Geometry -> Vertex Colors -> Visible Only"?

I would defer to @julienduroure on how (or if) to do this, I'm not able to pick this PR back up at the moment.

scurest:
Vertex color is kinda weird in another way. COLOR_0 always affects the material, COLOR_1,2,... do nothing, but you can't export COLOR_1,2,... without a COLOR_0. This is in contrast to TEXCOORD_n where you can export whatever you want but you don't have to actually use them. You _have_ to use COLOR_0 if you export it.

So far so good. But in 3.2 vertex color layers are replaced by color attributes and the _sort order is basically out of your control_. Since the first one in the list becomes the special COLOR_0, you've lost control of which one that is!

My idea was maybe color attributes should be treated like all the other generic attributes (export as _AttributeName, controlled by "Mesh > Attributes" or whatever). If "Geometry > Vertex Colors" is enabled, the _active_ color attribute would instead be exported as COLOR_0. In no case would the exporter write COLOR_1 or higher. This solves the problem of "how do you decide which one is the special COLOR_0".

Issue #203, opened 2019-01-03
Exporter performance


emackey:
Several test models I've used lately, including the sample WaterBottle model, have a very small amount of mesh data, and yet still take a long time to export compared to other Blender exporters.

I've been investigating the speed of the exporter (mostly with some hacked [logging of times](https://github.com/emackey/glTF-Blender-IO/compare/master...emackey:debug-timing-performance)).

Turns out, for the models I've been testing, the performance bottleneck is exporting the images.  In particular, the [`__get_image_data` function](https://github.com/KhronosGroup/glTF-Blender-IO/blob/3b02643cb4700efd0054e8f004065cb840dc71c0/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py#L96) spends several seconds per exported texture, slowly dismantling the color channels and reassembling them.  This is really only needed for the OcclusionRoughnessMetallic texture, and only when the user's mapping of that texture's channels do not match the glTF spec.  For other channels like BaseColor, Normal, Emission, and correctly baked ORM textures (such as from imported models), this is wasted export time.

But even more time-consuming than that, is a little parameter here in the PNG writer:

https://github.com/KhronosGroup/glTF-Blender-IO/blob/4c15398154af66e817f66a53fab3d4f03662238d/addons/io_scene_gltf2/io/exp/gltf2_io_image_data.py#L135

The number `9` here is the compression level setting, being passed to zlib.  From the [zlib manual](https://www.zlib.net/manual.html):

```
#define Z_NO_COMPRESSION         0
#define Z_BEST_SPEED             1
#define Z_BEST_COMPRESSION       9
#define Z_DEFAULT_COMPRESSION  (-1)
```

> The compression level must be `Z_DEFAULT_COMPRESSION`, or between 0 and 9: 1 gives best speed, 9 gives best compression, 0 gives no compression at all (the input data is simply copied a block at a time). `Z_DEFAULT_COMPRESSION` requests a default compromise between speed and compression (currently equivalent to level 6).

I tried setting this to `-1` for the default compression, and saw a significant speedup but some cost in file size.  I hesitate to add yet another option to our already overwhelming list of export options, but this one has a big effect on time spent and resulting size.

And of course, if the exporter already has access to a source PNG or JPG, it should be just copying the already-compressed data from there directly, not re-running the zlib compression as that takes a lot of time.

UX3D-nopper:
We will have a look at it today/tomorrow.

[...]

PR #812, opened 2019-12-01
Export: use Compositor to build images instead of pixel pushing (~4x faster at packing big metal/rough/occlusion textures)

scurest:
~~Depends on #820.~~ merged

Makes more progress on #203.

-----

Whereas #820 created a happy path for encoding images, the unhappy path is still just as bad as before. This PR aims to improve the unhappy path.

It does this with the rather hacky method of creating a temp scene and using the Compositor to construct a node tree that will render the image described by the `ExportImage`. This is straight-forward to do from the description, for example, this

```
export_image.fills = {
    Channel.R: FillImage(image=..., src_chan=Channel.A),
    Channel.G: FillWhite(),
    Channel.B: FillImage(image=..., src_chan=Channel.G),
}
```

gives this Compositor graph

compos

Then the scene is rendered to a tempfile and read back as usual.

## Color spaces

To use the Compositor, all images have their color-space set to `Non-Color` and the image is rendered with a `None` colorspace. I don't know if this is _correct_, but I _think_ this replicates what the pixel/numpy stuff was doing before. I've compared the PNGs this creates to master and found no differences (I also disabled the happy path and found no differences in the base color, etc. images either).

## Results

To test this, I made a material that intentionally goes down the unhappy path. The roughness channel is read from one channel of one generated image, and the metallic is read from one channel of another.

example

Export times by the size of the generated textures (my RAM=4GB)

| Texture Size | Before | After |
|---|---|---|
| 1024x1024 | 0.96s | 0.24s |
| 4096x4096 | 16.6s | 3.7s |
| 8192x4096 | OOM | 7.2s |
| 8192x8192 | OOM | 15.9s |

As expected, the export time is roughly linear in the number of pixels to process. Using the compositor is about 4x faster than the code using `image.pixels` and numpy (ie. it works at about 4Mpixels/second vs 1Mpixels/second), at least for large images. Memory use also went down a little bit.

In practice, again, this only changes the unhappy path, so anything that already avoids it isn't helped. However the unhappy path can take an oversized portion of the export time. For example, this knocks another 50% off the export time for DamagedHelmet on top of the 50% that #820 already took.

| | master | #820 | This PR |
|---|---|---|---|
| DamagedHelmet.glb | 15s | 7s | 3.4s |

edit: #798 became #820.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
@@ -119,7 +119,98 @@ class ExportImage:
             return self.__encode_from_image(fill.image)
 
     def __encode_unhappy(self) -> bytes:
-        # This will be a numpy array we fill in with pixel data.
+        result = self.__encode_unhappy_with_compositor()
+        if result is not None:
+            return result
+        return self.__encode_unhappy_with_numpy()
+
+    def __encode_unhappy_with_compositor(self) -> bytes:
+        # Builds a Compositor graph that will build the correct image
+        # from the description in self.fills.
+        #
+        #     [ Image ]->[ Sep RGBA ]    [ Comb RGBA ]
+        #                [  src_chan]--->[dst_chan   ]--->[ Output ]
+        #
+        # This is hacky, but is about 4x faster than using
+        # __encode_unhappy_with_numpy. There are some caveats though:
+
+        # First, we can't handle pre-multiplied alpha.
+        if Channel.A in self.fills:
+            return None
+
+        # Second, in order to get the same results as using image.pixels
+        # (which ignores the colorspace), we need to use the 'Non-Color'
+        # colorspace for all images and set the output device to 'None'. But
+        # setting the colorspace on dirty images discards their changes.
+        # So we can't handle dirty images that aren't already 'Non-Color'.
+        for fill in self.fills:
+            if isinstance(fill, FillImage):
+                if fill.image.is_dirty:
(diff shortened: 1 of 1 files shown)


julienduroure:
Something that we need to be careful about: CombineRGBA does an automatic pre-multiply of alpha on  other channels (seems that this is something that we can't disable), so it can lead to wrong output.

scurest:
Nice call! It looks like an Alpha Conversion from Straight to Premultiply (???) can fix it but I'd need to figure out why it works.

alpha-premul

scurest:
Here's another problem: images you've painted on top of are _always_ in sRGB space and ignore their colorspace_settings, so their color comes out wrong

painted-on-f

m0ppers:
OMG! thank you so much ❤️ great success here with your branch 🚀 Testing with 8k textures. Speedup is ENORMOUS and memory is finally acceptable

emackey:
@scurest Is this obsolete now that #820 is merged?

scurest:
@emackey Yeah, it was. I've rebased onto master though. It's ready for review now.

For the two problems discovered above, I just detect those cases and fallback to the numpy code. Fortunately the alpha problem doesn't affect metal-rough-occlusion textures at all, so you'll only hit it if your packing your textures and you're also painting on them in Blender—hopefully not the common case.

I considered adding a warning like `hit a slow path for image encoding; expect bad performance` in the numpy code for images greater than, say, 1Mpixel. But I don't know if it would be a good idea or not.

PR #820, opened 2019-12-07
Export: avoid pointless image<->Python data conversions

### LINKED ISSUE #790: High res textures as height for bump node cause massive freeze

scurest:
Fixes #790.

Take two on #798 m(_ _)m

Change ExportImage to store a description of how its image should be built instead of storing an array of pixel data. This lets use make more intelligent decisions when we encode, namely only doing Image<->Python data conversions when necessary for assembling an image out of channels. This gives a nice win in export time and memory use (~twice as fast for models with big textures).

Here's a list of how this works vs master

|Case|master|This PR|
|:---:|:---:|:---:|
|Image needs to be assembled out of channels|Images converted to Python data, channels packed, converted back to an Image, saved and read back|Same as master|
|Existing Image that isn't on disk (non-dirty)|Image converted to Python data, not changed, converted back to an Image, Image saved and read back|Image copied with `image.copy`, saved and read back|
|Existing Image that isn't on disk (dirty)|Same as if non-dirty|Same as master, except it avoids going through the numpy array part of the conversion|
|Existing Image that is on disk|Image converted to Python data, file on disk read back|File on disk read back|

There should be no changes from master in any of the exported images except what's described in https://github.com/KhronosGroup/glTF-Blender-IO/pull/798#issuecomment-561284590.

The only real difference with #798 is in __encode_from_image. I abandoned the attempt to save a copy of an image, and just copied the whole image and saved that normally. This should be much easier to check correctness of (except for one thing.....)


addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
```diff
@@ -14,121 +14,198 @@
...
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        tmpfilename = tmpdirname + '/img'
+        tmp_image.filepath_raw = tmpfilename
 
-            with open(tmpfilename, "rb") as f:
-                encoded_image = f.read()
+        # NOT A TYPO!!! If you delete this line, the
+        # assignment on the next line will not work.
+        tmp_image.file_format
+        tmp_image.file_format = file_format
```
scurest: Believe it or not.... :)

insanity

I can reproduce with both today's 2.82 alpha and 2.79.

Other than that, this is the same way master saves an image.

scurest:
Has anyone looked at this?

emackey:
@julienduroure Hopefully this can make it upstream in time for 2.82.  The stable branch is [scheduled to be created Jan 9th](https://devtalk.blender.org/t/30-december-2019/11068).  This is worthy of the "critical fix" tag in my humble opinion.

julienduroure:
@emackey yep, don't worry, I will be back home before 9th,  so it will be merged on time

PR #824, opened 2019-12-10
Camera import fixes

bovesan:
According to the specification:

- [Node rotation is stored as quaternion](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#node)
- [Camera field is stored as .yfov](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#cameras)

Tested OK for my use, but I have not written tests.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_camera.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_camera.py
@@ -35,6 +35,11 @@ class BlenderCamera():
         # Blender create a perspective camera by default
         if pycamera.type == "orthographic":
             cam.type = "ORTHO"
+        else:
+            if hasattr(pycamera.perspective, "yfov"):
+                cam.angle_y = pycamera.perspective.yfov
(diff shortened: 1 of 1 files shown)


julienduroure:
@scurest Can you please check if this PR is still relevant after we will merge your vnode PR #857?
(The rotation part, because I think yfov management is quite new here)

scurest:
@julienduroure It sets every node's rotation mode to quaternion, so that part is also done there.  
If you want to pull this first I can resolve the merge conflict created in #857.

julienduroure:
@bovesan Can you please rebase this PR with master?
Seems only the yfov will be needed, the other part should be already fixed by #857 

Thanks!

bovesan:
Oops. That looks wrong. Should I just do a new, fresh pull request in the morning?

bovesan:
Fixed my branch and did manual test. Camera is imported with correct FOV and rotation is set to Quaternion 👍

julienduroure:
Thanks !

PR #829, opened 2019-12-12
Only make extensions required if they are really required

m0ppers:
I have a hacked version that adds some super custom extensions to the blender export. the exporter is gladly quite easy to extend. Unfortunately right now my custom extension is always exported as being required even though I register it like this in gather_materials:

```
 extensions["AA_shadow"] = Extension("AA_shadow", {"shadowTexture": shadow}, False)
```

The last parameter is intended to control if the extension is required or not. unfortunately in the final export step this is not being honored right now


emackey:
Did you test if the Draco extension is still "required" after this change?  (I didn't test yet, but the code looks like Draco is manually added to both lists, so that sounds good).

emackey:
This should probably consider `KHR_texture_transform` and `KHR_materials_unlit` required as well.

emackey:
Update: I tested this, it's great.  Basically this is a small bugfix:  There is already infrastructure for tracking required vs. not, and the flag is just not being checked on export, which is what's fixed here.

All existing extensions are marked required by default.  We could consider relaxing this in a later PR.

emackey:
Thanks @m0ppers.

PR #831, opened 2019-12-14
Import: rewrite material code

### LINKED ISSUE #78: factor: change separate / 3 x multiply  / combine to MixRGB (multiply mode)

### LINKED ISSUE #168: Review all node locations on node trees

### LINKED ISSUE #303: Importer's handling of spec/gloss
I tried importing the [SpecGlossVsMetalRough](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/SpecGlossVsMetalRough) sample model, but the results don't look correct in Blender.  The two bottles use different techniques, but are supposed to look visually identical:

specglossvsmetalrough_visualissue

## Problem Nodes

Looking at the node graph produced by the importer, I see a couple issues:

- The specular color is not treated with the correct colorspace, per spec.  It should *not* be marked "Non-color data" because there is color in the RGB channels.  It should be a normal "Color" image.
- The roughness values are all zeros.

This screenshot shows nodes & links as created by the importer, organized for presentation here and marked up a bit, but functionally unchanged:

Imported nodes

I'll present several different options here, starting with one that I think would be the easiest path to improvement.

## Option 1

In the screenshot below, the following edits have been made to the node structure:

* Spec/Gloss image switched to "Color" mode.
* The Alpha channel of the spec/gloss image (which corresponds to Glossiness) is sent through an Invert node (`1.0 - x`), and is used as the Roughness value for both the "Glossy BSDF" and "Diffuse BSDF" nodes.
* The Glossy BSDF and Diffuse BSDF nodes are added together at full strength.  (So a "Mix" shader with a factor was replaced by an "Add" shader, with no factor).

diffuseglossnodes_simple

This node setup works pretty well for models like this one, where bright diffuse and bright specular colors are not used on the same pixel.  Technically, this is not "energy conserving", as it's missing one key formula from the `KHR_materials_pbrSpecularGlossiness` spec.  I'll address that in Option 2.

## Option 2

Building on top of Option 1, this adds 5 additional nodes to the graph.  It makes things fairly complex, but it replicates a [formula from the spec](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_pbrSpecularGlossiness#specular---glossiness) that conserves energy by not allowing the diffuse and specular channels to double-up the reflected amount of light.

The formula in question is:  *c<sub>diff</sub>* = `diffuse.rgb * (1 - max(specular.r, specular.g, specular.b))`

diffuseglossnodes_energyconserving

In the above node graph, the specular color is separated out into R, G, B components, the max of all 3 components is found, inverted, and multiplied by the diffuse color texture.  The rest of the node graph is the same as Option 1 above.

This node graph produces a result that appears to exactly match the reference.  But it's a very complex node tree, and the exporter has really zero hopes of ever exporting something like this correctly, so I would certainly understand if this path is not followed.  But this one is the most technically correct as far as I know.

## Option 3

Blender Eevee has added a new `Specular` node, specifically for the spec/gloss workflow, that can take the place of all of this.

However, there are serious caveats:

- It's only Blender 2.80 Beta, not 2.79x.
- It's only Eevee, not Cycles.
- It's Beta and subject to change before release.

But there are also some advantages:

- It has a hookup for a baked ambient occlusion map, which we've not seen before in Blender at all.
- Its readability is the highest of any of these options.
- The exporter could easily be upgraded to round-trip something like this.

eeveespecularnode

### LINKED ISSUE #355: Import: incorrect material alpha
The `baseColor` term for metallic-roughness materials is computed with the element-wise product

    baseColor = baseColorTexture * baseColorFactor * COLOR_0

I made a test file with three boxes, red, green, and blue, that each have an alpha value of 0.5. They achieve this with different combinations of textures/factors/vertex colors. Right now they look like this

[TransparentBoxes.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/2955609/TransparentBoxes.zip)

boxes-blender

Correct rendering in Babylon.js

babylon

1. **Cube 1 (Red)**

    baseColor = [1, 0, 0, 0.5]

    * No baseColorTexture
    * No vertex color
    * baseColorFactor = [1, 0, 0, 0.5]

2. **Cube 2 (Green)**

    baseColor = [0, 1, 0, 0.5]

    * baseColorTexture is a 1x1 white PNG (no alpha)
    * No vertex color
    * baseColorFactor = [0, 1, 0, 0.5]

3. **Cube 3 (Blue)**

    baseColor = [0, 0, 1, 0.5]

    * No baseColorTexture
    * Vertex color = [1, 1, 1, 0.5] for all vertices
    * baseColorFactor = [0, 0, 1, 1]

    This one might be hard to fix since AFAICT there isn't a way to get the A out of RGBA vertex colors in a shader node.

(Also cc @donmccurdy: the blue cube is also incorrect in the Three.js viewer.)

scurest:
Rewrites the material import code to fix bugs and have better composability.

Closes #78 \
Closes #168 \
Closes #303 (with option 1) \
Closes #355 

[Here](https://github.com/KhronosGroup/glTF-Blender-IO/files/3963434/MaterialTest.zip) is a test file that tests every possible node configuration so you can see how the shader graph looks.

Changes:

* Code was restructured into functions that each make one "piece" of the node tree, taking the location to create at and the sockets to attach to as arguments. This lets the pieces be moved around and combined much more freely.

  For example, changing this to use the Principled Nodes' Emission/Alpha sockets, or to use a Specular node (for option 3 of #303) would now be extremely easy.
* All nodes get their location set. No more nodes bunched up on top of each other.
* The base color term is now calculated correctly (taking into account the vertex color alpha too). The alpha is then fed into this node setup:

    alpha

    In 2.79, for the MASK alpha mode only, there is also a Greater-Than node to perform alpha clipping and the alpha blend mode is set to `CLIP`:

    alphaclip

* An occlusion texture node no longer searches for an existing ImageTexture node to use; this was buggy because it didn't take into account the UV Map or texture transform. (Also it sometimes pulled the R from the wrong texture; see `Full SpecGloss` in the test file.)
* The lightpath trick for unlit materials has been simplified slightly. In opaque materials, there is no reason to take into account the texture alpha.
* For some reason the exporter is now able to roundtrip the baseColorFactor sometimes when it couldn't before (see `BaseColor:+VertColor+AlphaFactor+Texture`, `BaseColor:+VertColor+ColorFactor+Texture`, and `BaseColor:+VertColor+ColorFactor+AlphaFactor+Texture` in the test file; there are no other differences exporting that file compared to master).


julienduroure:
Thanks a lot!

PR #836, opened 2019-12-16
Export: add option to export textures into a folder

### LINKED ISSUE #759: Feature request: An option to export textures into "textures" folder
Hi

Upon export, if "gltf Separate" option is selected, by default textures will be exported next to .gltf file. Can you please change this behavior or add an export option to create a new folder "textures" next to .gltf file and export textures into that folder?

Edit: 
Or even better, let the user to specify the relative path for exporting textures. As gltf already supports relative path in image "uri".

scurest:
For the "glTF Separate" format, you can now pick a path relative to the location of the .gltf to put texture files in. Defaults to the empty path, ie. the same folder as the .gltf, so there's no change in the default behaviour.

Closes #759

texture_dir_ui

Since the __add_image function needed to know the output path, this takes the suggestion from the TODO there and passes the export options to the constructor of the Exporter object.


Ali-RS:
Thank you @scurest  :)

PR #838, opened 2019-12-17
Allow addons with multiple extensions

jjcasmar:
This patch allows addons with multiple user extensions. It does so by returning a list of extension constructors which is then added to the list of extensions.

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -353,7 +353,7 @@ class ExportGLTF2_Base:
             preferences = bpy.context.preferences
         for addon_name in preferences.addons.keys():
             try:
-                if hasattr(sys.modules[addon_name], 'glTF2ExportUserExtension'):
+                if hasattr(sys.modules[addon_name], 'glTF2ExportUserExtension') or hasattr(sys.modules[addon_name], 'glTF2ExportUserExtensions'):
                     extension_panel_unregister_functors.append(sys.modules[addon_name].register_panel())
                     self.has_active_extenions = True
             except Exception:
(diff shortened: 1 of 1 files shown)


jjcasmar:
@scurest @emackey @julienduroure

addons/io_scene_gltf2/__init__.py
```diff
@@ -353,7 +353,7 @@ def invoke(self, context, event):
             preferences = bpy.context.preferences
         for addon_name in preferences.addons.keys():
             try:
-                if hasattr(sys.modules[addon_name], 'glTF2ExportUserExtension'):
```
scurest: I think this guard should rather check if the module hasattrs at least one of `glTF2ExportUserExtension` or `glTF2ExportUserExtensions`.

The name `register_panel` alone is too generic to assume that any addon that just has that is intending for us to call it.

addons/io_scene_gltf2/__init__.py
```diff
@@ -353,7 +353,7 @@ def invoke(self, context, event):
...
+                if hasattr(sys.modules[addon_name], 'register_panel'):
```
scurest: I think this guard should rather check if the module hasattrs at least one of `glTF2ExportUserExtension` or `glTF2ExportUserExtensions`.

The name `register_panel` alone is too generic to assume that any addon that just has that is intending for us to call it.

jjcasmar: Done

jjcasmar:
@scurest  all issues has been addressed. Can we move a bit please?

scurest:
@jjcasmar You need to ask someone who has a "Collaborator" badge next to their name. I'm just a random contributor like you, I have no power to move anything along.

jjcasmar:
@julienduroure can we move this a bit please? Without this commit, user extensions is a bit limited as you need one addon per extension, which is a bit cumbersome.

emackey:
@jjcasmar Thanks for this contribution.

@scurest I always appreciate your reviews.

barrykeenanresn:
@jjcasmar How do I use your patch to define multiple extensions? I can see some docs here on how to define a single extension: https://github.com/KhronosGroup/glTF-Blender-IO/tree/main/example-addons/example_gltf_exporter_extension

But I'm not sure how I would define multiple? Any guidance appreciated.

barrykeenanresn:
@jjcasmar This seems to work, if I just declare a list in the top-level __init__.py of my add-on:

```
from .export_extensions.one import OneExtension
from .export_extensions.two import TwoExtension

glTF2ExportUserExtensions = [OneExtension, TwoExtension]
```

PR #841, opened 2019-12-25
Import: simplify/cleanup image import

scurest:
* Remove dead `gltf_index` code.
* Use `D.images.load(check_existing=True)` to check for existing images instead of looping ourselves.
* Guard the tempfile with a `finally` to make sure it always gets cleaned up.
* Previously for the _Pack images_ option, an image file would be imported with

        read file -> write to tempfile -> load tempfile -> pack -> erase tempfile

    now it does just

        load file -> pack


julienduroure:
Thanks!

Issue #845, opened 2019-12-29
Issue trying to export an image used in an extension


jjcasmar:
I am trying to create an extension for supporting some very simple materials like spheerical mapping. The material has some parameters and an image and in Blender is implemented a ShaderNodeCustomGroup. 

For exporting this material, I am looking for the socket where the ImageTexture node is connected and I use the `gather_texture_info` function which is used for other textures in the standard material. This function returns a `gltf2_io.TextureInfo`

However, when I add this object to my extension definition, I get an error "TextureInfo is not JSON serializable" although is used several times accross the code. 

```
texture_info = self.gather_texture_info((env_socket,), export_settings)
gltf2_material.extensions[bl_info['name']] = self.Extension(
      name=bl_info['name'],
      extension={"img": texture_info},
      required=False
)
```

How should I add the texture info so It works?

scurest:
Turn it into a dict with `.to_dict()`.

jjcasmar:
Shouldn;t that be called automatically when trying to serialize the TextureInfo? For example, in gltf_io.py, for the baseColorTexture there is

```
result["baseColorTexture"] = from_union([lambda x: to_class(TextureInfo, x), from_none],
                                                self.base_color_texture)
```
which somehow calls the appropiate stuff. 

I was expecting this happens automatically given my `TextureInfo` is in the extensions object...

jjcasmar:
In fact, doing `to_dict()` gives the following error:

```
  File "/home/kdab/.config/blender/2.81/scripts/addons/kuesa/materials/KDAB_custom_materials.py", line 139, in gather_material_hook
    extension={"img": texture_info.to_dict()},
  File "/home/kdab/.config/blender/2.81/scripts/addons/io_scene_gltf2/io/com/gltf2_io.py", line 661, in to_dict
    result["index"] = from_int(self.index)
  File "/home/kdab/.config/blender/2.81/scripts/addons/io_scene_gltf2/io/com/gltf2_io.py", line 29, in from_int
    assert isinstance(x, int) and not isinstance(x, bool)
AssertionError
```

scurest:
Nothing actually gets traversed automatically in the `to_dict` pass. Every single class makes a decision about what to call on each of its fields. Extensions are just dicts and nothing gets called on their fields.

[...]

PR #849, opened 2019-12-31
Convert user extensions to dictionaries

jjcasmar:
Types used by user extensions are unknown and when the gltf dictionary is created, it can produce dictionaries with non serializable data. 

Recursively descend into extensions to convert non serializable data into a dictionary.

This allows for example to add a `TextureInfo` to an extension and export correctly, which is needed for example for pbrSpecularGlossiness

I want to thanks @scurest for helping with this issue. 
https://github.com/KhronosGroup/glTF-Blender-IO/issues/845

### DIFF
--- a/addons/io_scene_gltf2/io/com/gltf2_io.py
+++ b/addons/io_scene_gltf2/io/com/gltf2_io.py
@@ -87,6 +87,25 @@ def to_float(x):
     return x
 
 
+def extension_to_dict(obj):
+    if hasattr(obj, 'to_list'):
+        obj = obj.to_list()
+    if hasattr(obj, 'to_dict'):
+        obj = obj.to_dict()
+    if isinstance(obj, list):
+        return [extension_to_dict(x) for x in obj]
+    elif isinstance(obj, dict):
+        return {k: extension_to_dict(v) for (k, v) in obj.items()}
+    return obj
+
+def from_extension(x):
+    x = extension_to_dict(x)
+    assert isinstance(x, dict)
+    return x
+
+def from_extra(x):
+    return extension_to_dict(x)
+
 class AccessorSparseIndices:
     """Index array of size `count` that points to those accessor attributes that deviate from
     their initialization value. Indices must strictly increase.
@@ -117,9 +136,9 @@ class AccessorSparseIndices:
         result["bufferView"] = from_int(self.buffer_view)
         result["byteOffset"] = from_union([from_int, from_none], self.byte_offset)
         result["componentType"] = from_int(self.component_type)
-        result["extensions"] = from_union([lambda x: from_dict(lambda x: from_dict(lambda x: x, x), x), from_none],
+        result["extensions"] = from_union([lambda x: from_dict(from_extension, x), from_none],
                                           self.extensions)
-        result["extras"] = self.extras
+        result["extras"] = from_extra(self.extras)
         return result
 
 
@@ -152,9 +171,9 @@ class AccessorSparseValues:
         result = {}
         result["bufferView"] = from_int(self.buffer_view)
         result["byteOffset"] = from_union([from_int, from_none], self.byte_offset)
-        result["extensions"] = from_union([lambda x: from_dict(lambda x: from_dict(lambda x: x, x), x), from_none],
+        result["extensions"] = from_union([lambda x: from_dict(from_extension, x), from_none],
                                           self.extensions)
-        result["extras"] = self.extras
+        result["extras"] = from_extra(self.extras)
         return result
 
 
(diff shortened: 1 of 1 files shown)


jjcasmar:
@julienduroure @emackey

emackey:
@scurest The conversation in #845 is a little hard to follow.  Do you think this is robust?

addons/io_scene_gltf2/io/com/gltf2_io.py
```diff
@@ -117,7 +129,7 @@ def to_dict(self):
         result["bufferView"] = from_int(self.buffer_view)
         result["byteOffset"] = from_union([from_int, from_none], self.byte_offset)
         result["componentType"] = from_int(self.component_type)
-        result["extensions"] = from_union([lambda x: from_dict(lambda x: from_dict(lambda x: x, x), x), from_none],
+        result["extensions"] = from_union([lambda x: from_dict(extension_to_dict, x), from_none],
```
scurest: I like this better than what I did since it keeps it more consistent.

But it's technically now possible to do

    "extensions" : {"my_great_extension": 42}   // ie. extension value not a dict

whereas before this was impossible. (That's what I had two functions for in the PoC.) Could you replace it with a call to, like,

````python
def from_extension(x):
    x = extension_to_dict(x)
    assert isinstance(x, dict)
    return x
````

File: addons/io_scene_gltf2/io/com/gltf2_io.py [hunk 2]
```diff
@@ -1170,12 +1182,12 @@ def to_dict(self):
...
                                            self.buffer_views)
         result["cameras"] = from_union([lambda x: from_list(lambda x: to_class(Camera, x), x), from_none], self.cameras)
-        result["extensions"] = from_union([lambda x: from_dict(lambda x: from_dict(lambda x: x, x), x), from_none],
+        result["extensions"] = from_union([lambda x: from_dict(extension_to_dict, x), from_none],
                                           self.extensions)
         result["extensionsRequired"] = from_union([lambda x: from_list(from_str, x), from_none],
                                                   self.extensions_required)
         result["extensionsUsed"] = from_union([lambda x: from_list(from_str, x), from_none], self.extensions_used)
-        result["extras"] = self.extras
+        result["extras"] = extension_to_dict(self.extras)
```
scurest: I think you missed this one.

scurest:
@emackey  I'll try to summarize the issue. Here's how things normally work

* A TextureInfo stores a Texture object in it's `texture` property during the gathering phase
* After the gathering phase there is a __traverse pass where the Texture is moved into the top-level `gltf.textures` array and the `texture` property is replaced with an index into that array.
* After that there is a to_dict pass where all the gltf2_io classes are turned into dicts in preparation for JSON dumping. to_dict checks types and expects the `texture` property to be an integer.

For extensions the problem is

1. You can't put the TextureInfo object in your extension directly because the to_dict pass doesn't descend into extensions and turn objects into dicts like it does for normal properties.
2. You can't turn the TextureInfo object into a dict with to_dict during the gathering phase like eg. gltf2_blender_gather_lights does for lights because you can only call to_dict after the __traverse pass has replaced `texture` with an integer. (This isn't a problem for current extensions since they never use things like TextureInfo that are affected by __traverse.)

So this PR changes (1).

----

Like jjcasmar says, this is really the natural place for this to happen. This is what happens to everything else in the to_dict pass, and it's exactly where this would happen if to_dict _did_ understand the structure of extensions.

So it seems pretty safe to me.

jjcasmar:
Do we want to support also for extras? I'm not sure if extras should be so complex.

scurest:
I think its fine if this PR just does extensions.

emackey:
@jjcasmar Can you make the changes mentioned in @scurest's review above?  Thanks.

jjcasmar:
Done. I have also added the extras export.

scurest:
@jjcasmar It doesn't seem to have changed (maybe you forgot to `git add` or something?).

jjcasmar:
@scurest should be now

scurest:
extras don't have to be dicts. Also you missed updating the `Gltf` class.

emackey:
@jjcasmar The force-pushing is making it difficult to track all the changes coming into this branch.  Do you mind making the remaining changes as normal commits, without squash/force-push?

scurest:
@emackey (You might already know this, but you can click on the words "force-pushed" in the "jjcasmar force-pushed the..." lines to see a diff.)

jjcasmar:
I think all the comment have been addressed, right?

scurest:
* `from_extra` requires that the extras be either a dict or None. That isn't right, the spec does not constrain the type of extras. You can just do this

    ```py
    def from_extra(x):
        return extension_to_dict(x)
    ```

* You didn't update the `Gltf` class. It is still the odd man out. It looks like this

    ```py
    result["extras"] = extension_to_dict(self.extras)
    ```

    while all the other classes look like this
    ```py
    result["extras"] = from_extra(self.extras)
    ```

jjcasmar:
You are right, I completely missed that comments. Now they are addressed

emackey:
@scurest Let me know if this works for you.

@julienduroure We should probably include this in 2.82 since 3rd-party extensions are new in that version.

scurest:
@jjcasmar Cool. I tested everything. Lgtm :+1:

PR #851, opened 2020-01-03
Export: be slightly smarter when exporting existing images

scurest:
Small followup to #820.

When we go to encode an existing Blender image:

* Do not try to use a file for images that don't come from a file.
* Do not try to use a file for images that are dirty.
* Do try to use a packed file if one exists (note that the importer packs images by default).

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
@@ -179,11 +179,15 @@ class ExportImage:
 
     def __encode_from_image(self, image: bpy.types.Image) -> bytes:
         # See if there is an existing file we can use.
-        if self.file_format == image.file_format:
-            src_path = bpy.path.abspath(image.filepath_raw)
-            if os.path.isfile(src_path):
-                with open(src_path, 'rb') as f:
-                    return f.read()
+        if image.source == 'FILE' and image.file_format == self.file_format and \
+                not image.is_dirty:
+            if image.packed_file is not None:
+                return image.packed_file.data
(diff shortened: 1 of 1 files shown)


addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
```diff
@@ -179,11 +179,15 @@ def __encode_from_numpy_array(self, array: np.ndarray) -> bytes:
...
     def __encode_from_image(self, image: bpy.types.Image) -> bytes:
         # See if there is an existing file we can use.
-        if self.file_format == image.file_format:
-            src_path = bpy.path.abspath(image.filepath_raw)
-            if os.path.isfile(src_path):
-                with open(src_path, 'rb') as f:
-                    return f.read()
+        if image.source == 'FILE' and image.file_format == self.file_format and \
+                not image.is_dirty:
+            if image.packed_file is not None:
```
emackey: `packed_file` can be `not None` for `image.source == 'GENERATED'` images as well.

I think as long as the file format matches, and the packed file is non-dirty, the packed data could be returned regardless of `image.source`.

scurest: The packed file isn't actually used for GENERATED images though is it? This is the same as every image has a `generated_color` but unless the image is GENERATED, it isn't actually used for anything.

emackey: Oh I see.  I was generating an image and painting on it as a test, but once I packed it the `source` flipped back to `FILE`.

Issue #715, opened 2019-10-07
Bad roundtrip: wrong up direction


scurest:
Import [this file](https://github.com/KhronosGroup/glTF-Blender-IO/files/3695160/b00A.zip) and you get this

![ok](https://user-images.githubusercontent.com/11024420/66279201-962f2300-e875-11e9-81be-304166bf3e13.png)

Then export (with default settings) and import again and you get this

![bad](https://user-images.githubusercontent.com/11024420/66279205-a2b37b80-e875-11e9-92db-9861ff1a7c12.png)

julienduroure:
Can you please provide the original textures, because I can't read the original glb file with external viewer, missing b00A.png

Opening the generated glb (after roundtrip) seems to have the same orientation than in blender (heads are up) 
![image](https://user-images.githubusercontent.com/1696020/66285481-1fbe0f80-e8cd-11e9-8c0e-af1f038162c9.png)
So not sure what is wrong yet.

Maybe link to the fact that in glb file, joints are children of mesh, this is usually not the case in tested gltf file, and this is not the case in blender workflows. --> To be investigated

julienduroure:
Just thinking: This is a glb and saying that an external texture file is missing???
What's wrong with this glb file?

scurest:
.glbs can refer to external resources. Not everything has to be packed into it. [Here it is](https://github.com/KhronosGroup/glTF-Blender-IO/files/3697391/b00A.zip) without textures, if that helps.

----

Anyway, I was working on a branch that would get rid of the Yup2Zup object by converting glTF -> Blender coordinates immediately on reading any coordinates. You can see it [here](https://github.com/KhronosGroup/glTF-Blender-IO/compare/master...scurest:coordinate-change). This (totally unintentionally) fixes this issue.


But it doesn't round-trip bone directions and when I was investigating it, I found something kind of weird: consider a set-up like this

![bones](https://user-images.githubusercontent.com/11024420/66315218-f6a47b80-e8da-11e9-8d7d-6473b70791fb.png)

Bones point along their +Y axis in Blender. That's the -Z axis after changing to glTF space. So I expected the upper bone to be translated by, like, [0,0,-1] in glTF. But it's translated by [0,1,0], no matter whether you export with _Transform > +Y-Up_ option or not.

Do you know why that is?

julienduroure:
Thanks for the new file. I will check it.

[...]

Issue #324, opened 2019-02-25
Imported armature issues


Quinten123:
[Hummingbird_Gltf.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/2899935/Hummingbird_Gltf.zip)

Whenever i import this gltf file the armature comes out weird. Bones are uncorrectly sized, unconnected to the parent, and rotated. 
![screensht_1](https://user-images.githubusercontent.com/42477397/53330092-ef050900-38ed-11e9-8a07-5bbe72353756.JPG)
![screensht_2](https://user-images.githubusercontent.com/42477397/53330097-f3312680-38ed-11e9-8f21-5e52784d75a0.png)

julienduroure:
I don't think there is any issue here.
"bone" object (with a head and a tail) has no equivalent in glTF (where you can find "joint"). That means that :
- There is no size stored in glTF, because a joint is ponctual
- bones are correctly parented, but there is no notion of connection between joints (except parent /  child relationship that are OK in your example file)
- rotation seems to be legit regarding data that are glTF file
- animation seems to be OK

I will set this issue as "enhancement", because we probably could fake some rotation on created bones to have them visually better aligned.

Quinten123:
Thank you, I do hope it will it will be improved, so it's easier to see the structure of the bones and work with them more easily. Right now it's a bit difficult. If you need more gltf files like this i have a bunch more.

scurest:
Bones are rotated and sized plausibly in [gltf-blender-importer](https://github.com/ksons/gltf-blender-importer/issues). At least non-leaf bones

![hum](https://user-images.githubusercontent.com/11024420/53841604-ca9ad380-3f62-11e9-9b2e-a403271df644.png)

Basically, we pick a rotation cr(b) for each bone b, and the local-to-parent T(b) for that bone is replaced with T'(b) = cr(pb)^{-1} T(b) cr(b) where pb is the parent of b. Then when you compose with the parent, T'(pb) T'(b), the cr(pb)^{-1} at the right of T(b) is cancelled out by the cr(pb) on the left of T(pb), and so on up the tree, so it doesn't end up affecting the skin. You have to maintain the relation between T'(b) and T(b) when you animate pose bones too.

[...]

Issue #251, opened 2019-01-24
Import: exception when skin["inverseBindMatrices"] is missing


scurest:
````
Traceback (most recent call last):
  File "/home/scurest/.config/blender/2.79/scripts/addons/io_scene_gltf2/__init__.py", line 467, in execute
    return self.import_gltf2(context)
  File "/home/scurest/.config/blender/2.79/scripts/addons/io_scene_gltf2/__init__.py", line 485, in import_gltf2
    BlenderGlTF.create(self.gltf_importer)
  File "/home/scurest/.config/blender/2.79/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py", line 36, in create
    BlenderScene.create(gltf, scene_idx)
  File "/home/scurest/.config/blender/2.79/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_scene.py", line 84, in create
    BlenderNode.create(gltf, node_idx, None)  # None => No parent
  File "/home/scurest/.config/blender/2.79/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_node.py", line 105, in create
    BlenderSkin.create_bone(gltf, pynode.skin_id, node_idx, parent)
  File "/home/scurest/.config/blender/2.79/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_skin.py", line 152, in create_bone
    BlenderSkin.set_bone_transforms(gltf, skin_id, bone, node_id, parent)
  File "/home/scurest/.config/blender/2.79/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_skin.py", line 58, in set_bone_transforms
    inverse_bind_matrices = BinaryData.get_data_from_accessor(gltf, pyskin.inverse_bind_matrices)
  File "/home/scurest/.config/blender/2.79/scripts/addons/io_scene_gltf2/io/imp/gltf2_io_binary.py", line 50, in get_data_from_accessor
    accessor = gltf.data.accessors[accessor_idx]
TypeError: list indices must be integers or slices, not NoneType
````

[Test file](https://github.com/KhronosGroup/glTF-Blender-IO/files/2789730/noinvbinds.zip)

According to the spec the default is that all the inverse bind matrices are the identity.

PR #857, opened 2020-01-04
Import: fix skinning (and other node hierarchy problems)

### LINKED ISSUE #239: Import: exception for 'Node' object has no attribute 'blender_bone_name'
```
  File "/home/scurest/.config/blender/2.79/scripts/addons/io_scene_gltf2/__init__.py", line 468, in execute
    return self.import_gltf2(context)
  File "/home/scurest/.config/blender/2.79/scripts/addons/io_scene_gltf2/__init__.py", line 487, in import_gltf2
    BlenderGlTF.create(self.gltf_importer)
  File "/home/scurest/.config/blender/2.79/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py", line 35, in create
    BlenderScene.create(gltf, scene_idx)
  File "/home/scurest/.config/blender/2.79/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_scene.py", line 78, in create
    BlenderSkin.create_vertex_groups(gltf, skin_id)
  File "/home/scurest/.config/blender/2.79/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_skin.py", line 163, in create_vertex_groups
    obj.vertex_groups.new(name=gltf.data.nodes[bone].blender_bone_name)
AttributeError: 'Node' object has no attribute 'blender_bone_name'
```

[Test file](https://github.com/KhronosGroup/glTF-Blender-IO/files/2777901/b00A.zip)

Not sure what's wrong, but maybe caused by a mesh being instantiated on one of the joints of its skin.

### LINKED ISSUE #257: Import: distorted animation on joint with node scale of -1

### LINKED ISSUE #483: Maya exported GLTF2 model breaks
**Describe the bug**
Maya exported model works in three.js, but not Blender.

**To Reproduce**
1. Create model in Maya
2. Export using https://github.com/WonderMediaProductions/Maya2glTF/commit/d30636ffa1a47415720cbdb9b1ad51534902c22f
3. Import in Blender https://github.com/KhronosGroup/glTF-Blender-IO/commit/fbb2c50723e5eaa2635672fedfb0f53df7d0798f
4. Breaks in Blender.
5. Open in three.js via Visual Studio Code. Looks fine.

**Expected behavior**

Render like three.js.

**Screenshots**

Screen Shot 2019-05-20 at 6 48 00 PM

Screen Shot 2019-05-20 at 6 46 05 PM

Screen Shot 2019-05-20 at 6 49 19 PM

**.blend file/ .gltf**
If applicable, a zipped folder containing a .blend file for exporter issues and a .gltf file for importer issues. Please take care to include all resources, such as textures in the .zip file.

https://github.com/assimp/assimp-mdb/tree/master/model-db/fbx/sphere_rig_advanced

[sphere_rig_advanced.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3200404/sphere_rig_advanced.zip)

**Version**
 - OS: osx
 - Blender Version 2.80 May 20, 2019

### LINKED ISSUE #509: Model animation imports incorrectly
Only one part of the model animates, the same model animates correctly on https://gltf-viewer.donmccurdy.com/

**Screenshots**
image

**.blend file/ .gltf**
[bigvic.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3238846/bigvic.zip)
Also contains the FBX which I converted to glTF

**Version**
 - OS: Linux/Windows
 - Blender Version 2.80.190529.fee600f47980-1

edit: same behaviour with 30th May Blender and git head of the gltf plugin

### LINKED ISSUE #698: Gltf2 Import Error Aquarius
**Describe the bug**

Animated model is broken.

**To Reproduce**
1. Go to https://sketchfab.com/3d-models/aquarius-astellia-622bde32272d4545b89cc09113ca7439
2. Click on download fbx
3. Open in Maya 
4. Use maya2gltf develop branch https://github.com/WonderMediaProductions/Maya2glTF/commit/523fd6d1148a99dc4b6e732d10c73a44f6071e97
5. Export gltf2  with 0.01 scaling and -glb
3. Import gltf2 animation
4. Animations are broken.

**Expected behavior**
Correct animations.

**Screenshots**

Capture

Capture2

**.blend file/ .gltf**

[aquarius-astellia.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3666195/aquarius-astellia.zip)

[export.gltf.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3666235/export.gltf.zip)

**Version**
 - OS: Windows
 - Blender Version 2.8.0
 - gltf2-io e5e86a87f6cc8fe736c5918fc8acc60fc2394d37
 - maya2gltf develop branch https://github.com/WonderMediaProductions/Maya2glTF/commit/523fd6d1148a99dc4b6e732d10c73a44f6071e97

### LINKED ISSUE #699: Gltf2 Import Error Kera
**Describe the bug**

Animated model is broken.

**To Reproduce**
1. Go to https://sketchfab.com/3d-models/kera-the-archer-desert-fox-animated-2f1f09ce84da452b90939b54e6bc2c0a
2. Click on download gltf
3. Import gltf2 animation
4. Animations are broken.

**Expected behavior**
Correct animations.

**Screenshots**

Capture

Capture2

**.blend file/ .gltf**

[kera-the-archer-desert-fox-animated.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3666198/kera-the-archer-desert-fox-animated.zip)

**Version**
 - OS: Windows
 - Blender Version 2.8.0
 - gltf2-io e5e86a87f6cc8fe736c5918fc8acc60fc2394d37

### LINKED ISSUE #700: Gltf2 Import Error Make Human
**Describe the bug**

Animated model is broken.

**To Reproduce**
1. Go to https://sketchfab.com/3d-models/make-human-01-1-96b0e226131343f99a1724f41e453dde#download
1. Export from maya2gltf with 0.01 scaling and `-glb`.
1. Open gltf2 in blender
1. Error

**Expected behavior**
Correct animations.

**Screenshots**
Capture

Capture2

**.blend file/ .gltf**

[export.gltf.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3666230/export.gltf.zip)

[make-human.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3666213/make-human.blend.zip)

[export.glb.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3666216/export.glb.zip)

**Version**
 - OS: Windows
 - Blender Version 2.8.0
 - gltf2-io e5e86a87f6cc8fe736c5918fc8acc60fc2394d37
 - maya2gltf develop branch https://github.com/WonderMediaProductions/Maya2glTF/commit/523fd6d1148a99dc4b6e732d10c73a44f6071e97

### LINKED ISSUE #703: Gltf2 Import Error Witch
**Describe the bug**

Animated model is broken.

**To Reproduce**
1. Go to  https://sketchfab.com/3d-models/witchapprentice-c015ab905a26410092db35c946a68032
2. Click on download fbx
3. Open in Maya 
4. Use maya2gltf develop branch https://github.com/WonderMediaProductions/Maya2glTF/commit/523fd6d1148a99dc4b6e732d10c73a44f6071e97
5. Export gltf2  with 0.01 scaling and -glb
3. Import gltf2 animation
4. Animations are broken.

**Expected behavior**
Correct animations.

**Screenshots**

Capture2
Capture

**.blend file/ .gltf**

[witchapprentice.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3667885/witchapprentice.zip)

[witchapprentice.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3667887/witchapprentice.blend.zip)

**Version**
 - OS: Windows
 - Blender Version 2.8.0
 - gltf2-io e5e86a87f6cc8fe736c5918fc8acc60fc2394d37
 - maya2gltf develop branch https://github.com/WonderMediaProductions/Maya2glTF/commit/523fd6d1148a99dc4b6e732d10c73a44f6071e97

### LINKED ISSUE #847: Athena - Import Difference between Three.js and glTF-Blender-IO
**Describe the bug**
I'm trying to write a GLTF2 exporter. So the output is somewhat out of specification and rough, but the animation in this file plays differently in Blender compared to Three.js.

**To Reproduce**
1. Go to blender
2. Import gltf.
3. See that the female character's hair moves and not the entire body.

**Expected behavior**
The animation should play like in Codium's three.js.

**Screenshots**
Screenshot from 2019-12-30 09-01-00
Screenshot from 2019-12-30 09-00-01

**.blend file/ .gltf**

[athena.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4010424/athena.zip)

**Version**
 - OS: Linux
 - Blender Version 2.81

**Additional context**
The main difference is three.js works and Blender import doesn't. So this import problem is blocking my use case of doing gltf2 exports to Blender.

### LINKED ISSUE #848: Ruby Rose - Import Difference between Three.js and glTF-Blender-IO
**Describe the bug**
I'm trying to write a GLTF2 exporter. So the output is somewhat out of specification and rough, but the animation in this file plays differently in Blender compared to Three.js.

**To Reproduce**
1. Go to blender
2. Import gltf.
3. See that the female character moves.

**Expected behavior**
The animation should play like in Codium's three.js.

**Screenshots**

image

image

**.blend file/ .gltf**
[ruby_rose.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4010493/ruby_rose.zip)

The blend file is 40mb, so you can open the gltf2 in blender 2.81. I can post it if you want.

**Version**
 - OS: Linux
 - Blender Version 2.81

**Additional context**

https://sketchfab.com/3d-models/ruby-rose-2270ee59d38e409491a76451f6c6ef80

The main difference is three.js works and Blender import doesn't. So this import problem is blocking my use case of doing gltf2 exports to Blender.

### LINKED ISSUE #850: Mannequin Top to Bottom - Import Difference between Three.js and glTF-Blender-IO
**Describe the bug**
I'm trying to write a GLTF2 exporter. The animation in this file plays differently in Blender compared to Three.js.

**To Reproduce**
1. Go to blender
2. Import gltf.
3. See that the entire level moves instead of the character.

**Expected behavior**
The animation should play like in Codium's three.js.

**Screenshots**

VSCodium's three.js

Screen Shot 2020-01-01 at 9 42 59 AM

Screen Shot 2020-01-01 at 9 49 48 AM

VSCodium with Three.js
[Screen Recording 2020-01-01 at 9.46.33 AM.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4014343/Screen.Recording.2020-01-01.at.9.46.33.AM.zip)

Blender
[Screen Recording 2020-01-01 at 9.45.47 AM.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4014342/Screen.Recording.2020-01-01.at.9.45.47.AM.zip)

**.blend file/ .gltf**

[manne.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4014345/manne.blend.zip)

[Archive.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4014340/Archive.zip)

Has gltf2 validation report with 0 failures and some warnings.

**Version**
 - OS: Mac
 - Blender Version 2.81

**Additional context**
The main difference is three.js works and Blender import doesn't. So this import problem is blocking my use case of doing gltf2 exports to Blender.

scurest:
Very large patch, but it's a massive improvement for all node hierarchy problems, most notably skinning.

Fixes every file I tested ＼（Ｔ∇Ｔ）／

Fixes #239.  
Fixes #257.  
Fixes #483.  
Fixes #509.  
Fixes #698.  
Fixes #699.  
Fixes #700.  
Fixes #703.  
Fixes #847.  
Fixes #848.  
Fixes #850.  
Also fixes the model from #251.

Super big thanks to @fire btw for filing all the issues with test models <3 They were really invaluable, thanks!

Example before/after (#703)

out

## Virtual nodes

The main piece of machinery added is an intermediate graph of "virtual" nodes, or _vnodes_. The glTF node graph is copied to a vnode graph, and a series of passes are done to transform it into a form we can import into Blender. This gives us a lot of flexibility in doing "global" transforms to the node graph before we commit to it by creating real Blender nodes for everything.

Vnodes are stored in a flat dict and referenced by keys just like glTF nodes are stored in a flat array and referenced by indices. Integer keys correspond to vnodes that came from glTF nodes. A root node is always inserted with key `'root'`.

## Summary of the vnode passes

Here's a visualization of how the graph gets transformed from "glTF form" into "Blender form"

flow

1. A dummy root is added to turn the forest into a tree. This simplifies a lot of things. The dummy root doesn't actually get created in Blender though.
2. All nodes used by joints are turned into bones and their common ancestor is marked as an armature. (Assume the colored nodes were used by joints.)
3. The skinned mesh is moved to become a child of the armature that modifies it.
4. The node with both a mesh and camera has the camera split off into a separate node.

## Joints, skins, and armatures

In master, the importer believes that every node is used by at most one joint, and it uses the inverse bind matrix of that joint to set the bone's local-to-parent. This is obviously not viable if there is not a unique joint that uses a node.

This PR changes our attitude about bones and joints. The importer now thinks each bone should have the same world transform it has in glTF (no dependence on the inverse binds). It's also no longer correct to think of skins as imported as armatures; instead a vnode pass marks nodes that should be armatures or bones for the sole purpose of satisfying the following Blender constraint

> The nodes used to skin a mesh must all be bones, and must be proper descendants of a common node that is an armature.

There's no problem with overlapping skins if we think about it this way.

## Skinned meshes

Where in the glTF node graph a skinned mesh is instantiated doesn't affect its transform; only the nodes used by its joints do. So we always move a skinned mesh from wherever it was in glTF to (a fresh child of) the armature that contains its bones. The fresh child has the identity transform, so the skinned mesh has the same world transform as the armature, and then there's a clear path bones-to-arma, then arma-to-world, where the mesh isn't affected by anything else.

We also need to calculate the positions of the vertices in the bind pose ie. the pose the edit bones are in. This might not be the same as the glTF, mainly because glTF nodes can be scaled but edit bones can't. Skinning code for the verts was added to primitive creation. This is the place the inverse bind matrices are now used.

## Animation

Not much to say here, but it was updated to use vnode ids. The bone code has become simpler; there's a straight-forward formula for the TRS of the pose bones, which is nice.

We do have to be cognizant of objects parented to a bone (we need to translate them back to the head of the bone since they're at the tail by default).

## Things I dropped

These are things I stripped out that I'd like to address later

* All scene/collection handling was dropped. The glTF scenes are ignored and the whole node forest is imported into the current Blender scene. This is not actually a big difference from before; in 2.8 everything already went into the current scene.
* We no longer make any attempt to get rid of the Yup2Zup node. It will be much simpler and more foolproof to [convert coordinates whenever we read them out of the glTF file](https://github.com/KhronosGroup/glTF-Blender-IO/issues/715#issuecomment-539011610), which will get rid of it for good.
* The stuff for selecting objects/setting the active object was dropped.
* The stuff for making bone chains was dropped. You can't change the bone length after import because it has significance in the transform hierarchy (it affects objects parented to the bone). It should be possible to add a vnode pass to rotate/resize bones (see #324) after this.


fire:
Thanks for the callout. I was getting frustrated.

What's a good way to test this?

fire:
I tested https://github.com/KhronosGroup/glTF-Blender-IO/issues/850 and it worked. Thank you so much.

💯

julienduroure:
Hello @scurest 
Thanks for this huge PR.
I didn't go deeply into it yet, but here are some general thoughts (meanly in things you currently dropped):

First, I would like to include guys from sketchfab into testing process. @AurL Aurélien, can you please have a look into it, or forward it to the new responsible og glTF in Sketchfab?


> All scene/collection handling was dropped. The glTF scenes are ignored and the whole node forest is imported into the current Blender scene. This is not actually a big difference from before; in 2.8 everything already went into the current scene. 

One possible solution is to import all in current scene, and to create (if needed) some collection(s) to reflect glTF scene. That way, no issue with glTF node in multiple glTF scenes, because all Blender objects will be in same scene, but collections can be a good way to display glTF scenes. This can be done later, not directly linked to this PR.

> We no longer make any attempt to get rid of the Yup2Zup node. It will be much simpler and more foolproof to convert coordinates whenever we read them out of the glTF file, which will get rid of it for good.

Using debug values to get rid of Yup2Zup is part of the production pipeline in Airbus DS. So this has to be restored, we can't drop it. 

> The stuff for selecting objects/setting the active object was dropped.

Same here, Airbus DS need to have the root of imported glTF active at end of import. This need to be restored.

The stuff for making bone chains was dropped. You can't change the bone length after import because it has significance in the transform hierarchy (it affects objects parented to the bone). It should be possible to add a vnode pass to rotate/resize bones (see #324) after this.

Yes, confirmed. We can drop it for now (and we need to think how we can manage it later).

scurest:
Thanks for looking @julienduroure. I know it's huge ><

The active object is easy. Could you clarify what is supposed to be selected though? In master it depends on a lot of different things (like whether there are animations and what the debug_value is).

Should I just select all (and only) the imported objects?

scurest:
I made the first object in the default glTF scene active and selected. Tell me if you wanted something else.

I also pushed the code for getting rid of the Yup2Zup node by hard-applying the conversion whenever we read. You'll notice that the direction bones point changes and correction nodes are inserted for cameras/lights (just like how the exporter works). I can change these in a later PR (I have the code for it written already).

julienduroure:
Regarding active object: 
At end of import, the active object must be the root object just imported. In most cases, this Blender Object is the glTF tree root (but can be different because of armature). This should be simpler to find it in vnode tree. If there are multiple root objects, pick (randomly) one to be active. Regarding scenes : if there is a _scene_ property in glTF file, use this scene. If there is no _scene_ property, use the scenes[0] scene.

Regarding Yup2Zup node : No selection is required at end of import. 

Current rules :
* We don't want to convert Yup to Zup:
  * Because user don't want it (debug = 100)
* We want to convert Yup to Zup, and we will NOT remove it at end of process
  * Because there is some animation on object (check on gltf.animation_object in code), we can't remove it. So the process is to check all imported objected without any parent (most of the time these are root object in glTF tree, but it can be different because of armature), and set all these objects parent to Yup2Zup empty.
* We want to convert Yup to Zup, and this empty will be removed at end of process
  * When there is no animation on object. In that case, We select all root objects (once again, this is most of the time root glTF nodes, but not always), and then apply transforms on all selected objects, and then remove Yup2Zup node

julienduroure:
@scurest Seems we updated this thread in same time :)
I will have a look on your proposal for active/selection and for Yup2Zup removal.
First remark, as I wrote in my previous post that I wrote during your update: No selection is needed at end of import, only active object is really required. So I think you can remove the selection part (we will save some time).

scurest:
We did :) I amended the last commit to set the active object like you said and not select anything.

julienduroure:
Can you please also restore log on debug 101 ?

scurest:
Sure! I have it print the vnode id now.

fire:
@scurest I've added another test case. https://github.com/KhronosGroup/glTF-Blender-IO/issues/862

scurest:
#862 occurs when an armature object has both its loc/rot/scale animated, and also has pose bone animations, and two actions would get created. That didn't ever used to happen since the importer appears to create a new node for the armature that wasn't in the glTF node tree, so it's loc/rot/scale couldn't be animated. In this branch, an existing node is turned into the armature (except when that's not possible and the dummy root is turned into an armature).

So the animation_node code now uses the same cache as the animation_bone code to be sure only one action is created.

Here you can see the difference in the node tree, master/this branch. You can see the extra "6" node in master.

diff

Thanks for catching that @fire :)

fire:
Is this targeted for 2.82?  https://developer.blender.org/project/view/89/

julienduroure:
@fire No, BCon3 is tomorrow, I don't have time to test it heavily. This is a very good improvement (thanks again scurest :) ), but this is quite a big change, so we need time to test it carefully.
So this will go with 2.83

scurest:
@julienduroure How's this coming?

julienduroure:
Tests are in progress.
Target is definitively to merge it as soon as possible, probably in the next two weeks

julienduroure:
Tests that I/we want/need to perform:

- [ ] Airbus DS non regression tests
- [ ] Some armature tests ( Cesium man / monster, ...)
- [ ] Spot light orientation
- [ ] Camera orientation
- [ ] no scene in gltf file tests

scurest:
* I tested CesiumMan, Monster, and Brainstem.

    You might have noticed there's actually [some special handling in `add_primitive_to_bmesh`](https://github.com/KhronosGroup/glTF-Blender-IO/blob/8242f6d523195c936c4d196b829af16675da9fdc/addons/io_scene_gltf2/blender/imp/gltf2_blender_primitive.py#L89-L90) to keep CesiumMan working because it's [skin weights don't sum to 1](https://github.com/KhronosGroup/glTF-Sample-Models/issues/215). Anyway, it looks the same as in master.

    If you check edit mode, Brainstem and Monster are actually both improved by this branch.

    compare-brainstem-editmode
    compare-monster-editmode

* For light orientation, I made a spotlight whose line-of-sight passes through a plane, exported it, and re-imported it. This is affected by whether we're doing Yup2Zup. It works both ways.

    show-light

* For camera orientation, I used the Cameras sample model. This is affected by whether we're doing Yup2Zup. It works both ways.

    show-cameras

* I tested no scenes by just deleting the `scene` and `scenes` property from CesiumMan and loading.

scurest:
Here's an example of how multi-type nodes get resolved.

gltf

The Top node has both a mesh and a camera. In master the camera gets lost.

out

julienduroure:
Thanks a lot @scurest for this huge improvement !

donmccurdy:
These changes look excellent, thank you @scurest!

PR #868, opened 2020-01-09
Minor cleanup of example extension

emackey:
This moves the example extension to a new folder, separates the addon's name from the glTF extension's name, and adds some comments.

/cc @jjcasmar @donmccurdy


example-addons/example_gltf_extension/__init__.py
```diff
@@ -1,22 +1,37 @@
...
 bl_info = {
-    "name": "KDAB_example_extension",
+    "name": "Example glTF Extension",
     "category": "Generic",
-    "blender": (2,80,0),
+    "version": (1, 0, 0),
+    "blender": (2, 80, 0),
+    'location': 'File > Export > glTF 2.0',
+    'description': 'Example addon to add a custom extension to an exported glTF file.',
+    'tracker_url': "https://github.com/KhronosGroup/glTF-Blender-IO/issues/",
```
donmccurdy: Could the 'tracker_url' value, and perhaps 'developer' and 'url' properties below, be more obviously placeholders? Would we expect users to keep 'location' as shown here? I'm not sure what that does...

emackey: Yes, I updated some of these.

I'm hesitant to replace `tracker_url` with a fake value, as it produces a clickable button on the addons panel in Blender.  Instead I put a comment on it.

The `location` field is shown as plain text on the addons panel.  The reason is that Blender users have many addons to choose from, and it may not be obvious to the user what any given addon has done to their UI when they enable or disable it.  This line gives them a hint as to where to look next, to use the addon they just enabled.  In this case, they must look inside the glTF export configuration panel itself, to find the new UI additions.

emackey:
This has been updated for review.

PR #869, opened 2020-01-09
Upgrade test suite to new glTF Validator

emackey:
There were some breaking changes to the list of things reported by the validator.

All roundtrip models have new validation reports.  The `03_skinned_cylinder` model was re-exported from Blender to pick up on skinning bug fixes (it was failing validation prior to roundtrip).

Also update Mocha & ESLint.


emackey:
See #875 instead.

PR #874, opened 2020-01-11
Fix #860 when bevel change normals

### LINKED ISSUE #860: Custom normals from "Apply Modifiers" not used
When using the "Apply Modifiers" checkbox during glTF export, some modifiers (such as the Bevel modifier's "Harden Normals" checkbox) can create custom normals that are not used during export.

It seems likely that the exporter is checking the original Blender object, not the temporary object with all modifiers applied, for the presence of "Custom split normal data", and getting the wrong answer.

Here's a test scene (made in Blender 2.81a).  The selected object has a Bevel modifier that uses hardened normals:

[HardSurfaceTest.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4027920/HardSurfaceTest.blend.zip)

When "Harden Normals" is turned off or not respected, the bevels interact with the triangulation and cause problems:

HardenNormals_bad

But, when "Harden Normals" works, the result is nice:

HardenNormals_good

One workaround is simple, if the user is willing to switch to a "destructive" workflow:  Just manually apply the Bevel modifier prior to export.  Then, the custom normals are generated prior to export, and things work correctly.

But for a non-destructive workflow, a user may wish to keep their original modifier stack (which may include more than just a Bevel) and use the "Apply Modifiers" checkbox in the exporter.

The exporter should detect that the process of applying those modifiers to a temporary object has produced custom split normal data on that temporary object, and export the newly minted custom normals to glTF.

julienduroure:


emackey:
On first the try, this does fix my test model.  I have concerns about this addon needing to keep a list of which modifiers can alter normal vectors.  It seems like it would be far more robust to detect when the normals have been altered, rather than rely on modifier names.  For example in this case, normals are only modified if the "Harden Normals" checkbox is checked in the modifier.

Unfortunately I don't have time this week to dig into this in detail.  But I'm OK merging this as an interim solution.

PR #880, opened 2020-01-19
Add support for use_inherit_rotation and inherit_scale.

### LINKED ISSUE #791: Take use_inherit_scale and use_inherit_rotation into account
I have a character that's animated with some stretch and squash. In order to do so, we used the property use_inherit_scale and use_inherit_rotation. For example, for the idle animation, the character breathes and the torso is scaled on the y and x axis. When exporting to gltf, the entire subtree of bones after this bone scales as well, even if the child bone of the torso bone has a inherit scale disabled.

I'd appreciate if the exporter took into account these properties and reversed location/rotation upon exporting

We have tried to implement the fix but it goes beyond our abilities. Another alternative would be to redo parts of the animations, but that's a long work.

**Additional context**
We tried a fix and we figured out that the place to look at is in exp/gltf2_blender_gather_animation_samplers.py. When a keyframe is calculated, it should be checked for inherit location/rotation and the keyframe shall be saved accordingly.
I attach here a blend file as a sample reproduction case for this issue:

[cubetestgltf.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3882778/cubetestgltf.blend.zip)

Skywolf285:
Should fix exporting of animations that have use_inherit_rotation set to false and/or inherit_scale not set to "FULL".

Not 100% sure it works perfectly under all circumstances at the moment. I have tested it with my own models and some test animations (see attached zip) and these export correctly. But there might still be some edge-cases where it fails.

Tested on Blender 2.81a.

Fixes #791

[TestModels.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4082890/TestModels.zip)

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
@@ -162,21 +162,12 @@ def get_bone_matrix(blender_object_if_armature: typing.Optional[bpy.types.Object
             if bake_bone is None:
                 matrix = pbone.matrix_basis.copy()
             else:
-                if bpy.app.version >= (2, 81, 0):
-                    if (pbone.bone.use_inherit_rotation == False or pbone.bone.inherit_scale != "FULL") and pbone.parent != None:
-                        rest_mat = (pbone.parent.bone.matrix_local.inverted_safe() @ pbone.bone.matrix_local)
-                        matrix = (rest_mat.inverted_safe() @ pbone.parent.matrix.inverted_safe() @ pbone.matrix)
-                    else:
-                        matrix = pbone.matrix
-                        matrix = blender_object_if_armature.convert_space(pose_bone=pbone, matrix=matrix, from_space='POSE', to_space='LOCAL')
-
-                else: #Using Blender 2.80
-                    if (pbone.bone.use_inherit_rotation == False or pbone.bone.use_inherit_scale == False) and pbone.parent != None:
-                        rest_mat = (pbone.parent.bone.matrix_local.inverted_safe() @ pbone.bone.matrix_local)
-                        matrix = (rest_mat.inverted_safe() @ pbone.parent.matrix.inverted_safe() @ pbone.matrix)
-                    else:
-                        matrix = pbone.matrix
-                        matrix = blender_object_if_armature.convert_space(pose_bone=pbone, matrix=matrix, from_space='POSE', to_space='LOCAL')
+                if (pbone.bone.use_inherit_rotation == False or pbone.bone.inherit_scale != "FULL") and pbone.parent != None:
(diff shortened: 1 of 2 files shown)


QbieShay:
Hello @Skywolf285, sorry for not following up for so long. I tested on my own models, it works. Thank you so much.

fire:
@scurest Can you review this?

julienduroure:
@Skywolf285 must rebase it with master first.
I didn't get time to review/test it yet. Don't have so much time right now during the covid-19 crisis. I hope things will get to normal soon

scurest:
@fire One of the maintainers (someone with a "Collaborator" badge) would need to review it.

I'm not familiar with the exporter so I can't say much, but besides rebasing (@Skywolf285 only 2.90 support is needed now), the only other obvious thing is this only fixes animations, not static poses. It appears to work for animations in a quick test though.

Skywolf285:
2.90? I assume you mean 2.80?

Code should be rebased right now (New to this whole git thing. Took me a while to figure it out. Hope I did it right) and removed 2.79 support.

As for static poses: I take we are talking about exporting with _Use Current Frame_ checked? Does this still include armature information? Or are these just exported meshes with modifiers applied? In case of the latter it should use Blender's build-in method for applying modifiers which should handle this use case correctly already.

If it's the former and it's confirmed to not work it shouldn't be too hard of a fix. Would only have to know which file is responsible for getting the bone matrix when exporting static poses. Although I would find it odd if that to uses a different function to get the bone matrix.

scurest:
You rebased correctly :)

No, I did mean 2.90. The bpy.app.version checks have all been removed. You only need to support the latest Blender version now. Just get rid of the `else: #Using Blender 2.80` stuff.

For static poses, I mean like a .blend without animations. I presume the file to change would be `gltf2_blender_gather_joints.py` if you want to do it.

Skywolf285:
Alright, made the changes.

Static poses originally also didn't work with (some?) Bone Constraints. From my limited testing should the changes made also fix this (see attached file). Would be appreciated if people could throw more complex scenarios at it as I'm not entirely sure why `gltf2_blender_gather_joints.py` used `bpy.types.PoseBone.matrix_basis` instead of doing what `gltf2_blender_gather_animation_sampler_keyframes.py` did.

[RunAnimNoAnimData.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4571692/RunAnimNoAnimData.zip)

fire:
Is anyone waiting for feedback?

julienduroure:
I think I am the bootleneck :)
I still need to heavily test it. Lockdown don't let me time to work on it, but I will as soon as possible

fire:
@julienduroure That's great, the stable release is out, so there should be more time now to test.

julienduroure:
I didn't test it on heavy complex files, but all simple & medium complex test files are OK. Merging.

Thanks!

fire:
What is the policy on backporting to LTS Blender?

Issue #882, opened 2020-01-20
Export: proposal for image format options


scurest:
I want to propose changes to the behaviour of the image format options in the exporter

### Automatic (the default)

Basically I want to propose doing this

````py
if export image corresponds to an existing image:
    if existing_image.file_format == 'JPEG':
        file_format = 'JPEG'
    else:
        file_format = 'PNG'
else:
    file_format = 'PNG'
````

IOW, this option would always favor PNGs except that (quoting from #203)

> if the exporter already has access to a source PNG or JPG, it should be just copying the already-compressed data from there directly

This fixes two things:

* Currently this option does not make good use of an already-compressed JPEG. It will only encode an existing image as a JPEG if its _name_ ends with `.jpg`/`.jpeg`. It would be better not to use the name, but to look at the `image.file_format` field.

    This can make a big difference in performance. For Sponza for instance

    * exporting with Automatic, all the JPEGs are re-encoded as PNGs
        * export time: 32s
        * resulting filesize: 116M
    * exporting with JPEG Format (.jpg), all JPEGs just copied over
        * export time: 22s
        * resulting filesize: 45M
* When an image needs to be packed from channels, the packed image can sometimes be a JPEG if one of the constituents is (note: only sometimes!). This is lossy and causes channel-bleed (#818). I think a better default is to prefer being lossless.

### PNG format (.png)

Remove it. Since Automatic would favor PNGs, it's main difference would be that it re-encodes JPEGs as PNG. It would basically become a "make my file bigger" option.

### JPEG format (.jpg)

This is okay, no change.

emackey:
I like this idea.  In particular I think there should never be a case where an existing JPG needlessly becomes a PNG without any channel reassignments.

donmccurdy:
Agreed. 👍

scurest:
Okay, cool. I put up a PR for it.

emackey:
Fixed in #883.

HumanRobot3D:
> When an image needs to be packed from channels, the packed image can sometimes be a JPEG if one of the constituents is (note: only sometimes!). This is lossy and causes channel-bleed (#818). I think a better default is to prefer being lossless.

If all constituents are JPEG I think the resulting packed image should be JPEG too.

https://github.com/KhronosGroup/glTF-Blender-IO/issues/1318

donmccurdy:
More recent discussion in: https://github.com/KhronosGroup/glTF-Blender-IO/pull/1099

PR #883, opened 2020-01-21
Export: implement image options change (#882)

scurest:
Implements #882, see there for details.

This works especially well for re-exporting imported glTF files since their image names don't usually end in `.jpg`/`.jpeg`. For another example of a file helped by avoiding the JPEG->PNG re-encode, DamagedHelmet goes from

* Export Time: 3.2s → 2.2s
* Filesize: 16M → 8M

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
@@ -42,7 +42,7 @@ def gather_image(
         # The export image has no data
         return None
 
-    mime_type = __gather_mime_type(blender_shader_sockets_or_texture_slots, export_settings)
+    mime_type = __gather_mime_type(blender_shader_sockets_or_texture_slots, image_data, export_settings)
     name = __gather_name(blender_shader_sockets_or_texture_slots, export_settings)
 
     uri = __gather_uri(image_data, mime_type, name, export_settings)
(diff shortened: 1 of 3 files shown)


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
```diff
@@ -95,27 +95,24 @@ def __gather_extras(sockets_or_slots, export_settings):
...
-                ".jpg": "image/jpeg",
-                ".png": "image/png",
-            }[extension]
-        return "image/png"
+    if export_settings["gltf_image_format"] == "AUTO":
+        image = export_image.blender_image()
+        if image is not None:
+            if image.file_format == 'JPEG':
+                return "image/jpeg"
+            else:
```
emackey: Maybe these nested `if` statements can be collapsed?

(Also, in other projects, ReSharper always complains if I put `else` after `return`.  The `else` can't do or mean anything after a `return` statement.  Feel free to take or leave this comment).

```
if image is not None and image.file_format == 'JPEG':
    return "image/jpeg"
return "image/png"
```

scurest: Ok, I did that instead.

emackey:
Hi @scurest, we're thinking of giving you commit access to this repository.  Can you email me directly at `elm19087` at gmail dot com?  Thanks!

PR #884, opened 2020-01-22
glTF importer: keep going if invalid custom property seen

garyo:
Some glTFs may have unusual data in "extras" -- the spec places no limits on what goes in there. But Blender doesn't like anything except ints, floats, and simple dicts. Currently, when such an invalid data item is found, the import fails with a TypeError. This patch just prints an error to the console with the name of the invalid property and keeps going.

Filing here per Julien's request in https://developer.blender.org/D6644.


julienduroure:
Do you have example of extras that currently fail?

garyo:
My real example is huge -- I'll try to come up with a small failure case. I didn't trace the problem down to exactly where the TypeError comes from - figured the fix is simple enough.

scurest:
`"extras": { "a": [[1], 2] }` fails.

julienduroure:
Thanks.
I tested with "extras": { "a": [["1"], "2"] }
This case is working

emackey:
@julienduroure This looks like a request to bring all of master into the `2.82` branch.  I don't think this PR was done correctly.

julienduroure:
The PR was done against an old master,  so I first update it to be done against  2.82 branch.

emackey:
Ah OK, thanks.

PR #886, opened 2020-01-23
Import/Perf: pick low-hanging performance fruit for big meshes

scurest:
So I tried importing the model from [T72167: glTF file loads extremely slowly (5+ min)](https://developer.blender.org/T72167). Seems a lot of time is spent in mesh creation, which is pretty slow for big models, but there was some low-hanging fruit that was really easy to pick. This knocks off about 40% of the import time for me.

| Shading option | before | after |
|:---|:---|:---|
| Use Normal Data | 4 min 6 sec | 2 min 30 sec |
| Smooth Shading | 3 min 12 sec | 1 min 38 sec |

The difference is pretty much undetectable in small models.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
@@ -125,19 +125,14 @@ class BlenderMesh():
         if gltf.import_settings['import_shading'] == "NORMALS":
             mesh.create_normals_split()
 
-        # use_smooth for faces
+        use_smooths = []  # whether to smooth for each poly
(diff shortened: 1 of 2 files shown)


scurest:
Also helps MetalRoughSpheres

| Shading option | before | after |
|:---|:---|:---|
| Use Normal Data | 5.1s | 4.0s |
| Smooth Shading | 3.5s | 2.5s |

julienduroure:
Funny fact:
You commented using foreach_set (in #76) one year ago, exaclty (Januray 23rd 2019 / January 23rd 2020)

Thanks @scurest !

scurest:
:D

PR #888, opened 2020-01-24
Import/Perf: create edit bones in batches

scurest:
Previously we would switch to edit mode, create an edit bone, then switch back to object mode for every single bone, which was slow.

````
for each bone:
    switch to edit mode
    create edit bone
    switch to object mode
    do pose bone stuff
````

This switches to edit mode once for the whole armature, creates all bones, and then switches back.

````
for each arma:
    switch to edit mode
    create all edit bones
    switch to object mode
    do pose bone stuff
````

Performance:

| model | master | this branch |
|:---|:---|:---|
| Ruby Rose (#848) | 3.5s | 1.5s |
| Level (#862) | 11.3s | 8.9s |

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_node.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_node.py
@@ -37,9 +37,12 @@ class BlenderNode():
 
         if vnode.type == VNode.Object:
             BlenderNode.create_object(gltf, vnode_id)
+            if vnode.is_arma:
+                BlenderNode.create_bones(gltf, vnode_id)
 
         elif vnode.type == VNode.Bone:
-            BlenderNode.create_bone(gltf, vnode_id)
+            # These are created with their armature
+            pass
 
         elif vnode.type == VNode.DummyRoot:
             # Don't actually create this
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: addons/io_scene_gltf2/blender/imp/gltf2_blender_node.py [hunk 1]
```diff
@@ -119,43 +133,50 @@ def create_bone(gltf, vnode_id):
...
 
         bpy.ops.object.mode_set(mode="OBJECT")
         pose_bone = blender_arma.pose.bones[vnode.blender_bone_name]
 
-        # Put scale on the pose bone (can't go on the edit bone)
-        _, _, s = vnode.trs
-        pose_bone.scale = s
+        for id in bone_ids:
+            # Put scale on pose bone (edit bones have no scale)
+            _, _, s = vnode.trs
```
julienduroure: Seems you forgot to update vnode here.
vnode = gltf.vnodes[id]

scurest: You're right. And the pose_bone too. Thanks!

PR #889, opened 2020-01-24
Import: avoid "husks" left behind when moving skinned meshes

scurest:
Followup to a TODO from #857.

When a skinned mesh is moved to a new child of its arma, the old node where it was may be left with no purpose, becoming an empty "husk". When it won't change anything, rather than create a *new* node, we can move the old node to become a child of the arma instead.

Example using Monster:

husk-explanantion

This is really important for round-tripping: even if the skinned mesh *does* have to get moved to a new child on the *first* import, it will stay in the same place for all subsequent imports. IOW once the importer is happy with how your node graph looks, it won't change it anymore.

Before. On each round-trip the mesh is moved to a new child, so the "husks" build up over time.

before-round

After. Importer is happy with where the mesh is, so it leaves it alone.

after-round

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_vnode.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_vnode.py
@@ -178,12 +178,11 @@ def move_skinned_meshes(gltf):
     joints in its skin affect it.
 
     To do this in Blender:
-     * Move a skinned mesh to become a child of the armature that affects it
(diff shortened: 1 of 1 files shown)


julienduroure:
Seems OK for me. Thanks!
I will merge it soon

scurest:
Actually, one more thing. We can't move it if it's an ancestor of the armature node. I'll add another check for that.

PR #890, opened 2020-01-24
Fix #852 Don't take mute strip into account

### LINKED ISSUE #852: Export only unmuted NLA strips
Greetings,

I would like to propose an enhancement so that only unmuted NLA strips assigned to an object get exported. Currently all of the NLA strips get exported, which is not always the desired behaviour.

A typical .blend file with an animated character can include various tracks, clips and their compositions so the animation clips can be tested how they behave and blend together. At export time however, only specific clips would need to be exported, without all the extras used for testing or showcase purposes.

This functionality is also found in Blender's .fbx exporter and it's essential for a very fine control over exported animations, while keeping a flexible and comfortable workflow. The NLA strips can be muted directly or by muting the whole track. In both cases, anything that is muted is not exported.

exportNLAmuted

julienduroure:


julienduroure:
Basic implementation.
The rule didn't change : only 1 strip must be on track to export it. But we are taken only active strip to count the number of strip on a track
@donmccurdy any thoughts?

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
```diff
@@ -233,9 +233,9 @@ def __get_blender_actions(blender_object: bpy.types.Object,
                 for track in blender_object.data.shape_keys.animation_data.nla_tracks:
                     # Multi-strip tracks do not export correctly yet (they need to be baked),
                     # so skip them for now and only write single-strip tracks.
-                    if track.strips is None or len(track.strips) != 1:
+                    if track.strips is None or len([strip for strip in track.strips if strip.mute is False]) != 1:
                         continue
-                    for strip in track.strips:
+                    for strip in [strip for strip in track.strips if strip.mute is False]:
```
donmccurdy: Maybe we should store the non-muted strips in a new variable at the top of the loop, so we don't have to repeat `if strip.mute is False` twice here? It would be easy to forget on new references.

PR #867, opened 2020-01-08
Import/Perf: speed up animation keyframe creation

scurest:
We currently loop over all keyframes to set their interpolation type. It's a significant speedup to change the options controlling the default keyframe interpolation in the prefs, create the keyframes, then change the prefs back.

It's these options in the preferences, under Animation > F-Curves:

prefs

Gives a pretty great speedup for anything with animations. VC is especially good.

| Model | master | this branch |
|:--|:--|:--|
| [VC](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/VC) | 2.8s | 0.5s |
| [Brainstem](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/BrainStem) | 1.5s | 1.1s |
| Level (#862) | 17s | 13s |

PR #892, opened 2020-01-26
Import: revert #867 (speed up animation keyframe creation)

scurest:
This reverts #867.

It turns out `prefs.keyframe_new_interpolation_type` only affects keyframes created with `keyframe_insert`, not with `keyframe_points.add`. The reason I thought this worked is I forgot to turn off "Always Sample Animations" when I was testing it :(

My bad, guys.


julienduroure:
No problem, I forgot to test it too :-/

PR #902, opened 2020-01-31
Export/UserExts: bugfix for when addons are missing script files

scurest:
Minor bugfix for loading user extensions.

We iterates over all enabled addons and do `sys.modules[addon_name]` to get their module. Sometimes though this raises an error which makes exporting fail.

(The easiest way to get this to happen is to go into the `scripts/addons` folder and rename, let's say `io_scene_obj`, to something else so that Blender can't find it. It will be listed as "Missing script files" in the Addons pane.)

This PR just skips those addons.

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -471,11 +471,15 @@ class ExportGLTF2_Base:
         else:
             preferences = bpy.context.preferences
         for addon_name in preferences.addons.keys():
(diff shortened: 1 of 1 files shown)


scurest:
@julienduroure I PRed this against the `blender-v2.82-release` branch. Was that right?

emackey:
If you merge/rebase it should rerun the tests with the URL fix.

julienduroure:
@scurest yes, it was! Thanks :)

Issue #840, opened 2019-12-19
No way to play imported animations


scurest:
There should be some way to play the animations that were imported. Cf #671.

edit: in case it helps anyone, I am using [this script](https://gist.github.com/scurest/ed3d80213fcf422bf4314eb4f9f514c6) for now

edit: for animations with one track, you can also just star that track. This should work for most character models.

donmccurdy:
Since there's more detail in https://github.com/KhronosGroup/glTF-Blender-IO/issues/671, would you like us to reopen that issue and track remaining tasks there?

scurest:
It's probably better to have a separate issue.

donmccurdy:
That's fine, some of the details in that issue are important. Copying over:

> Is there a way to actually play an animation encoded as a set of identically-named tracks in Blender? Short of manually selecting them all the right actions? If not, could the addon add a widget somewhere that would do that automatically for you?

I'd prefer to avoid modifying Blender's UI through this addon — if that's even allowed? The addon is enabled by default, so from a user's perspective it would just be a new piece of native Blender UI, and I'm sure the Blender team would have some opinions about that. 🙂

A separate addon that can independently manage NLA tracks with similar names could be an option. Or, from [these](https://blender.stackexchange.com/a/752/43930) [suggestions](https://www.oneminutevideotutorials.com/2018/10/30/best-practices-for-creating-new-scenes-in-blender/), the importer could create separate scenes with linked object data — a scene for every animation. That seems very likely to complicate or inflate the export, though. :/

scurest:
You could add an operator without putting in anywhere in the UI I guess (you'd have to call it from the operator search (F3)). The discoverability would be poor, but at least it'd exist.

I don't like the idea of using a scene per animation, but how would it even work? The animation_data is per object, not per scene. The only stuff that's per scene is, like, the markers and the frame start/end and stuff. From your first link AFAICT it looks like each animation would be picked out by one subrange of the timeline instead of our tracks-with-the-same-name thing?

> I'm sure the Blender team would have some opinions about that.

Did the Blender team have any kind of opinion about how we do animations in general? The need to organize multiple actions into one "animation" is definitely not unique to glTF.

julienduroure:
Hello,
I don't like the idea of creating a new scene.
[...]

PR #913, opened 2020-02-04
Import: set scene's frame_{start,end} when restoring animation

scurest:
Minor enhancement for animations. When restoring an animation, set the scene's start/end frame so it will loop at the proper places when you play it in the Timeline.

out

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation.py
@@ -41,21 +41,25 @@ class BlenderAnimation():
             BlenderAnimation.anim(gltf, anim_idx, child)
 
     @staticmethod
-    def restore_animation(gltf, vnode_id, animation_name):
-        """Restores the actions for an animation by its track name on
-        the subtree starting at node_idx."""
-        vnode = gltf.vnodes[vnode_id]
+    def restore_animation(gltf, animation_name):
+        """Restore the actions for an animation by its track name."""
+        frame_end = 0
 
-        obj = None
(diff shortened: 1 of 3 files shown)


julienduroure:
What about if a file has multiple animation with different lengths? Currently, it will set start/end, based on restored animation, but if other animation is more long, it will be "truncated" by default, until the user change start/end by hand.

Note sure what we have to do here:
* keeping as in master?
* Based on restored animation?
* Based on longer animation?

scurest:
Right, to "play" an animation, you have to change the start/end by hand, just like you have to set all the actions by hand. You already have to set the start/end for every animation including the default one. This PR would do it for you for the default one.

No matter what the start/end can only be set correctly for one animation at a time. The others will all either be extended or truncated. So from my point of view the choice is either we set it, in which case we can make it correct for exactly one animation, or we don't, in which case we make it correct for zero animations.

scurest:
If you don't want to change it, it could also just be an enhancement idea for the theoretical feature that will play an NLA strip for you (#840). The script I'm using for that sets the start/end for you.

julienduroure:
I am absolutely not saying I don't want to change it.
I only starting a discussion with all of you to be sure that we are choosing to better solution.
Sorry if you thought it was a remark against your solution. It is definitely not.

scurest:
Ahh, I didn't think that, but feel free to remark against it if you want to lol :) I was also weighing the value of having the animation loop right vs. preserving the existing values of the start/end.

scurest:
Well, since the story of how animations should work hasn't been totally nailed down yet, I'll close this as premature for now. It can get revisited when everyone agrees what the design should be.

PR #914, opened 2020-02-05
Import: manage morph weights at node level

### LINKED ISSUE #312: Manage morph weights at node level
* [ ] Manage weights at node level
* [ ] Be careful of instances of meshes with different weights at node level

scurest:
Fixes #312.  
Also see #290.

This will use the `pynode.weights` property to set a mesh's shapekey weights.

It changes the key for the `pymesh.blender_name` cache to take into account both the skin and the morph weights. To be precise, two instances of the same mesh in glTF will be instances of the same mesh in Blender if

* if they are skinned, they have the same `pynode.skin`
* if their mesh has morph targets, they have the same `pynode.weights`
* if their mesh has morph targets, neither one has its morph weights animated

@julienduroure I think this should satisfy what you said [here](https://github.com/KhronosGroup/glTF-Blender-IO/issues/290#issuecomment-462125634).

----

Here's an [example](https://github.com/KhronosGroup/glTF-Blender-IO/files/4156994/MorphCube.zip):

* all cubes are instances of the same mesh in glTF
* the left-most cube has its `pynode.weights` animated
* the next two cubes both have no `pynode.weights` property
* the last two cubes both have `pynode.weights = [0.5, 0.5]`.

Cubes with the same letter are instances of the same mesh

morphs

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py
@@ -96,20 +96,9 @@ class BlenderGlTF():
             # Something is wrong in file, there is no nodes
             return
 
-        for node_idx, node in enumerate(gltf.data.nodes):
-
+        for node in gltf.data.nodes:
             # Weight animation management
             node.weight_animation = False
-
-            # skin management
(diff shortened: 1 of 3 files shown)


julienduroure:
Thanks @scurest !

Issue #231, opened 2019-01-14
Exporting Alternate Materials


robert-nash:
I know that this is quite a broad workflow issue which is not necessarily specific to this addon but I wonder if the addon might be able to facilitate this.

I am exporting a model from Blender to Babylon.JS as outlined `[here](https://blender.stackexchange.com/questions/128741/exporting-alternate-materials-to-gltf)` and I want to be able to have alternative materials for a mesh such that the user can choose between different materials. At the minute it seems to be only possible to export materials which are assigned to a material and then only from the first material slot. 

I wonder if there is anyway that this addon could export materials which aren't assigned to a mesh in Blender?

maloguertin:
Also interested in this.

donmccurdy:
Is what you're describing here basically that the exporter would include all materials available in the scene when exporting to glTF, even if some of those materials aren't used by any mesh? So they'd be unattached, but engines could access them and use them as needed?

Note that there wouldn't be any sense of exactly which mesh the materials are alternates for. It's just a bag of materials, and the engine can use mesh and material name (and custom code) to assign them properly.

If that's all OK, this seems like a reasonable thing to have available as an option, but probably disabled by default. I think the same logic could apply to textures — users might create a lightmap or other texture that glTF doesn't support, and just want to bundle it, so their runtime can do something with it.

[A more formal proposal](https://github.com/KhronosGroup/glTF/issues/1569) for alternate materials has been made on the glTF repository, but I'm not sure whether that fits well into this Blender exporter, or whether it's stable enough to implement.

maloguertin:
> Is what you're describing here basically that the exporter would include all materials available in the scene when exporting to glTF, even if some of those materials aren't used by any mesh? So they'd be unattached, but engines could access them and use them as needed?
> 
> Note that there wouldn't be any sense of exactly which mesh the materials are alternates for. It's just a bag of materials, and the engine can use mesh and material name (and custom code) to assign them properly.
> 
> If that's all OK, this seems like a reasonable thing to have available as an option, but probably disabled by default. I think the same logic could apply to textures — users might create a lightmap or other texture that glTF doesn't support, and just want to bundle it, so their runtime can do something with it.
> 
> [A more formal proposal](https://github.com/KhronosGroup/glTF/issues/1569) for alternate materials has been made on the glTF repository, but I'm not sure whether that fits well into this Blender exporter, or whether it's near enough completion to implement.

[...]

PR #919, opened 2020-02-10
Enable a gltf_gltf2_hook

jjcasmar:
This patch allows the user to create a gltf2_hook that returns a new list of scenes and animations. The purpose of this is to allow a second pass that can do more things than what currently the exporter does. 

As an example, the exporter checks if an object in bpy.data.objects is exported, and in that case, it checks if it has animations. If it has animations, it exports the animation.

If you want to use glTF2 as an animation library, its not possible, because the animations wont be exported. EXT_property_animation (although not ratified yet), can't be implemented neither as you cant export material animations. 

With a second pass, we allow this use cases and more. 

@scurest @emackey  @julienduroure

### DIFF
--- a/addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
+++ b/addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
@@ -12,11 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-def export_user_extensions(hook_name, export_settings, gltf2_object, *args):
-    if gltf2_object.extensions is None:
-        gltf2_object.extensions = {}
+def export_user_extensions(hook_name, export_settings, *args):
+    if args and hasattr(args[0], "extensions"):
+        if args[0].extensions is None:
+            args[0].extensions = {}
 
     for extension in export_settings['gltf_user_extensions']:
         hook = getattr(extension, hook_name, None)
         if hook is not None:
-            hook(gltf2_object, *args, export_settings)
+            hook(*args, export_settings)
--- a/addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
+++ b/addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
@@ -20,4 +20,8 @@ def export_user_extensions(hook_name, export_settings, *args):
     for extension in export_settings['gltf_user_extensions']:
         hook = getattr(extension, hook_name, None)
         if hook is not None:
-            hook(*args, export_settings)
+            try:
+                hook(*args, export_settings)
+            except Exception as e:
+                print(hook_name, "fails on", extension)
+                print(str(e))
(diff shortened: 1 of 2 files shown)


scurest:
I think you could simplify and generalize it by just adding this to `__gather_gltf`

```diff
 def __gather_gltf(exporter, export_settings):
     active_scene_idx, scenes, animations = gltf2_blender_gather.gather_gltf2(export_settings)
 
+    plan = {'active_scene_idx': active_scene_idx, 'scenes': scenes, 'animations': animations }
+    export_user_extensions('gather_gltf_hook', export_settings, plan)
+    active_scene_idx, scenes, animations = plan['active_scene_idx'], plan['scenes'], plan['animations']
+
     if export_settings['gltf_draco_mesh_compression']:
         gltf2_io_draco_compression_extension.compress_scene_primitives(scenes, export_settings)
         exporter.add_draco_extension()
```

You'd also have to take my suggestion from [here](https://github.com/KhronosGroup/glTF-Blender-IO/pull/732#discussion_r349339259) about making `export_user_extensions` more flexible.

jjcasmar:
The code you have posted would work perfectly without modifying the `export_user_extensions`, no?

addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py
```diff
@@ -12,11 +12,17 @@
...
-        gltf2_object.extensions = {}
+def export_user_extensions(hook_name, export_settings, *args):
+    gltf2_object = args[0]
+    try:
+        object_extensions = getattr(gltf2_object, "extensions")
+    except AttributeError:
+        pass
+    else:
+        if object_extensions is None:
+            setattr(gltf2_object, "extensions", {})
```
scurest: Easier to do

````py
if args and hasattr(args[0], 'extensions'):
    if args[0].extensions is None:
        args[0].extensions = {}
````

Comment (jjcasmar) [reply addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py]: Then you need to do it in all the extensions... By default, the extensions is None, and you dont know which extension is going to be called first

Comment (jjcasmar) [reply addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py]: Ive modify it as you said

**Comment by jjcasmar** [comment]
Any news on this?

File: addons/io_scene_gltf2/io/exp/gltf2_io_user_extensions.py [hunk 2]
```diff
@@ -12,11 +12,15 @@
...
+            args[0].extensions = {}
 
     for extension in export_settings['gltf_user_extensions']:
         hook = getattr(extension, hook_name, None)
         if hook is not None:
-            hook(gltf2_object, *args, export_settings)
+            try:
+                hook(*args, export_settings)
+            except Exception:
+                pass
```
emackey: Should the exception get logged to the console?  Might be hard to debug without that.

emackey:
@jjcasmar Just curious, would this also enable #231, material variants?

@scurest Are your concerns resolved here?  Do you think this is a good idea?

jjcasmar:
@emackey It would be possible, but not nice. The problem is that the exporter creates a glTF.Scene, which contains glTF.Nodes. Each node may contain a mesh, which may contain a material. There is no other way to specify a material in the glTF.Scene. So with this patch you would be able to add a dummy node, with a dummy mesh, with a material, but there is no way to add directly a material. 

This is a general problem created by how the exporter works. The exporter creates an intermediate representation which is not 1:1 a glTF representation (for example, there is no textures root store).

julienduroure:
@scurest @emackey Any concerns here, or could I merge it?

scurest:
I don't really have any problems, but I'm not familiar with this code (my comments were just ways to make stuff shorter).

jjcasmar:
Is there any other review?

emackey:
Seems fine.  Thanks @jjcasmar.

PR #922, opened 2020-02-13
glTF2: Speed up imports with vertex colors using numpy

garyo:
This builds on previous speedups by using numpy to do the linear-to-SRGB color conversion of vertex colors when importing. In my test case, this took 55 sec off the import time, going from 261 sec to 206 sec -- not bad! The change to `gltf2_blender_gltf.py` to optionally enable profiling isn't required, but I found it very useful. Similarly for the commented-out timing info in `gltf2_blender_primitive.py`.

See https://developer.blender.org/T72167#871092 for the original slowness issue.

This also contains a commit 38d8928 to fix loading vertex colors. Without that I get a `KeyError: 'Col'` trying to import. Seems to be related to e4e75ae6.

FYI, the top remaining times in my import are as follows:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      120   49.605    0.413  100.281    0.836 gltf2_blender_primitive.py:40(add_primitive_to_bmesh)
      120   41.151    0.343   48.071    0.401 gltf2_blender_mesh.py:95(bmesh_to_mesh)
 66133440   12.309    0.000   12.309    0.000 {method 'unpack_from' of 'Struct' objects}  
```


addons/io_scene_gltf2/io/com/gltf2_io_color_management.py
```diff
@@ -29,8 +30,53 @@ def color_linear_to_srgb(c):
     Convert from linear to sRGB color space.
 
     Source: Cycles addon implementation, node_color.h.
+    c may be a single color value or an array.
     """
-    if c < 0.0031308:
-        return 0.0 if c < 0.0 else c * 12.92
+    if type(c) in ([], np.ndarray):
```
scurest: Think you meant `list` instead of `[]`.

addons/io_scene_gltf2/io/com/gltf2_io_color_management.py
```diff
@@ -29,8 +30,53 @@ def color_linear_to_srgb(c):
...
+        colors = np.array(c, np.float32)
```
scurest: If `c` is already a numpy array, this creates an unnecessary copy doesn't it?

garyo:
Good point about alpha; I'll fix that. I'll update re-push when the 2.79 support is dropped (I presume that's all the "bpy.app.version < 2.xx" diffs I've seen around.)

addons/io_scene_gltf2/io/com/gltf2_io_color_management.py
garyo: It just returns (a reference to) the original in that case.

addons/io_scene_gltf2/io/com/gltf2_io_color_management.py
garyo: Thanks, will fix.

addons/io_scene_gltf2/io/com/gltf2_io_color_management.py
scurest: Pretty sure it returns a copy. You probably meant to pass `copy=False`.

garyo: You're absolutely right. Will fix.

julienduroure:
@garyo 2.79 support is dropped, you can re-push!

garyo:
OK, force-pushed new version. Thanks!

scurest:
For the alpha-skipping, what about

````py
colors = np.array(c, np.float32) if type(c) == list else c

not_small = colors >= 0.0031308
small_result = np.where(colors<0.0, 0.0, colors * 12.92)
large_result = 1.055 * np.power(colors, 1.0 / 2.4, where=not_small) - 0.055
result = np.where(not_small, large_result, small_result)

if colors.ndim > 1 and colors.shape[-1] == 4:
    result[..., 3] = colors[..., 3]  # don't change alpha channel

return result
````

**Comment by garyo** [comment]
Good idea - I was thinking to save some computation but the way it turned out it didn't save any, so your solution is fine. Coming up...

**Comment by garyo** [comment]
That last is what I was originally trying to express. Hope you like it!

**Comment by julienduroure** [comment]
@scurest @garyo This need a rebase after a recent merge

**Comment by garyo** [comment]
Rebased.

File: addons/io_scene_gltf2/io/com/gltf2_io_color_management.py [hunk 3]
```diff
@@ -29,8 +30,65 @@ def color_linear_to_srgb(c):
...
+    assert a.shape == srgb.shape
+    expected = np.reshape([color_linear_to_srgb(x) for x in a.flatten()], a.shape)
+    np.testing.assert_allclose(srgb, expected)
+
+    a = np.array([[0, 1, 2, 0.1], [2, 3, 4, 0.5]], dtype=np.float32)
+    srgb = color_linear_to_srgb(a)
+    assert a.shape == srgb.shape
+    expected = np.reshape([color_linear_to_srgb(x) for x in a.flatten()], a.shape)
+    expected[:, 3] = a[:, 3] # it shouldn't process alpha
+    np.testing.assert_allclose(srgb, expected)
```
julienduroure: I don't think we need to keep these test function in the final code. Can you please remove them?

addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py
```diff
@@ -23,8 +23,26 @@ def __new__(cls, *args, **kwargs):
         raise RuntimeError("%s should not be instantiated" % cls)
 
     @staticmethod
-    def create(gltf):
-        """Create glTF main method."""
+    def create(gltf, profile=False):
```
julienduroure: We can maybe use debug value to enable the profiling?

addons/io_scene_gltf2/io/com/gltf2_io_color_management.py
garyo: OK -- they are written per `pytest` standard, so if/when the addons get a test suite they should automatically get run as part of that suite. But if you prefer, I can remove them.

addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py
garyo: Happy to do that -- what value do I use?

garyo: Done

addons/io_scene_gltf2/io/com/gltf2_io_color_management.py
garyo: Done

julienduroure:
Thanks a lot :)
This is now merged

julienduroure:
See #996 
Seems there is a performance issue when COLOR accessor is shared between multiple primitives.

PR #931, opened 2020-02-19
Export KHR_materials_clearcoat

donmccurdy:
Based on https://github.com/KhronosGroup/glTF/pull/1756.

To do:
- [x] ~~Further testing.~~
- [x] ~~Attach extension to root material, pending further discussion.~~
- [x] ~~Write clearcoatTexture and clearcoatRoughnessTexture to R and G channels respectively.~~

Example:
- [Archive.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4227358/Archive.zip)
- Screenshot (left without clearcoat, right with clearcoat)

Screen Shot 2020-02-19 at 12 51 20 PM

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
@@ -129,10 +129,18 @@ def __gather_emissive_texture(blender_material, export_settings):
 def __gather_extensions(blender_material, export_settings):
     extensions = {}
 
+    # KHR_materials_unlit
+
     if gltf2_blender_get.get_socket_or_texture_slot(blender_material, "Background") is not None:
         extensions["KHR_materials_unlit"] = Extension("KHR_materials_unlit", {}, False)
 
-    # TODO specular glossiness extension
+    # KHR_materials_clearcoat
+
+    clearcoat_extension = __gather_clearcoat_extension(blender_material, export_settings)
+    if clearcoat_extension:
+        extensions["KHR_materials_clearcoat"] = clearcoat_extension
+
+    # TODO KHR_materials_pbrSpecularGlossiness
 
     return extensions if extensions else None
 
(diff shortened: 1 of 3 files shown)


donmccurdy:
This should be mostly ready, just trying to ensure clearcoatTexture and clearcoatRoughnessTexture get packed correctly at this point.

donmccurdy:
Ok, this seems to be working. My material setup looks like this, is it what we'd expect? 

Screen Shot 2020-03-05 at 10 59 45 PM

Note that clearcoat is packed into <kbd>R</kbd>, and clearcoat roughness into <kbd>G</kbd>. So on the example above, where each is a distinct black/white texture with diagonal stripes in different directions, the baked result looks something like this:

perpendicular stripes in R and G channels, where blue channel in image is solid throughout

emackey:
This is looking really solid, thanks @donmccurdy!  The one bug I've been able to find so far is that normal strength (scale) for the clearcoat normal map is not copied to the glTF.  It should be in:

```
.materials[i].extensions.KHR_materials_clearcoat.clearcoatNormalTexture.scale
```

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
```diff
@@ -217,3 +225,55 @@ def __has_image_node_from_socket(socket):
...
+
+    if len(clearcoat_roughness_slots) > 0:
+        combined_texture = gltf2_blender_gather_texture_info.gather_texture_info(clearcoat_roughness_slots, export_settings)
+        if has_clearcoat_texture:
+            clearcoat_extension['clearcoatTexture'] = combined_texture
+        if has_clearcoat_roughness_texture:
+            clearcoat_extension['clearcoatRoughnessTexture'] = combined_texture
+
+    if __has_image_node_from_socket(clearcoat_normal_socket):
+        clearcoat_extension['clearcoatNormalTexture'] = gltf2_blender_gather_texture_info.gather_texture_info(
```
emackey: Easy fix found!

```suggestion
        clearcoat_extension['clearcoatNormalTexture'] = gltf2_blender_gather_material_normal_texture_info_class.gather_material_normal_texture_info_class(
```

donmccurdy: Nice, thanks!

PR #933, opened 2020-02-20
Removing 2.79

julienduroure:
@donmccurdy @emackey @pjoe @scurest 
I made a PR for removing 2.79

I think some testing code can be refactored now that there is only 2.80+, but let's start with this PR


scurest:
Lgtm.

You probably want to bump [this line](https://github.com/KhronosGroup/glTF-Blender-IO/blob/a22225e887751035378a7c9fbef0fb9da8fef51e/addons/io_scene_gltf2/__init__.py#L19) since the addon will no longer work in 2.81.6.

pjoe:
This would be fantastic - makes things so much simpler :)

The CircleCI build should also be quite a bit faster now, looks like it goes from ~5m to 2m - sweetness indeed  🚤

It may be possible to cut buildtime further, by re-distrubuting the remaining tests across 4 runners (which is what Circle allows for OSS) - but that's an optimization that can be done after this is merged.

+1 (or +1000)

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -28,7 +28,7 @@ If applicable, a zipped folder containing a .blend file for exporter issues and
 
 **Version**
  - OS: [e.g. macOS, linux]
- - Blender Version [e.g. 2.79, 2.80 (2019-03-18)]
+ - Blender Version [e.g. 2.82, 2.83 (2019-03-18)]
```
emackey: Can this say something like `2.82, 2.83, or build date` ?

addons/io_scene_gltf2/__init__.py
```diff
@@ -485,81 +479,9 @@ def execute(self, context):
...
-            col.prop(self, 'export_morph')
-            if self.export_morph:
-                col.prop(self, 'export_morph_normal')
-                if self.export_morph_normal:
-                    col.prop(self, 'export_morph_tangent')
-    else:
-        def draw(self, context):
-            pass
+
+    def draw(self, context):
```
emackey: Is this empty function still needed?  Might be needed by UI code?

julienduroure: Yes, still needed by UI code to get panels working correctly

.github/ISSUE_TEMPLATE/bug_report.md
julienduroure: Done !

tests/pbr_node/README.md
```diff
@@ -1,5 +1,5 @@
 PBR node
 --------
 
-In Blender, link item 'NodeTree/glTF Metallic Roughness' or 'NodeTree/glTF Specular Glossiness' from the glTF2.blend file. gltf2_Principled.blend is for Blender 2.79 and later versions.  
+In Blender, link item 'NodeTree/glTF Metallic Roughness' or 'NodeTree/glTF Specular Glossiness' from the glTF2.blend file. gltf2_Principled.blend is for old Blender 2.79 and for 2.80+ later versions.
```
donmccurdy: I don't think I know what this README.md file is saying anymore, and the line below ("More on this can be found in the user documentation.") doesn't seem to be true.

This is just a test case for the legacy custom node, and we aren't running tests against 2.79 now ... @emackey how should we be explaining this part now?

emackey: Thanks.  I pushed a change to this branch.

As a point of trivia, note the original text said "2.79 and later versions...".  Yes, old Blender-glTF-Exporter was originally developed by UX3D in the Blender 2.78 era, when there was no such thing as a Principled BSDF node.  The original node group here was once compatible with 2.78, although we've never tested this addon's code against 2.78, only the previous exporter.  So the "Principled" version of the node group was actually built using then-new 2.79 and its shiny new Principled BSDF node.

emackey:
New copy tool snuck in at the last minute.  Dare I ask:  Does it handle file deletions?

julienduroure:
Yes, it does :)

donmccurdy:
Wonderful, thank you!

PR #934, opened 2020-02-21
Import: use emission/alpha sockets on Principled node

scurest:
Since <2.81 support was dropped we can assume the Principled node has emission and alpha sockets. Use them.

I tested exporting the test file from #831 and nothing changed.

out

The emission link has to go a long way though doesn't it ￣□￣；


julienduroure:
Thanks!

PR #936, opened 2020-02-21
Import: use Eevee Specular node for pbrSpecGloss materials

scurest:
Since <2.80 support was dropped, use option 3 from #303 (a Eevee Specular node) for spec/gloss materials.

A MixShader + Transparent node is still used for alpha.

spegl-after

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_KHR_materials_pbrSpecularGlossiness.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_KHR_materials_pbrSpecularGlossiness.py
@@ -15,77 +15,60 @@
 import bpy
 from ...io.com.gltf2_io import TextureInfo
 from .gltf2_blender_pbrMetallicRoughness import \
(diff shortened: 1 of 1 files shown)


julienduroure:
What about Cycles? This node is only for Eevee
One (heavy) solution is to have different node setup for Eevee and Cycles, using material output dropdown list

image

scurest:
It just switches the scene to Eevee (see the first few lines in `pbr_specular_glossiness`). Is that okay?

julienduroure:
What I mean is that currently all node trees we created are compatible in both Eevee and Cycles.
This node tree is only compatible with Eevee. Even if you switch to Eevee in code, if user want to use this asset in Cycles, he can't.
I think we need to keep compatibility in both render engine as much as possible. Of course, in this case, the node is only for Eevee, but I think we need to have a minimal fallback for Cycles

scurest:
Okay, that makes sense. Since like you said, adding a separate track of nodes for different Material Outputs would make the graph heavier rather than lighter, I'll just leave this as option 1.

PR #941, opened 2020-02-23
Import: guess original bind pose from inverseBindMatrices

scurest:
Try to guess the original bind pose for bones from the inverse bind matrices. This puts skinned meshes in edit mode back much closer to how they were before the vnode stuff went in.

out

This assumes the inverse bind matrices were computed the obvious way

    (bind matrix for bone b) =
        (bind local for (child of) skeleton root) *
        ... *
        (bind local for grandparent of b) *
        (bind local for parent of b) *
        (bind local for b)

    (inverse bind matrix for bone b) = inverse((bind matrix for bone b))

Any scale is just discarded since edit bones don't have scales. If we can't guess the original bind local for a bone, we fallback to using the value from the default/rest pose.

### Possible problems?

This worked for all the files I tested with (I re-tested with all the models fire posted, CesiumMan, Monster, etc.)

However I'm sort of nervous about how it will work on "weirder" glTF files, so I thought to add an import option (on by default) to enable this. If you don't want to add it though, I can just drop the second commit.

Possible problems:

* The glTF might just not contain enough info to reconstruct the bind local (eg. no skin.skeleton, "gaps" in the joint tree like X -> Y -> Z where X and Z are joints but Y isn't)
* A node might have more than one inverse bind matrix if its used in multiple skins (this will use whichever one comes last)
* inverseBindMatrices might not actually contain the "real" inverse bind matrices computed with the formula above (eg. the inverse binds might be pre-applied so inverseBindMatrices are all 1)
* And sometimes it's apparently better to retarget onto the default pose even if we *can* find the original pose (Brainstem is the only case I found)

brainout

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -870,11 +870,22 @@ class ImportGLTF2(Operator, ImportHelper):
         default="TEMPERANCE",
     )
 
+    guess_original_bind_pose: BoolProperty(
+        name='Guess original bind pose',
+        description=(
+            'Try to guess the original bind pose for skinned meshes from '
+            'the inverse bind matrices.\n'
+            'When off, use default/rest pose as bind pose'
+        ),
+        default=True,
+    )
+
     def draw(self, context):
         layout = self.layout
 
         layout.prop(self, 'import_pack_images')
         layout.prop(self, 'import_shading')
+        layout.prop(self, 'guess_original_bind_pose')
         layout.prop(self, 'bone_heuristic')
 
     def execute(self, context):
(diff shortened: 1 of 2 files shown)


scurest:
What's everybody else thinks about adding the import option for this? I'm kind of undecided (／_＼)

scurest:
I retested with #946 applied and all the bones looked okay except for the model from #848

out

The problem appears to be

>inverseBindMatrices might not actually contain the "real" inverse bind matrices computed with the formula above

because some leaf bones inexplicably have an inverseBindMatrix of 1, which means they all wind up at the origin.

julienduroure:
Well, with option to use the guess or not, I think we can commit

scurest:
Yeah, I also came down on the option being a good idea. And even for the RubyRose model, it's still slightly better than how it was in 2.82.

PR #945, opened 2020-02-25
Import: add mechanism for "local rotations" of vnodes

scurest:
This adds a way to perform a "local" rotation of a vnode, where the world transform of the vnode is rotated

```
(vnode's new world transform) = (vnode's old world transform) (local rotation)
```

without changing the world transform of any of vnode's children.

Because of non-uniform scalings, the local rotation has to be restricted to be a rearrangement of the local axes (eg. something like (X,Y,Z) -> (X,-Z,Y)); more on this below.

This is used for

* rotating cameras/lights without inserting an extra "correction" node (done in this PR)
* rotating bones so they look nicer (will do in a followup PR)

### How it works

We add the local rotation on the right the local TRS for the vnode. Then we add the inverse rotation to the left of the local TRS of its children to cancel it out.

    (vnode's new local transform) = (vnode's old local transform) (local rotation)
    (child's new local transform) = (local rotation inverse) (child's old local transform)

So vnode's world transform gets multiplied by the local rotation, but the world transform of its children doesn't change.

So now every vnode's local transform looks like

```
(local TRS) = (rotation after) (original TRS from glTF file) (rotation before)
            =     Rot[ra]         Trans[t] Rot[r] Scale[s]       Rot[rb]
```

We need to put this in TRS form. To do this I used this fact

> If Rot[rb] is a signed permutation matrix (ie. something like (X,Y,Z) -> (X,-Z,Y) that just rearranges the axes and flips signs), then
>
>     Scale[s] Rot[rb] = Rot[rb] Scale[M s]
>
> where M is the inverse of the corresponding unsigned permutation matrix.

(If we knew s was always uniform we wouldn't need the matrix: we'd have `Scale[s] Rot[rb] = Rot[rb] Scale[s]`.)

Then

```
Rot[ra] Trans[t] Rot[r] Scale[s] Rot[rb] =
Trans[Rot[ra] t] Rot[ra] Rot[r] Scale[s] Rot[rb] =    (use L Trans[t] = Trans[Lt] L)
Trans[Rot[ra] t] Rot[ra] Rot[r] Rot[rb] Scale[M s] =  (use fact above)
Trans[Rot[ra] t] Rot[ra r rb] Scale[M s]              (combine rotations)
```

This is used for adjusting the base TRS values from the glTF file to the final TRS values that should be used in Blender, both in the static pose (`VNode.trs()`) and for animated data (`VNode.base_locs_to_final_locs`, etc.).

### DIFF
--- a/addons/io_scene_gltf2/blender/com/gltf2_blender_math.py
+++ b/addons/io_scene_gltf2/blender/com/gltf2_blender_math.py
@@ -170,3 +170,45 @@ def transform_value(value: Vector, _: Matrix = Matrix.Identity(4)) -> Vector:
 def round_if_near(value: float, target: float) -> float:
     """If value is very close to target, round to target."""
     return value if abs(value - target) > 2.0e-6 else target
+
+def scale_rot_swap_matrix(rot):
+    """Returns a matrix m st. Scale[s] Rot[rot] = Rot[rot] Scale[m s].
+    If rot.to_matrix() is a signed permutation matrix, works for any s.
+    Otherwise works only if s is a uniform scaling.
+    """
+    m = nearby_signed_perm_matrix(rot)  # snap to signed perm matrix
+    m.transpose()  # invert permutation
+    for i in range(3):
+        for j in range(3):
+            m[i][j] = abs(m[i][j])  # discard sign
+    return m
+
+def nearby_signed_perm_matrix(rot):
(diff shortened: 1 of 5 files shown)


julienduroure:
Seems to be OK :)
Tested with lots of "real cases" from sketchfab. I will commit it soon

Issue #879, opened 2020-01-19
Transform +Y Up is not apply to armature bone


k8w:
**Describe the bug**
1. Export glTF, with `+Y Up` enabled
2. The rotation of the bone is not change (as the whole scene)

**Expected behavior**
The bone rotation is transformed by the `+Y Up`.
Just like it in export fbx, there is a setting `Armature/Primary Bone Axis`

**Version**
 - OS: [e.g. macOS, linux] Win10
 - Blender Version [e.g. 2.79, 2.80 (2019-03-18)] 2.81a

julienduroure:
Seems this ticket is quite old, and no more up to date with recent development.
If there are still some issue regarding orientation, please create a new ticket

PR #946, opened 2020-02-25
Import: prettify bones

### LINKED ISSUE #324: Imported armature issues

scurest:
Depends on #945. You can see how the local rotation stuff gets used if you want.

----

Fixes  #324.  
Also see #455.

This attempts to rotate/resize edit bones so they're more readable. It's based on some simple heuristics. There's lots of stuff that you might want to do differently, but I think this is a good start. Anyway all the tools to set lengths and rotate editbones are now in place at least.

An import option was added to control the heuristic for bone directions (it doesn't affect bone lengths). The only two options currently are:

### Blender (+Y)

This rotates bones so their tips lie on their local +Y axis (in glTF space). This option will round-trip bone directions when re-importing models exported from Blender.

(Also see #879)

### Temperance (this is the default)

Name is provisional. It works well for many different models so I made it the default. It tries to place the bone tip on whichever axis is closest to the centroid of its children. Leaf bones are rotated the same way as their parents.

final

Generally it's not too bad for character models at least.

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -835,11 +835,23 @@ class ImportGLTF2(Operator, ImportHelper):
         description="How normals are computed during import",
         default="NORMALS")
 
+    bone_tip_heuristic: EnumProperty(
+        name="Bone Dir",
+        items=(
+            ("BLENDER", "Blender (+Y)",
+                "Round-trips bone directions in glTFs exported from Blender.\n"
+                "Bone tips are placed on their local +Y axis (in glTF space)"),
+        ),
+        description="Heuristic for placing bone tips. Tries to make bones pretty",
+        default="BLENDER",
+    )
+
     def draw(self, context):
         layout = self.layout
 
         layout.prop(self, 'import_pack_images')
         layout.prop(self, 'import_shading')
+        layout.prop(self, 'bone_tip_heuristic')
 
     def execute(self, context):
         return self.import_gltf2(context)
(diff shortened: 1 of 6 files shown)


fire:
Will this make the 2.83 release?

julienduroure:
Yes, plan is to get it in time for 2.83, if everything get smooth

fire:
Do you know what it'll take to have an option that'll force align it to https://github.com/pixiv/three-vrm/blob/v0.3.0/src/vrm/material/shaders/mtoon.frag standards? I'm not sure the exact details. 

It's something like make all bone rotations line up with the global axes.

scurest:
@fire Not sure what you're asking. What should I be looking for in that fragment shader?

fire:
https://github.com/vrm-c/vrm-specification/blob/master/specification/0.0/README.md#vrm-rules 

> No local rotation for bone

My bad, I posted the relevant page. I'm not sure where to put this report, but here sounds like a good place for an option.

julienduroure:
Every tests I did are OK :)
A very big thanks for this PR!
Any reason why this PR is still draft?

scurest:
No problem.

> If I merge this PR, do I need to merge #945 first, or this PR #946 contains all modification from #945 too?

This PR contains all the commits from #945 (branch for #945 is pointing at the second commit here).

> Any reason why this PR is still draft?

Mostly because it depends on #945.

Do you want to change anything about the new option though? Since it'll be part of the UI/Python API. I just changed `bone_tip_heuristic` to `bone_heuristic` since a future heuristic might also affect bone lengths.

Eg. different name than "Temperance"? "strategy" instead of "heuristic"?

fire:
"Temperance Strategy" sounds better. I don't see people use the word heuristic often.

julienduroure:
Well, English is not my primary language, so not sure what is better

scurest:
I'll mark it as ready since #945 looked good. Merging just this should automatically close #945 too.

julienduroure:
Merged :)
Thanks again to @scurest for this big improvement

PR #947, opened 2020-02-25
Fix export of multi-armature models

Peach1:
Fix export failing when .shape_keys does not exist in an Armature's child.data; Since Armatures can parent Armatures, child.data can also be an Armature (which has no .shape_keys attribute)


julienduroure:
Thanks!

Issue #500, opened 2019-05-27
Duplicates of linked collection not exported.


roxlu:
**Describe the bug**
When you create a collection in `A.blend`, then create `B.blend` and link this collection, then create a duplicate of the instance, the duplicated instance is not exported. 

**To Reproduce**
See _describe the bug_

**Expected behavior**
Each instance should be exported as node in GLTF.

**Version**
 - OS: Arch Linux 5.1.4
 - Blender Version: 2.8, 2019-05-25 

**Additional context**
Might be relevant: https://github.com/KhronosGroup/glTF-Blender-IO/issues/484

roxlu:
For anyone who runs into this issue, I found a workaround for the time being. 1) Exprort to FBX, then use 2) https://github.com/facebookincubator/FBX2glTF

julienduroure:
I started to have a look on it.
Some thoughts about it (for myself, to not forgot)

* instances are used for dupli instance and for linking instances
* Need to invalidate the cache to get a new node, to avoid having same node multiple time in scene tree
* Have to check how it will work for armature instances (proxy)
* Check that nodes are not in scene root
* instance offset seems to not work as expected

roxlu:
Thanks @julienduroure awesome you're looking into this. 

I dove into Blender/Python a bit and got a very simplistic exporter working that fits my needs. I'm sure everything in that code is familiar to you, but maybe there are some parts that might be interesting. Especially because it works with linked files/collections/etc. 

https://gist.github.com/roxlu/a3ec74d3e214526c026bdc06adc24117#file-exporter-py-L848

kmturley:
I am getting the same issue.

monkeys.blend contains one Monkey head, with two instances using Blender Collection Instances:
https://docs.blender.org/manual/en/latest/scene_layout/object/properties/instancing/collection.html
<img width="910" alt="Screen Shot 2020-02-28 at 4 33 05 PM" src="https://user-images.githubusercontent.com/1710493/75597034-0d916900-5a48-11ea-94d8-73c7023655ab.png">

Exported directly to gltf format only one shows:
<img width="615" alt="Screen Shot 2020-02-28 at 4 32 03 PM" src="https://user-images.githubusercontent.com/1710493/75597105-5ba66c80-5a48-11ea-8b0f-e10ab44f435b.png">
monkeys-blender.gltf 2KB
monkeys-blender.bin 69KB

meshes:
```
{
    "name" : "Suzanne.004",
[...]

Issue #921, opened 2020-02-13
Cannot export if skeleton is linked


Favkis:
I have linked skeleton from another .blend file

link skeleton
attach mesh to skeleton
export
done, it crashes

should export like normal model with skeleton

error:
https://media.discordapp.net/attachments/506192854066921478/677416435504513024/unknown.png

windows 10
blender 2.81

julienduroure:
Hello,
This is a known issue, duplicate of #500 
I will close this ticket, #500 will remain open until all collection issues are solved (collection instance and linked collection)

julienduroure:
@Favkis This is solved by PR #950.
Target is to merge it soon (and will be available in blender 2.83)

Issue #665, opened 2019-09-08
Translated collection instances are exported without translation


tdaniel22:
**Describe the bug**
Translated collection instances are exported without translation.

**To Reproduce**
1. Create a cube and put it in a new collection.
2. Create an instance of this new collection.
3. Translate, rotate and scale the new instance.
4. Export the scene.

**Expected behavior**
The exported glTF node corresponding to the cube instance should be translated, rotated and scaled.
```
"nodes" : [
    {
        "mesh" : 0,
        "name" : "Cube"
    },
    {
        "children" : [
            0
        ],
        "name" : "Duplication_Offset_CubeInstance",
        "rotation" : [
            0.3826834559440613,
            0,
            0,
            0.9238795042037964
        ],
        "scale" : [
            1.5,
            1.5,
            1.5
        ],
        "translation" : [
            3,
            0,
            0
        ]
    }
]
```

**Observed behavior**
The exported instance node does not contain a translation.
```
"nodes" : [
    {
        "mesh" : 0,
        "name" : "Cube"
    },
    {
        "children" : [
            0
        ],
        "name" : "Duplication_Offset_CubeInstance",
        "rotation" : [
            0.3826834559440613,
            0,
            0,
            0.9238795042037964
        ],
        "scale" : [
            1.5,
            1.5,
            1.5
        ]
    }
]
```

**Version**
 - OS: Linux
 - Blender Version 2.80

**Additional context**
In [`gltf2_blender_gather_nodes.py`](https://github.com/KhronosGroup/glTF-Blender-IO/blob/368c80402a59e2df0b6ec42ef6584c10c86f59ba/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py#L348), instead of using the "normal" translation, collections are translated by the [`instance_offset`](https://docs.blender.org/api/current/bpy.types.Collection.html#bpy.types.Collection.instance_offset) vector. Commenting out those lines fixes the problem. I'm not sure exactly what the `instance_offset` property corresponds to in Blender. Maybe both vectors should be combined?

julienduroure:
Hello,
Thanks for reporting.
This is linked to #500 . Currently, duplication / instance is quite buggy

julienduroure:
@tdaniel22  This is solved by PR #950.
Target is to merge it soon (and will be available in blender 2.83)

PR #950, opened 2020-02-26
Instances

julienduroure:
- [x] Collection instances
- [x] instance offset
- [x] Linked object
- [x] Linked collection
- [x] Weird stuff where 2 objects can have the same name in blender :(
- [ ] If object is animated --> Collection instances are not animated for now
- [x] Can't go to object mode if active object is a linked object when exporting
- [x] armature proxy is detected as collection instance
- [x] Test link / proxy with some driven shapekeys
- [x] manage proxies table for children, not only at root


julienduroure:
Will fix #500 , #665 , #921

julienduroure:
I think people can start to test it.
There is still a TODO (for object inside collection animation), but it is not a common case, so it can be done in another PR

PR #957, opened 2020-03-02
Fix #956 Rename export_selected to use_selected

### LINKED ISSUE #956: Improve consistency of addon's Python API
Reported on Twitter, https://twitter.com/mcclure111/status/1233959819767427073, it seems that other I/O addons have some conventions on parameter naming that we've missed: `export_selected` should be ~~`use_selected`~~ `use_selection` , for one. We should fix that, and check if there are others that could be more consistent.

I think I would start with the FBX addon's naming as a guide, since its options and features are most similar to ours.

julienduroure:


julienduroure:
@donmccurdy @emackey Can you please have a look?

emackey:
I tried loading an old .blend file that I know has export_selected turned on, and it didn't trigger the backwards compatibility code here.  Any ideas?

julienduroure:
Ah, yes, I managed compatibility only for call from command line.
I will think how to manage for old .blend files

donmccurdy:
Looks good to me!

julienduroure:
@emackey This should be OK now for settings that are saved inside .blend

donmccurdy:
Ah sorry, it looks like I wrote the wrong thing on https://github.com/KhronosGroup/glTF-Blender-IO/issues/956. The expected name is `use_selection`, not `use_selected`, per https://docs.blender.org/api/current/bpy.ops.export_scene.html.

julienduroure:
Corrected!

PR #958, opened 2020-03-03
Fix #953 fix animation names

### LINKED ISSUE #953: Export: NLA strip animations sometimes aren't named with their track name
Sometimes (when the NLA strip animation consists of a single action I think), an exported NLA strip animation will be named after an action instead of after the track name.

For example, if you import CesiumMan, you get one NLA strip animation. The track is named "Anim_0" and the action is named "Anim_0_Node_21".

If you then export, you get an animation named "Anim_0_Node_21".

Round-trip a few more times and you wind up with an animation named "Anim_0_Node_21_Node_21_Node_21".

The name should stay "Anim_0" across round-trips.

julienduroure:


julienduroure:

scurest:
Tested. Lgtm.

IIUC the rule for assigning names is now like this, right?

````
if the track has a default name (like NlaTrack)
    if the animation has only one action
        animation name = that action's name
    else
        animation name = None
else
    animation name = track name
````

julienduroure:
if the track has a default name (like NlaTrack)
    if the animation has only one action
        animation name = that action's name
    else
        animation name = None --> Animation name = action name here too
else
    animation name = track name

Issue #909, opened 2020-02-03
Vertex group index error


timsun28:
**Describe the bug**
While trying to export a model from Blender I get the following error: IndexError: bpy_prop_collection[index]: index 16 out of range, size 15
vertex_group_name = blender_vertex_groups[vertex_group_index].name

I have previously succesfuly exported this model as a GLB file, but after messing around with removing certain bones from vertex groupes I get this error and can't export the model anymore. 

If this issue is not solvable without supplying my model, I will be able to send it in private to someone who would want to help me, but I would rather not put it on github. 

Thank you in advance!

**Version**
 - OS: Ubuntu 19.10
 - Blender Version 2.81.16

Full error log: 
Traceback (most recent call last):
  File "/snap/blender/34/2.81/scripts/addons/io_scene_gltf2/__init__.py", line 420, in execute
    return gltf2_blender_export.save(context, export_settings)
  File "/snap/blender/34/2.81/scripts/addons/io_scene_gltf2/blender/exp/gltf2_blender_export.py", line 40, in save
    json, buffer = __export(export_settings)
  File "/snap/blender/34/2.81/scripts/addons/io_scene_gltf2/blender/exp/gltf2_blender_export.py", line 54, in __export
    __gather_gltf(exporter, export_settings)
  File "/snap/blender/34/2.81/scripts/addons/io_scene_gltf2/blender/exp/gltf2_blender_export.py", line 69, in __gather_gltf
    active_scene_idx, scenes, animations = gltf2_blender_gather.gather_gltf2(export_settings)
  File "/snap/blender/34/2.81/scripts/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather.py", line 36, in gather_gltf2
    scenes.append(__gather_scene(blender_scene, export_settings))
  File "/snap/blender/34/2.81/scripts/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_cache.py", line 65, in wrapper_cached
    result = func(*args)
  File "/snap/blender/34/2.81/scripts/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather.py", line 55, in __gather_scene
    node = gltf2_blender_gather_nodes.gather_node(blender_object, blender_scene, export_settings)
  File "/snap/blender/34/2.81/scripts/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py", line 45, in gather_node
    node = __gather_node(blender_object, blender_scene, export_settings)
[...]

Issue #900, opened 2020-01-30
Vertex Group caching


julienduroure:
Currently, vertex group cache is not working properly (as we had before on any bpy.types).
So, we currently exclude any vertex groups when not needed (always, except skinning).
We need to find a way to cache it properly to avoid to set None for vertex groups.

Because of that, Vertex Groups are not propagated to mesh/primitive functions (except when skinning), and this will be needed for LERP on shapekeys when vertex groups restriction on shapekeys. (#747).

This will probably solve (or at least help) on #789

PR #963, opened 2020-03-08
Fix export of vertex groups in skinned model

Peach1:
Export failed when vertex_group_index was out of range

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
@@ -709,6 +709,9 @@ def extract_primitives(glTF, blender_mesh, blender_object, blender_vertex_groups
                     #
 
                     vertex_group_index = group_element.group
+                    
+                    if vertex_group_index < 0 or vertex_group_index >= len(blender_vertex_groups):
+                        continue
                     vertex_group_name = blender_vertex_groups[vertex_group_index].name
 
                     joint_index = None


julienduroure:
I supposed this is a fix for #909 ?
Even for PR, please always include some test cases, or link to ticket that will be solve: it will help me a lot!

Note that a more general subject about that is #900, but this PR can be a good first step

Peach1:
I can confirm that it fixes the same error in #909 
Same Error As 909
I can send a blend test case if needed

Peach1:
Here is a test case, the error occurred when one armature has multiple meshes attached  [export_test_pr963.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4303946/export_test_pr963.blend.zip)

Before the patch, it would not export
After the patch, it exports as expected

GLTF playback tested via: https://gltf-viewer.donmccurdy.com/

julienduroure:
Your analysis is not right, this is not linked to having multiple meshes attached on the same armature. Cf a simplified version of your file in attachement, with only 1 mesh and 1 bone --> still crash.
I prefer not to commit this fix, until we found the real root cause.


[export_vertex_group.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4305276/export_vertex_group.zip)

Peach1:
Thanks for the test case. 
The root issue is when vertex groups reference data that is no longer in the scene.

In the test case, there are vertex groups that refer to deleted geometry
Extra Vertex Groups

This leads to vertex_group_index going out of range and causes the crash `IndexError: bpy_prop_collection[index]: index 44 out of range, size 11`

The patch fixes the crash by filtering out the vertex group indexes that reference data not in range of the current scene.

Issue #613, opened 2019-08-08
Change console output text and style


EtagiBI:
Hello,

At the moment successful gltf/glb import gives a following console output:
```
CRITICAL Data are loaded, start creating Blender stuff
CRITICAL glTF import finished in X.Xs
```
In my view, these lines are quite confusing because the word "CRITICAL" usually indicates critical errors. It becomes even more misleading in Blender Python console because these lines appear there in red-pink color, which is - again - associated with errors. 

Is it possible to replace "CRITICAL" by "SUCCESS" and change the color to blue?

This feature request may seem a bit nit-picky, but when you have to debug a lot of code every single red line or "CRITITCAL" signal draws attention distracting from real problems.

emackey:
What platform are you on?  And what version of Blender?

julienduroure:
This should be solved when we will address #89 , but this is not planned yet (neither at short or medium term).
So I will probably patch importer logging way for avoid this

EtagiBI:
> What platform are you on? And what version of Blender?

I'm on Windows 10 Pro with Blender 2.8 installed.

PR #970, opened 2020-03-11
fix(init): log level critical to info

clementlecorre:
Hi,
I think the critical level is not good.
I'm using a http://sentry.io/ integration with your lib, his errors are triggered.
Thanks


scurest:
Fixes #613.

Since log.info messages don't show up by default, could we `print` here instead?

julienduroure:
Agreed with scurest, log.info don't show up by default, so not the best choice.
My best choice would be to use Formatter to display critical without red color

clementlecorre:
Ok, I replaced logging info with `print()`.

julienduroure:
Maybe not the best logging solution, but merging it as is for now, waiting a better solution

PR #971, opened 2020-03-11
Import: support all texture wrap modes

### LINKED ISSUE #926: Import: unsupported texture wrap modes
Only the texture wrap modes CLAMPxCLAMP and REPEATxREPEAT are supported, because those are the only options for wrapping on a Texture Image node.

TextureSettingsTest.gltf

texture-setting

scurest:
Closes #926. See there for more details.

Not sure what everyone thought about doing this with nodes like this. Also the exporter can't understand this node setup right now obviously.

You can test with TextureSettingsTest from the sample models. [Here's another](https://github.com/KhronosGroup/glTF-Blender-IO/files/4319834/gyroid.zip) test file too.

mirror

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_texture.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_texture.py
@@ -30,13 +30,19 @@ def texture(
 ):
     """Creates nodes for a TextureInfo and hooks up the color/alpha outputs."""
     x, y = location
+    pytexture = mh.gltf.data.textures[tex_info.index]
+    if pytexture.sampler is not None:
+        pysampler = mh.gltf.data.samplers[pytexture.sampler]
+    else:
+        pysampler = Sampler.from_dict({})
+
+    needs_uv_map = False  # whether to create UVMap node
 
     # Image Texture
     tex_img = mh.node_tree.nodes.new('ShaderNodeTexImage')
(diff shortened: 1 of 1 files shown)


emackey:
I'm slightly confused why, in Blender, a "separate xyz" node followed directly by a "combine xyz" node has what appears to be the effect of clamping to the `[0..1]` range.  Is no specific clamping node needed there?

Other than that question about Blender itself, I'm fine with this PR as-is.

scurest:
> I'm slightly confused why, in Blender, a "separate xyz" node followed directly by a "combine xyz" node has what appears to be the effect of clamping to the [0..1] range. Is no specific clamping node needed there?

The Separate/Combine itself isn't doing anything. The code [sets the "real" interpolation on the Image Texture node](https://github.com/KhronosGroup/glTF-Blender-IO/blob/20346d9510c3e19d540931a8050703d97aeaf63a/addons/io_scene_gltf2/blender/imp/gltf2_blender_texture.py#L85) to Extend (ie. CLAMP), so if we pass through the UV unchanged it will be CLAMPed.

Basically there's already a "real" wrapping being done by the Image Texture node. This adds a "fake" wrapping being done by manually with Sep->Math->Com nodes before the "real" wrapping. When the "real" wrapping already does what we want, we don't have to do the "fake" wrapping.

Issue #835, opened 2019-12-16
Exporter is deleting animation data


adevart:
**Describe the bug**
If you animate an object and animate shape keys on the same object, the exporter erases the keyframes on the object.

**To Reproduce**
Start with a cube, set keyframe on say Y translation. Set keys at frame 0 and 50.
Go into the object data properties panel and add shape key (basis key and animation key).
Select the animation shape key and modify the object shape. Then set keyframes on the shape key (e.g value 0 at 0, value 1 at 50).
Save the Blender file.
Export as GLTF with type gltf embedded.
After exporting, saving and reloading the Blender file will show the object's translation keyframes have been deleted by the exporter.
You can also verify this by exporting twice. The first gltf export has both sets of keyframes, the second one only has the shape keys.

**Expected behavior**
Expected outcome is that the exporter doesn't remove animation keys in the Blender scene.

**Version**
 - OS: macOS 10.14
 - Blender Version 2.81

julienduroure:
I cannot reproduce.
Can you please test with latest build (you can download it at https://builder.blender.org/download/ )

adevart:
That's interesting, I tried 2.81a and it doesn't seem to export the translation keys in the first export so two successive exports are the same and only contain the shape key animations (should be both translation and shape keys) but the translation keyframes are removed from Blender for me.

before exporting:
https://ibb.co/mb4rv68
after exporting:
https://ibb.co/YRzzg1m

scurest:
On line 60 here...

https://github.com/KhronosGroup/glTF-Blender-IO/blob/fd2896b1a22bcc61b6835871928e67b0ef9a0618/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py#L59-L64

When you try to assign a shapekey action (`blender_action.id_root == 'KEY'`) to an object's `animation_data.action`, the assignment doesn't go through, so `animation_data.action` doesn't change, buuut its `.users` count is still decremented.

In this case, the CubeAction gets its user count decremented to 0, so it gets erased when doing a save-reload.

PR #989, opened 2020-03-19
Fix #948

### LINKED ISSUE #948: Manage action.id_root = ""
Following #942 and #943 , we need to find a way to change/enhance the way we check animation here :
https://github.com/KhronosGroup/glTF-Blender-IO/blob/19665b05a59550dee1aeaf24d548e2ce8477f4ba/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py#L51

Seems that id_root is always set to "OBJECT" or "KEY" when creating "by hand" in the UI, but sometimes the field is empty (for example when importing fbx file).

What I am proposing:
* Open a bug report for fbx importer
* in glTF exporter: Check the value of id_root. If set --> no change regarding the way it's handle for now in master. If id_root is empty, we need an extra search to be able to determinate if the action is an object/armature animation, or if this is a morph animation.

julienduroure:
Fix #948

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
@@ -53,10 +53,10 @@ def gather_animations(blender_object: bpy.types.Object,
                 break
 
     # Export all collected actions.
-    for blender_action, track_name in blender_actions:
+    for blender_action, track_name, on_type in blender_actions:
 
         # Set action as active, to be able to bake if needed
-        if blender_action.id_root == "OBJECT" and blender_object.animation_data: # Not for shapekeys!
+        if on_type == "OBJECT": # Not for shapekeys!
             if blender_object.animation_data.action is None \
                     or (blender_object.animation_data.action.name != blender_action.name):
                 if blender_object.animation_data.is_property_readonly('action'):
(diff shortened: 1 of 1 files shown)


julienduroure:
@scurest Can you please have a look ?

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
```diff
@@ -53,10 +53,10 @@ def gather_animations(blender_object: bpy.types.Object,
                 break
 
     # Export all collected actions.
-    for blender_action, track_name in blender_actions:
+    for blender_action, track_name, on_type in blender_actions:
 
         # Set action as active, to be able to bake if needed
-        if blender_action.id_root == "OBJECT" and blender_object.animation_data: # Not for shapekeys!
+        if on_type == "OBJECT": # Not for shapekeys!
```
scurest: Why isn't the `and blender_object.animation_data` check needed anymore?

scurest:
This was originally to fix #835 IIRC? I'll retest it...

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
scurest: Okay, I see `__get_blender_actions` can only return `on_type == 'OBJECT'` when it has already checked `blender_object.animation_data`.

PR #995, opened 2020-03-25
Import: add "Fortune" bone heuristic

### LINKED ISSUE #967: Temperance setting does not fully correct for rotation in VRM models
**Describe the bug**
Temperance import Bone Dir setting appears to align all bones to cardinal axes on VRM files, leading to disconnected bones.

This is using the new Bone Dir import feature introduced in the pull request by @scurest and @julienduroure at https://github.com/KhronosGroup/glTF-Blender-IO/pull/946

Temperance is already a huge improvement over the previous behavior, and it seems pretty close to what I want: most bones are aligned correctly. however, all bones appear to be aligned to cardinal axes. For example you can see even thumb bones are pointing straight outward, rather than connected. Hair bones point straight down, even if the hair is angled. Leg bones do not quite connect. On another VRM model by a different author, arm bones are also slightly disconnected.

VRM models are interesting, because they are really the worst-case scenario for import: the spec *requires* them to have bones aligned with the *global* +Y axis on export!

**To Reproduce**
1. Download the attached Alicia model.
2. Blender: Import -> glTF 2.0 -> Select Alicia.vrm / bone dir = Temperance (default)
3. Join meshes, mesh -> viewport display -> display as wire
4. Notice thumb bones and hair bones are disconnected. Leg bones are slightly disconnected at the knee.

**Expected behavior**
I expect bones to appear connected to each other: the parent's tail should be in the same spot as the child's head. Leaf bones can follow the algorithm presented in the original PR.

**Screenshots**
Screen Shot 2020-03-09 at 10 26 06 AM

**.blend file/ .gltf**
[AliciaSolid_vrm-0.51.glb.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4308116/AliciaSolid_vrm-0.51.glb.zip)

Originally available from https://github.com/vrm-c/UniVRM/tree/master/Tests/Models/Alicia_vrm-0.51 , renamed from .vrm to .glb

**Version**
 - OS: [e.g. macOS, linux]
macOS 64 bit
 - Blender Version [e.g. 2.82, 2.83, or build date]
Tested on both 2.83 Alpha [March 09, 16:22:23 - 07bdbeda8462] and 2.81

**Additional context**
It seems to happen in other VRM files too. Maybe I misunderstand what this Bone Dir setting is supposed to do, but the behavior appears very different from the screenshots in the PR linked above.

scurest:
Closes #967. See there for details.

The "Fortune" bone heuristic helps bones look better in files where their local axes are weird, but in exchange may be inaccurate when a bone has a non-uniform scaling.

I also tried to improve the UI text for the bone dir options a bit.

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -859,12 +859,17 @@ class ImportGLTF2(Operator, ImportHelper):
     bone_heuristic: EnumProperty(
         name="Bone Dir",
         items=(
-            ("BLENDER", "Blender (+Y)",
-                "Round-trips bone directions in glTFs exported from Blender.\n"
+            ("BLENDER", "Blender (best for re-importing)",
+                "Good for re-importing glTFs exported from Blender.\n"
                 "Bone tips are placed on their local +Y axis (in glTF space)"),
-            ("TEMPERANCE", "Temperance",
-                "Okay for many different models.\n"
-                "Bone tips are placed at a child's root")
+            ("TEMPERANCE", "Temperance (average)",
+                "Decent all-around strategy.\n"
+                "A bone with one child has its tip placed on the local axis\n"
+                "closest to its child"),
+            ("FORTUNE", "Fortune (may look better, less accurate)",
+                "Might look better than Temperance, but also might have errors.\n"
+                "A bone with one child has its tip placed at its child's root.\n"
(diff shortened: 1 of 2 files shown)


julienduroure:
Thanks a lot @scurest !

scurest:
No problem!

PR #998, opened 2020-03-25
Export: less naive file format detection (fix #986)

### LINKED ISSUE #986: Blender 2.82a glTF 1.2.18 Exports PSD as JPG, there is no PNG option any more
Annotation 2020-03-19 043449

Blender 2.82a with Manually updated glTF Exporter 1.2.18 from Blender 2.83 (1.2.18) as well as the downloaded 1.2.48 from git repo.

Need the exported textures to be in PNG format from my working PSD textures, can't have them being saved to JPG as we need to do arbitrary post export compression of textures to work within our games size limits.

[Bug Blender 2.82a glTF 1.2.18 Exports PSD as JPG, there is no PNG option any more.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4352517/Bug.Blender.2.82a.glTF.1.2.18.Exports.PSD.as.JPG.there.is.no.PNG.option.any.more.zip)

scurest:
Fixes #986 

There are two parts:

### Fix detection of JPEG images

Checking if `image.file_format == 'JPEG'` turns out to be too naive a way to determine if an image is a JPEG. Reading [the docs](https://docs.blender.org/api/current/bpy.types.Image.html) more carefully, I see that `file_format` is the "format used for re-saving this file", not the format it was loaded with, which does not appear to be accessible to us.

So this adds a more defensive check `__is_blender_image_a_jpeg`. See how it looks to you.

### Don't assume an already-encoded image has the "right" format

The importer currently trusts that when it gets already-encoded data for an image with, say `file_format == 'JPEG'`, that data is JPEG data. This is obviously not true (it might be PSD data), so the encoder now does a sanity check with the magic numbers before it okays already-encoded data. If it fails, it encodes the image itself.

---
Testing is appreciated!

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
@@ -103,7 +103,7 @@ def __gather_mime_type(sockets_or_slots, export_image, export_settings):
 
     if export_settings["gltf_image_format"] == "AUTO":
         image = export_image.blender_image()
-        if image is not None and image.file_format == 'JPEG':
+        if image is not None and __is_blender_image_a_jpeg(image):
             return "image/jpeg"
         return "image/png"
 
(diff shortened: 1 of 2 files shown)


scurest:
Btw I partially screwed this up by deleted code for checking extensions in #883. The problem with only checking the extension is the importer uses temp files with no extension when it imports embedded images, and the exporter treated them as PNGs. So this tries to examine the actual file data for a magic number when it can.

scurest:
A simpler alternative would be to do #1005 first and use

```py
def __is_blender_image_a_jpeg(image: bpy.types.Image) -> bool:
    if image.source != 'FILE':
        return False
    path = image.filepath_raw.lower()
    return path.endswith('.jpg') or path.endswith('.jpeg') or path.endswith('.jpe')
```

Maybe that would be enough?

julienduroure:
I just merged #1005 
Agreed to go with your alternative solution!

scurest:
Okay, see how it looks now.

PR #1018, opened 2020-04-08
Export/Perf: use foreach_{get,set} for image encoding

### LINKED ISSUE #932: Image compositor crash / temporary mesh invalidation
This is a strange crash, and I had some trouble creating a minimal reproduce case for it.

To reproduce it, I needed to:

* Merge two textures in such a way that they call @scurest's (awesome) image compositor.
* Turn on "Smooth Shading" and "Auto Normals" for the mesh, such that it will need normal splitting.
* Turn on "Apply Modifiers" such that the normals will get split.

Here's a Blender 2.81a project that will reproduce the problem, but only on newer (2.82) versions of this glTF addon.

[crashy_cube.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4232226/crashy_cube.zip)

Unfortunately, this crash now happens in the stable release of 2.82.  It's new there, it didn't happen in the stable release of Blender 2.81a.

But, the problem can be reproduced in 2.81a, if one manually applies a newer version of this addon to it. That means the problem is localized to the addon, not a change in Blender itself.  From there I was able to run `git bisect` to get an exact ID for which commit introduces the problem.

Git bisect says: 4b460489cdba12bb01954a2748cd7209e6549959 is the first bad commit.

The crash writes this to the console:

```
  File "D:\emackey\Git_Large\glTF-Blender-IO\addons\io_scene_gltf2\__init__.py", line 487, in execute
    return gltf2_blender_export.save(context, export_settings)
  File "D:\emackey\Git_Large\glTF-Blender-IO\addons\io_scene_gltf2\blender\exp\gltf2_blender_export.py", line 40, in save
    json, buffer = __export(export_settings)
  File "D:\emackey\Git_Large\glTF-Blender-IO\addons\io_scene_gltf2\blender\exp\gltf2_blender_export.py", line 53, in __export
    __gather_gltf(exporter, export_settings)
  File "D:\emackey\Git_Large\glTF-Blender-IO\addons\io_scene_gltf2\blender\exp\gltf2_blender_export.py", line 62, in __gather_gltf
    active_scene_idx, scenes, animations = gltf2_blender_gather.gather_gltf2(export_settings)
  File "D:\emackey\Git_Large\glTF-Blender-IO\addons\io_scene_gltf2\blender\exp\gltf2_blender_gather.py", line 37, in gather_gltf2
    scenes.append(__gather_scene(blender_scene, export_settings))
  File "D:\emackey\Git_Large\glTF-Blender-IO\addons\io_scene_gltf2\blender\exp\gltf2_blender_gather_cache.py", line 65, in wrapper_cached
    result = func(*args)
  File "D:\emackey\Git_Large\glTF-Blender-IO\addons\io_scene_gltf2\blender\exp\gltf2_blender_gather.py", line 56, in __gather_scene
    node = gltf2_blender_gather_nodes.gather_node(blender_object, blender_scene, export_settings)
  File "D:\emackey\Git_Large\glTF-Blender-IO\addons\io_scene_gltf2\blender\exp\gltf2_blender_gather_nodes.py", line 46, in gather_node
    node = __gather_node(blender_object, blender_scene, export_settings)
  File "D:\emackey\Git_Large\glTF-Blender-IO\addons\io_scene_gltf2\blender\exp\gltf2_blender_gather_cache.py", line 65, in wrapper_cached
    result = func(*args)
  File "D:\emackey\Git_Large\glTF-Blender-IO\addons\io_scene_gltf2\blender\exp\gltf2_blender_gather_nodes.py", line 64, in __gather_node
    mesh=__gather_mesh(blender_object, export_settings),
  File "D:\emackey\Git_Large\glTF-Blender-IO\addons\io_scene_gltf2\blender\exp\gltf2_blender_gather_nodes.py", line 337, in __gather_mesh
    export_settings)
  File "D:\emackey\Git_Large\glTF-Blender-IO\addons\io_scene_gltf2\blender\exp\gltf2_blender_gather_cache.py", line 65, in wrapper_cached
    result = func(*args)
  File "D:\emackey\Git_Large\glTF-Blender-IO\addons\io_scene_gltf2\blender\exp\gltf2_blender_gather_mesh.py", line 43, in gather_mesh
    weights=__gather_weights(blender_mesh, vertex_groups, modifiers, export_settings)
  File "D:\emackey\Git_Large\glTF-Blender-IO\addons\io_scene_gltf2\blender\exp\gltf2_blender_gather_mesh.py", line 138, in __gather_weights
    if not export_settings[MORPH] or not blender_mesh.shape_keys:
ReferenceError: StructRNA of type Mesh has been removed
```

### LINKED ISSUE #1017: Export/Perf: use foreach_set/foreach_get for manipulating image data
2.83 will include faster methods for getting/setting `image.pixels`: https://developer.blender.org/D7053

We should try these out and see how we can best make use them. If they're fast enough we might even be able to get rid of the compositor hack...

scurest:
Closes #1017.

This uses the new `image.pixels.foreach_{get,set}` functions in Blender 2.83 to avoid going through lists when encoding images with numpy (ie. it now reads/writes directly from a numpy array). This is fast enough that we can delete the compositor code #812 added and just use the numpy code again (hooray!).

Also uses these for copying pixels over from dirty images on the happy path.

This is only slightly faster in my tests, but much simpler code and has no edge cases with bad performance to worry about.

As another bonus, because the compositor code that used to render a temp scene is gone, IMO this also closes #932.

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -16,7 +16,7 @@ bl_info = {
     'name': 'glTF 2.0 format',
     'author': 'Julien Duroure, Norbert Nopper, Urs Hanselmann, Moritz Becher, Benjamin Schmithüsen, Jim Eckerlein, and many external contributors',
     "version": (1, 2, 57),
-    'blender': (2, 82, 7),
+    'blender': (2, 83, 9),
     'location': 'File > Import-Export',
     'description': 'Import-Export as glTF 2.0',
     'warning': '',
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitives.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitives.py
@@ -60,8 +60,6 @@ def gather_primitives(
             material = gltf2_blender_gather_materials.gather_material(blender_material,
                                                                   double_sided,
                                                                   export_settings)
-            # NOTE: gather_material may invalidate blender_mesh (see #932),
-            # so make sure not to access blender_mesh again after this point
         except IndexError:
             # no material at that index
             pass
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
@@ -195,8 +195,13 @@ class ExportImage:
         try:
             tmp_image = image.copy()
             tmp_image.update()
+
             if image.is_dirty:
-                tmp_image.pixels = image.pixels[:]
+                # Copy the pixels to get the changes
+                tmp_buf = np.empty(image.size[0] * image.size[1] * 4, np.float32)
+                image.pixels.foreach_get(tmp_buf)
+                tmp_image.pixels.foreach_set(tmp_buf)
+                tmp_buf = None  # GC this
 
             return _encode_temp_image(tmp_image, self.file_format)
         finally:
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
@@ -14,7 +14,7 @@
 
 import bpy
 import os
-from typing import Optional
+from typing import Optional, Tuple
 import numpy as np
 import tempfile
 import enum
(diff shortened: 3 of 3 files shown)


scurest:
Should I also bump the `blender` version in bl_info to (2, 83, 9)? (That's the first release that has these new functions.)

scurest:
I went ahead and bumped it.

PR #1021, opened 2020-04-11
Export: Add gather_joint_hook for user extensions

netpro2k:
Adds a `gather_joint_hook` allowing user extensions to augment joint nodes on export. There already exists a `gather_skin_hook` but there is not a clean way to get references to the underlying blender objects for each joint in that hook.

Example usage here (via a monkeypatch currently) https://github.com/MozillaReality/hubs-blender-exporter/blob/add-bone-components/io_scene_hubs/__init__.py#L110

Note, for our usage we actually wanted to `Bone` and not the `PoseBone` but I don't really know enough about Blender internals to have an opinion on which is better for the hook to expose, so went with the `PoseBone` since this lines up most closely with what `gather_joint` is already doing.


netpro2k:
@julienduroure I am covered under Mozilla's corporate CLA with Khronos. Is it possible to manually register me with CLAassistant ?

julienduroure:
@donmccurdy Can you please answer or forward the question to someone from Khronos?

donmccurdy:
@netpro2k We are not able to bypass, sorry. Each user that submits a PR must sign the CLA Assistant, even if their member company has already signed the CLA with Khronos.

PR #1022, opened 2020-04-11
Dont fail when using custom property animations

jjcasmar:


jjcasmar:
@scurest @julienduroure @emackey

emackey:
This is targeted to master, but IMHO could be considered a bugfix (even if the old behavior was intentional, it was crashy).  So this could be cherry-picked into 2.83.

Issue #380, opened 2018-07-24
Increased vertex counts when mesh has UVs


miirak:
Okay so I have:

![nevtelen](https://user-images.githubusercontent.com/40399607/43142552-fa79ae38-8f58-11e8-9948-fd1d6e71a596.jpg)

vertices.

After export:
![nevtelen2](https://user-images.githubusercontent.com/40399607/43142580-05bb2ace-8f59-11e8-8cad-ed1afb769e6a.jpg)

28k

scurest:
@donmccurdy Can you check if this is fixed?

julienduroure:
I can confirm that when I disabled UV, normals and tangents, I now get 10500 vertices.
Think we can closed this ticket.
Don't hesitate to comment /  reopen if you still think there is an issue

PR #1023, opened 2020-04-11
Export: rewrite extract_primitives for simplicity/speed

scurest:
Rewrites `extract_primitives` with the goal of simpler code and better performance. IME this is a bit more than twice as fast when exporting big models.

The main idea is not to create the attribute arrays right away. Instead, in the first phase we gather a `vert` tuple for each vertex containing all its attributes⁠—like `(pos, normal, tangent, ...)⁠`—and keep a dict, `verts`, mapping vert tuples to their index.

That means _all_ the code for checking whether a vert already exists or creating it can be replaced by just `verts.setdefault(vert, len(verts))`.

Afterwards a second phase loops over all the verts and puts attributes into attribute arrays.

Random points

* Remove the limit on the number of color sets exported, see https://github.com/KhronosGroup/glTF-Blender-IO/pull/807#issuecomment-565036396.
* Be more attentive to the export options. For example, the old code would split a vertex because of normals even if normals weren't being exported, would loop to find the maximum number of bone influences even if skinning wasn't being exported or the mesh had no skin, etc.
* Avoid doing string manipulation in the tight loops.
* For skins, precompute the joint index for each group index instead of doing a linear scan every time.
* Coordinate changes (convert_swizzle_*) are done in the second phase when filling attribute arrays. This is simpler and also seems to be faster for some reason.
* ~~The tangent.w is computed using the unchanged (before coordinate conversion) normal/tangent/bitangent. Before it was being done with the changed ones. Need to figure out if this is okay.~~
* The morph positions/normals are computed using the unchanged coordinates. Before they were done with changed ones. This should be okay because of linearity.
* Sometimes the exported file is a tiny bit smaller, not sure why.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
@@ -18,39 +18,21 @@
 
 from mathutils import Vector, Quaternion, Matrix
 from mathutils.geometry import tessellate_polygon
-from operator import attrgetter
 
 from . import gltf2_blender_export_keys
 from ...io.com.gltf2_io_debug import print_console
 from ...io.com.gltf2_io_color_management import color_srgb_to_scene_linear
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_skins
 
-#
-# Globals
-#
-
-INDICES_ID = 'indices'
-MATERIAL_ID = 'material'
-ATTRIBUTES_ID = 'attributes'
-
-COLOR_PREFIX = 'COLOR_'
-MORPH_TANGENT_PREFIX = 'MORPH_TANGENT_'
-MORPH_NORMAL_PREFIX = 'MORPH_NORMAL_'
-MORPH_POSITION_PREFIX = 'MORPH_POSITION_'
-TEXCOORD_PREFIX = 'TEXCOORD_'
-WEIGHTS_PREFIX = 'WEIGHTS_'
-JOINTS_PREFIX = 'JOINTS_'
-
-TANGENT_ATTRIBUTE = 'TANGENT'
-NORMAL_ATTRIBUTE = 'NORMAL'
-POSITION_ATTRIBUTE = 'POSITION'
-
-GLTF_MAX_COLORS = 2
-
 
 #
 # Classes
 #
+
+class Prim:
+    def __init__(self):
+        self.verts = {}
+        self.indices = []
 
 class ShapeKey:
     def __init__(self, shape_key, vertex_normals, polygon_normals):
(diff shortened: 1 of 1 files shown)


scurest:
> Sometimes the exported file is a tiny bit smaller, not sure why.

Tracked this down. The reason is in master some verts get split up because of their tangents even when exporting tangents is turned off. If exporting tangents is turned on, the filesizes are the same.

scurest:
> The tangent.w is computed using the unchanged (before coordinate conversion) normal/tangent/bitangent. Before it was being done with the changed ones. Need to figure out if this is okay.

I redid this so it computes tangent.w from the converted normal/tangent/bitangent. Not totally happy with how the bitangent sign works but let's not worry about it now.

scurest:
I think this is ready for review now, when you have time.

Also, anyone have any good test files for this?

julienduroure:
Well, this is quite a big changes. 
What people think about 2.83 or 2.90 target ?
(Note that 2.83 will be a long term support version: see here for more info: https://code.blender.org/2020/02/release-planning-2020-2025/)
@scurest @donmccurdy @emackey

scurest:
I thought 2.90 target for sure. Since 2.83 is in bcon3 now.

But... what exactly does LTS mean for us, like, in practical terms?

emackey:
> what exactly does LTS mean for us, like, in practical terms?

Not sure exactly, but it sounds like Blender will keep a 2.83 LTS branch around for bugfixing for 2 years or so.  I guess we would have a corresponding branch here, and if any major bugs were found in this addon for 2.83, we would fix on that branch.

In practice I think most developer energy in this repo would stay on the master branch, as we're not staffed for long-term fixing of bugs on an LTS branch.  But it doesn't hurt to keep the branch around in case such fixes become needed.

emackey:
And of course, our LTS branch would match this addon in the Blender LTS branch, and have tests hooked up to Blender LTS.  Our master branch would move on with Blender's master, and be tested against their alpha builds.

scurest:
This also fixes #380 in the sense that when export of normals and UVs is disabled, the exported model has the same number of vertices as it does in Blender.

scurest:
I rebased after #1055 merged. This is the added block

https://github.com/KhronosGroup/glTF-Blender-IO/blob/48e6845d8479fa7f12045f3c7aec4464ee39a1a1/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py#L178-L188

vergol:
I can't wait to see this in regular Blender version.
Finally, my exported files do not bloat 10 times.

scurest:
I think it would only give you a significant file size improvement if you're exporting with normals turned off though (maaaybe 3x improvement in that case).

vergol:
I tested on some heavy model, with skinning, normals, vertex colors.
Previous size of .bin : 4,5 MB
The size when exported with this commit: 1,5 MB
In both cases I used normals and smooth shading.
There are problems with resulting file size in current version (happened for me in Blender 2.8, 2.81, 2.82, 2.83), I already saw some issues here on github. Maybe others simply don't care.

scurest:
That's kinda surprising. I wonder if its because of tangents. Are they the same size if you export them both with tangents?

julienduroure:
So ... let's continue the discussion to be sure everything is OK.
 @vergol Did you check the size when you export both with tangents as requested by scurest?

scurest:
@julienduroure ping

PR #1034, opened 2020-04-16
Perf: revert #922

### LINKED ISSUE #996: Import/Perf: question about #922
I want to ask a question about #922. For me it seems to be slower instead of faster.

* When I say `before`, I mean 1900252, the last commit before #922 went in.
* When I say `after`, I mean 354045d, the merge commit for #922.

Since #922, for each primitive we process all the elements in the COLOR_0 array with the linear->sRGB conversion instead of only the ones used by the current primitive. This means we've slowed down in the case where a primitive uses only part of the whole array.

I made [a test file](https://github.com/KhronosGroup/glTF-Blender-IO/files/4382910/sphere.zip) that demonstrates this. It's a sphere broken into six primitives sharing one COLOR_0.

| | before | after |
|:---|:---|:---|
| sphere.gltf | 1.7s | 2.6s |

So I went to test the file from [T72167](https://developer.blender.org/T72167) and found that for me _that_ is actually consistently slower after #922 too. Times are with _Shading > Smooth Shading_. (Note: you'll probably need >4GB of RAM for this import to work)

| | before | after |
|:---|:---|:---|
|modelGroup.glb|112s|115s|

Can someone else test modelGroup.glb with 1900252 and 354045d and see what times you're getting?

scurest:
Reverts #922. Closes #996.

I left the profiling code for debug_value == 102 in though.

This also has the commit replacing `pow` with `**` since it's free.

PRed against the 2.83 branch. I hope that's okay.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_primitive.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_primitive.py
@@ -14,8 +14,6 @@
 
 import bpy
 from mathutils import Vector, Matrix
-import numpy as np
-# import time
 
 from .gltf2_blender_material import BlenderMaterial
 from ...io.imp.gltf2_io_binary import BinaryData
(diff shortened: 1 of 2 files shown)


julienduroure:
> PRed against the 2.83 branch. I hope that's okay.

Perfect! I will merge it during the weekend. Thanks

PR #1035, opened 2020-04-17
Change the way sampled keyframe types are determined.

### LINKED ISSUE #1027: Exporting with baked animation fails with non-standard curve types.
**Describe the bug**
If you animate using special keyframe interpolation (like bounce, quarctic, etc) using the "t" menu on keyframes and then try to export a .gltf file, it doesn't matter if you have "Animation" +  "Always Sample Animations" properties true or not.     If you have non-default key interpolation types on any keyframes, you will get an error complaining about the unknown interpolation type.  

The user shouldn't have to go manually select all the objects with a special key interpolation on them and choose "object > animation > bake action" before exporting.  It should be automatic.  that's why the "always sample animations" checkbox exists no?

This is a bug since when you export with "Always Sample Animations" active, any key interpolation type becomes is irrelevant since all the animation is supposed to get sampled (baked) per frame regardless.   

**To Reproduce**
create a camera.
set some keyframes.
select all keyframes in the dope sheet and hit "t" choose quartic as the key interpolation type.
export a .gltf file with "Selected Objects" false,  "Animation" true, "Always Sample Animations" true.

see error below.

**Expected behavior**
I expect it to bake the animation for all keys, set the interpolation type to linear and save the .glb file without erroring.

**Screenshots**

**Version**
 - OS: windows
 - Blender Version 2.82a and 2.83

y'all are awesome.

emackey:
We already check for "mixed" (different interpolations within one channel) and unknown interpolations.  But there was a bit of bad logic at the end, where we try to discover if `STEP` can be used even after sampling.  This would end up crashing for unknown types, and also didn't fall back to LINEAR even if some LINEAR keys appeared after some CONSTANT ones.

Fixes #1027.

I targeted this to 2.83 release, since it's a bugfix.  Let me know if that's OK.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_samplers.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_samplers.py
@@ -271,19 +271,19 @@ def __gather_interpolation(channels: typing.Tuple[bpy.types.FCurve],
             # If only single keyframe revert to STEP
             if max_keyframes < 2:
                 return 'STEP'
-            else:
-                blender_keyframe = [c for c in channels if c is not None][0].keyframe_points[0]
 
-                # For sampled animations: CONSTANT are STEP, other are LINEAR
-                return {
-                    "BEZIER": "LINEAR",
-                    "LINEAR": "LINEAR",
-                    "CONSTANT": "STEP"
-                }[blender_keyframe.interpolation]
+            # If all keyframes are CONSTANT, we can use STEP.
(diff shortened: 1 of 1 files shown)


julienduroure:
> I targeted this to 2.83 release, since it's a bugfix. Let me know if that's OK.

Confirmed I will merge it on 2.83 branch during the weekend. Thanks!

spiraloid:
well well well would you look at that.  you guys rock.  close this issue up.  wicked!

[p.0009.glb.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4536610/p.0009.glb.zip)

Issue #1040, opened 2020-04-21
master branch now requires Blender 2.90


scurest:
Upstream changes in 5c9f574 (namely the new `heading=` arg to `layout.column()`) require Blender 2.90. This only affects the exporter UI, so automated tests still work in 2.83 for the moment though.

This is a bit involved to change since `blender-28` is hardcoded in a lot of places involving tests

1. ~~update CI to remove hardcoded `28` references~~
2. ~~update CI to download the 2.90 alpha (cf #873)~~
3. ~~update tests/package.json to remove `28` references~~
4. ~~update tests/test/test.js to remove `28` references~~
5. ~~update test instructions in readme that talk about `blender-28`~~
6. ~~update readme to say master will become 2.90~~
7. ~~bump Blender version in bl_info to (2, 90, 0)~~

Not sure how exactly you'll want to do all this. Since 2.79 support is gone and since Blender is going to move to more frequent version bumps, it would be nice if some of this could be "generified" so that maybe only (2), (6), and (7) would be necessary for the next version bump (to 3.0).

cc @emackey

scurest:
Also what's with the Docker files? Looks like they haven't been updated since 2.80.

emackey:
> Also what's with the Docker files? Looks like they haven't been updated since 2.80.

I just bumped them from Node 8 to Node 12.  Since testing happens against downloaded nightly builds, the Docker images got neglected for a long time.

julienduroure:
I did the last 2 points, seems we can close this ticket

scurest:
@julienduroure Thanks. I noticed that the [blender-v2.83-release branch has gotten the `heading=` arg](https://github.com/KhronosGroup/glTF-Blender-IO/blob/blender-v2.83-release/addons/io_scene_gltf2/__init__.py#L563) though. Does it work with the latest 2.83? (Go to Export, then open the Include box.)

blender-v2.83-release has also gotten [the CI script for 2.90](https://github.com/KhronosGroup/glTF-Blender-IO/blob/blender-v2.83-release/.circleci/config.yml).

PR #1041, opened 2020-04-21
Tests: update for Blender 2.90

scurest:
Made an attempt to update the tests for 2.90. Should do 1-5 of #1040. See if this looks okay.

* I renamed hard-coded `blender-28` names to `blender`.
* The tests now call the `blender` command instead of `blender-28`.
* At the next Blender version, it should be enough to bump the new `BLENDER_MAJOR` and `BLENDER_MINOR` variables at the top of the CI script.

### DIFF
--- a/tests/package.json
+++ b/tests/package.json
@@ -12,7 +12,6 @@
   "scripts": {
     "test": "mocha --reporter mochawesome",
     "test-bail": "mocha -b --reporter mochawesome",
-    "test-blender28": "mocha -b --reporter mochawesome -g blender28",
     "lint": "eslint **/*.js",
     "lint:fix": "eslint --fix **/*.js"
   },
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,9 +5,11 @@ jobs:
       - image: node:8-stretch
     parallelism: 2
     environment:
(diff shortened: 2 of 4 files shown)


tests/package.json
```diff
@@ -12,7 +12,7 @@
   "scripts": {
     "test": "mocha --reporter mochawesome",
     "test-bail": "mocha -b --reporter mochawesome",
-    "test-blender28": "mocha -b --reporter mochawesome -g blender28",
+    "test-blender": "mocha -b --reporter mochawesome -g blender",
```
emackey, line 15: I think this line is a holdover of the choice between 2.79 vs 2.80 vs both.  It's effectively the same as `test-bail` now, and can probably be removed.

emackey:
Thanks for doing this @scurest!

tests/package.json
scurest: I can remove it. The only difference is it doesn't run the ["General" tests](https://github.com/KhronosGroup/glTF-Blender-IO/blob/6aa052af7d711267a83779b71cef0254bcb574e2/tests/test/test.js#L207-L216).

scurest: Removed.

PR #1042, opened 2020-04-22
Normalize exported normals

Peach1:
GLTF validators reported non-unit normal vectors on skinned GLTF meshes. Normalization is required after transforming a normal with a matrix


emackey:
@Peach1 Do you have a small sample model that demonstrates this?

I think I ran into what might be the same problem in one of my own models recently, I'll see if I can dig it up too.

Issue #605, opened 2019-08-03
Quaternion not unit


julienduroure:
In https://sandbox.babylonjs.com/,

The following blend file export said that quaternion is not unit
[quaternion_check.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3463451/quaternion_check.blend.zip)
[quaternion_check.glb.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3463452/quaternion_check.glb.zip)

kythyria:
Not just animation. Blender doesn't normalise rotations if the rotation type is set to quaternion in the Item sidebar, so if you export without selecting something else for that object, the unnormalised quaternion will be written.

It's easy to do by accident: just enter an unnormalised quaternion by hand.

PR #1056, opened 2020-05-06
Fix #984 make sure rotation is normalized

### LINKED ISSUE #984: Scaling issue when non normalized quaternion
**Describe the bug**
Mesh scale is weird when exported glb from blender.

**To Reproduce**
Export mesh from blender as gltf
Scaling is wrong(?) on export

**Expected behavior**
Export sucessfully with shape keys as morph targets for threejs

**Screenshots**
image

**.blend file/ .gltf**
[scene (2).zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4350957/scene.2.zip)

**Version**
 - OS: [e.g. macOS, linux]
 - Blender Version 2.82.7

julienduroure:
Fix #984 make sure rotation is normalized


julienduroure:
Seems this will fix #605 too

Issue #1058, opened 2020-05-07
Generate orm texture if separate textures are present


makitsune:
When creating models in Blender, most people use separate occlusion, roughness and metallic textures. This makes it easy to edit each one separately especially when using tools like substance painter or even procedurally generating textures to be baked.

However, when you export to gltf, those 3 textures are ignored and it's a pain to manually convert everything to orm just for gltf. I even made a tool to make this easier: https://makitsune.github.io/orm
Once I've converted all my textures to orm, then I suddenly cant export to any other formats anymore because gltf is the only one that utilizes this format for now.

My proposal is simply, make an orm texture at export! Perhaps find the maximum texture size between each 3 and generate a new texture just for exporting.

I've been dying for this, together with my friends since the release candidate of 2.8 in ~July 2019. It's also been complicated working with freelancers who don't create their own orm textures.

donmccurdy:
From the docs:

> If images are connected to the Blender node in a manner that does not follow this convention, the add-on may attempt to adapt the image to the correct form during exporting (with an increased export time).

We've spend significant time on this and the addon does attempt to combine images, but it's only able to do so much.  In particular, I think that your images will need to be the same size. It should also be possible to use a separate occlusion texture, and to keep it separate, but roughness/metallic must be packed in the final glTF file.

Could you confirm if any part of that behavior is not working as expected for you?

Note that exports merging ORM textures will be significantly slower.

scurest:
So can this be closed?

makitsune:
Oh I never knew a pr was merged for this. I'll take a look today and reply back asap. This is a really awesome addition, thank you so much. It'll make things much less complicated for so many people and artists

makitsune:
I was able to test this and I came across a different problem...

When I don't apply a metallic texture, the metallic value doesn't get converted to a gray scale image.

Here's an example principled shader setup with base color, roughness and normal:

<img width="400" src="https://user-images.githubusercontent.com/8362329/89568889-dc4d6e00-d81b-11ea-8e38-4956b8ae2b86.png"/>

This exports the textures like this:

<img width="400" src="https://user-images.githubusercontent.com/8362329/89569118-29314480-d81c-11ea-88c4-a24de87445b2.png"/>

![image](https://user-images.githubusercontent.com/8362329/89569202-4534e600-d81c-11ea-8150-3a5a2a01e5e5.png)

When decomposing the ORM texture in GIMP, the metallic is 100% white:

[...]

PR #1059, opened 2020-05-07
Export: allow combining different-sized textures (eg for ORM)

scurest:
Thinking about #1058...

The exporter currently wants all textures to be the same size when it combines them (eg for ORM textures). It's actually really easy to lift this restriction.

Test Input:

out

Output Texture with this PR:

out2

----

First commit is just a refactor that adds a TmpImageGuard class that replaces the `try-finally` stuff with a `with`. This lets me abstract out a function for creating a temp copy of an image easier and cleans the logic up a little.

Second commit changes the code for combining the input images. The output dimensions are computed as the max of the input image dimensions. If an input image isn't the same size as the output, it's copied to a temp image, the temp image is scaled with `image.scale`, and that new image is used instead.

And finally the last commit just removes the check in gather_texture_info that wants all input images to be the same size.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
@@ -26,6 +26,7 @@ from io_scene_gltf2.blender.exp import gltf2_blender_gather_materials_pbr_metall
 from ..com.gltf2_blender_extras import generate_extras
 from io_scene_gltf2.blender.exp import gltf2_blender_get
 from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
+from io_scene_gltf2.io.com.gltf2_io_debug import print_console
 
 
 @cached
@@ -192,7 +193,13 @@ def __gather_orm_texture(blender_material, export_settings):
     else:
         result = (occlusion, roughness_socket, metallic_socket)
 
-    # Double-check this will past the filter in texture_info (otherwise there are different resolutions or other problems).
+    if not gltf2_blender_gather_texture_info.check_same_size_images(result):
+        print_console("INFO",
+            "Occlusion and metal-roughness texture will be exported separately "
+            "(use same-sized images if you want them combined)")
+        return None
+
+    # Double-check this will past the filter in texture_info
     info = gltf2_blender_gather_texture_info.gather_texture_info(result, export_settings)
     if info is None:
         return None
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_texture_info.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_texture_info.py
@@ -19,7 +19,6 @@ from io_scene_gltf2.io.com import gltf2_io
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_texture
 from io_scene_gltf2.blender.exp import gltf2_blender_search_node_tree
 from io_scene_gltf2.blender.exp import gltf2_blender_get
-from io_scene_gltf2.io.com.gltf2_io_debug import print_console
 from io_scene_gltf2.io.com.gltf2_io_extensions import Extension
 from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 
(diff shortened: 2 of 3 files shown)


donmccurdy:
How does exporting occlusionTexture work here? If the user has an occlusionTexture that's 4K, and 512px-square roughness and metalness textures, do we export occlusion separately or scale the other two up to merge them?

It's great that the implementation is clean, thanks for that — I am just a bit worried there could be side effects that surprise users (e.g. a file becoming much larger). It's a pity that Blender's addon API doesn't let us show non-crashing warnings to users before exporting, since something like "WARNING: Exporting in this configuration will increase texture sizes." or "WARNING: Textures cannot be exported, do ____ to resolve this." would be really helpful in many cases.

scurest:
> How does exporting occlusionTexture work here?

If they would have been merged if they were the same size, they will be merged now. I don't know how whether to merge gets decided. Is being different sizes the only reason they wouldn't have been merged?

donmccurdy:
I'm not sure. 😕

scurest:
Looks like being different sizes is the only reason they wouldn't get merged. The `orm_texture` parameter that gets passed around [will be None if the ORM texture can't get gathered](https://github.com/KhronosGroup/glTF-Blender-IO/blob/7c5f0d54ea1e44258dd354c61eb3e73dd1a6bc8b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py#L195-L198) and being different sizes looks like the only reason that would happen.

If `orm_texture` is None,  __gather_occlusion_texture and __gather_metallic_roughness_texture will fallback to gathering from just the (occlusion,)/(metal,rough) sockets, so they cope with not being able to combine O and RM. Clearcoat could do this too but it doesn't now.

So yes, O and RM always gets merged with this, whereas before they would be exported separately if they were different sizes. Do you want to do something different?

scurest:
@donmccurdy ?

scurest:
> For example, exporting occlusion as a separate texture if its size differs from metal/rough

If you did want to do this one, it's pretty simple and I believe it would make this PR purely additive (in the sense that no file that already works would change; only some textures that failed to export would begin working).

<details>
<summary>Patch and Example</summary>

````diff
diff --git a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
index 42c7115..f55d944 100644
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
@@ -26,6 +26,7 @@ from io_scene_gltf2.blender.exp import gltf2_blender_gather_materials_pbr_metall
 from ..com.gltf2_blender_extras import generate_extras
 from io_scene_gltf2.blender.exp import gltf2_blender_get
+from io_scene_gltf2.io.com.gltf2_io_debug import print_console
 
 
 @cached
@@ -192,7 +193,13 @@ def __gather_orm_texture(blender_material, export_settings):
     else:
         result = (occlusion, roughness_socket, metallic_socket)
 
-    # Double-check this will past the filter in texture_info (otherwise there are different resolutions or other problems).
+    if not gltf2_blender_gather_texture_info.check_same_size_images(result):
+        print_console("INFO",
+            "Occlusion and metal-roughness texture will be exported separately "
+            "(use same-sized images if you want them combined)")
+        return None
+
+    # Double-check this will past the filter in texture_info
     info = gltf2_blender_gather_texture_info.gather_texture_info(result, export_settings)
     if info is None:
         return None
diff --git a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_texture_info.py b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_texture_info.py
index 928bff4..57b0b5a 100644
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_texture_info.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_texture_info.py
@@ -19,7 +19,6 @@ from io_scene_gltf2.io.com import gltf2_io
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_texture
 from io_scene_gltf2.blender.exp import gltf2_blender_search_node_tree
 from io_scene_gltf2.blender.exp import gltf2_blender_get
-from io_scene_gltf2.io.com.gltf2_io_debug import print_console
 
@@ -128,3 +127,21 @@ def __get_tex_from_socket(socket):
     if result[0].shader_node.image is None:
         return None
     return result[0]
+
+
+def check_same_size_images(
+    blender_shader_sockets: typing.Tuple[bpy.types.NodeSocket],
+) -> bool:
+    """Check that all sockets leads to images of the same size."""
+    if not blender_shader_sockets or not all(blender_shader_sockets):
+        return False
+
+    sizes = set()
+    for socket in blender_shader_sockets:
+        tex = __get_tex_from_socket(socket)
+        if tex is None:
+            return False
+        size = tex.shader_node.image.size
+        sizes.add((size[0], size[1]))
+
+    return len(sizes) == 1
````

out

</details>

emackey:
@scurest Sounds good to me.

scurest:
I added it.

PR #1061, opened 2020-05-07
Import: implement KHR_materials_clearcoat

scurest:
Straightforward implementation of KHR_materials_clearcoat for the importer. See #931 for the exporter.

The image nodes are getting further and further from the PBR node as we add more so its not very pretty though. Oh well.

cc @donmccurdy


donmccurdy:
Thanks, this looks great!

> The image nodes are getting further and further from the PBR node as we add more so its not very pretty though. Oh well.

Note that extensions for several other sockets of Principled BSDF are not too far away... is there some refactoring we should do to allow for this?

scurest:
All the functions that make bits of the node graph take a location parameter. You could make a preliminary pass over the material deciding whether nodes will need to be created and where each bit should go before the "create stuff" pass.

PR #1065, opened 2020-05-09
Import: improve the layout of nodes in the material graph

scurest:
Done on top of #1061, merge that first.

Followup to https://github.com/KhronosGroup/glTF-Blender-IO/pull/1061#issuecomment-626096178. The pieces of the node graph (emission, base color, etc.) were placed at fixed positions before this. This makes a simple pass over the material before we start creating stuff to choose a position for each piece.

The pieces are now (more or less) centered vertically on the Principled node.

Here's an example before/after for a material with only a normal and clearcoatNormal texture

out

I rearranged the blocks so they go in the same order as the sockets on the Principled node. For example, emission (whose socket was at the very top in 2.79) is now near the bottom.

The node with the occlusion hookup is also placed just underneath the Principled node now.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_pbrMetallicRoughness.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_pbrMetallicRoughness.py
@@ -56,57 +56,110 @@ def pbr_metallic_roughness(mh: MaterialHelper):
         make_alpha_socket=False,
     )
 
+    locs = calc_locations(mh)
+
     emission(
         mh,
-        location=(-200, 860),
+        location=locs['emission'],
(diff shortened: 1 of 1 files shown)


scurest:
Another example. Material has an emissive, normal, and occlusion texture.

out

julienduroure:
@scurest Can you please rebase? It will be easier for me to review with only diff from this PR

scurest:
Yeah, no problem.

scurest:
@julienduroure Any update?

julienduroure:
Hello,
I approve the changes. I will merge it soon.

Sorry, not so much time to work on it these days.
Lockdown is not an easy time : Much work, less time. Hope my calendar will go back to normal in following weeks

PR #1080, opened 2020-05-29
Export: All available materials

StidOfficial:
Allow in export settings to export all available materials via `bpy.data.materials` into GLTF for use it  in diffrents variants of mesh materials. #231


addons/io_scene_gltf2/blender/exp/gltf2_blender_export.py
```diff
@@ -75,7 +75,9 @@ def __gather_gltf(exporter, export_settings):
         exporter.add_scene(scene, idx==active_scene_idx)
     for animation in animations:
         exporter.add_animation(animation)
-
+    if export_settings['gltf_all_materials']:
+        for material in materials:
+            exporter.add_material(material)
```
scurest, line 80: Can the `if export_settings['gltf_all_materials']:` check be moved from here...

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather.py
```diff
@@ -32,14 +33,20 @@ def gather_gltf2(export_settings):
...
         scenes.append(__gather_scene(blender_scene, export_settings))
         if export_settings[gltf2_blender_export_keys.ANIMATIONS]:
             animations += __gather_animations(blender_scene, export_settings)
         if bpy.context.scene.name == blender_scene.name:
             active_scene = len(scenes) -1
-    return active_scene, scenes, animations
+
+    for material in bpy.data.materials:
+        double_sided = not material.use_backface_culling
+        materials.append(gltf2_blender_gather_materials.gather_material(material, double_sided, export_settings))
```
scurest, line 47: ...to here? The way it works now,`gather_material` (which can be expensive) is called for all materials even if they won't be used.

scurest:
Also there's a material named "Dots Stroke" in every .blend by default (it's for the grease pencil) that gets exported, which is probably not expected since it doesn't show up in the material select dropdown. Should we check for it somehow in the `for material in bpy.data.materials` loop and skip it?

StidOfficial:
@scurest This commit probably resolv your 2 questions about `if export_settings['gltf_all_materials']:` and the "Dots Stroke" ?

scurest:
Yeah it does!

The rest is just IMO:

The new design, where it exports any material assigned to any slot of any object in an exported scene, seems a bit weird to me. The idea of using unused material slots does seem cool (since it would let you control what materials get exported). Seems more consistent to me to only look for unused slots on meshes that actually get exported though (under an option like Mesh > Unused Materials maybe).

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather.py
```diff
@@ -154,6 +152,21 @@ def __gather_animations(blender_scene, export_settings):
...
     return new_animations
 
+def __gather_materials(blender_scene, export_settings):
+    materials = []
+
+    for _blender_object in blender_scene.objects:
+        blender_object = _blender_object.proxy if _blender_object.proxy else _blender_object
+        for slot in blender_object.material_slots:
+            # Check if this slot as assigned material
+            if slot.material is None:
```
StidOfficial, line 171: @scurest For than i understand, you thing to change `if slot.material is None:` to `if slot.material is None or slot.material == blender_object.active_material:` ?

For exclude active_material in __gather_materials array for better performance ?

scurest: Er, no. I was thinking instead of looking in any material slot in *any object in an exported scene*, only looking in any material slot in *an object that gets exported*.

Is that what the lastest "Check if this object is exclude from view layer" check is supposed to do?

StidOfficial: You think is better to use `bpy.data.objects` directly ?
But i need `blender_scene` with `gltf2_blender_gather_nodes.gather_node`, i can use `users_scene` with Object but it's probably weird code because `users_scene` is tuple.

"Check if this object is exclude from view layer" is because if you uncheck collection, all objects associated with this collection are not exported and is not necessary to exported materials assigned to these objects

scurest: I was thinking something like [this](https://github.com/KhronosGroup/glTF-Blender-IO/compare/9057985...scurest:unused-materials).

StidOfficial: Yes, your codes is more efficient.
But probably change setting name like "Variants materials" for `KHR_materials_variants` ?

donmccurdy:
Let's merge https://github.com/KhronosGroup/glTF-Blender-IO/pull/1206 first — once that's in, it provides a dropdown that seems like a better place to put this option. Also, I'm not sure "All materials" is the right name for this option. I would expect an option with that name to include unused materials. Perhaps "Linked materials"?

julienduroure:
@donmccurdy Linked materials is not a good name. "Linked" (in blender) means that material come from another .blend file
You can have some unused local materials (using Fake User system)

donmccurdy:
Ok — If this were added to the dropdown from #1206, then perhaps:

- **Applied** (default) Export all materials applied to vertices on included objects
- **Slots** Export all materials included in a material slot for an included object
- ...

Or, any other ideas? @StidOfficial would an option to export all materials (whether used in a slot or not) also work for your use case?

StidOfficial:
If Slots option was included Applied + Slots it's good idea with dropdown menu.
I don't see another ideas.

takahirox:
Hi, I'm interested in this feature because `KHR_materials_variants` export support can be based on this feature. 

#1206 has been merged and Applied + Slots options to the dropdown sounds good to me. Can this PR get in if the options are moved to the dropdown menu?

@StidOfficial Would you mind if updating the PR for it? If you are busy, I would help.

StidOfficial:
You can,  i am not currently working on for this moment.

takahirox:
Thanks,  I would like to do. I'm not really sure the right way to take over a PR but I think I locally merge this PR, add some change, and make a new PR.

takahirox:
FYI. Regarding `KHR_materials_variants` extension support, [I started to work on an external `KHR_materials_variants` extension addon](https://github.com/KhronosGroup/glTF-Blender-IO/issues/231#issuecomment-792234727).

julienduroure:
I think we can now close this ticket, as material variants extension is now part of this addon (in Blender 3.3)

PR #1082, opened 2020-06-01
Fix exporting EXTEND textures

bitowl:
I want to export a texture with `Linear` interpolation and `Extend` wrapping. This currently does not create a sampler in the exported file. A sampler is only created when the `Closest` interpolation is used.

This might be a leftover of 34e23a81ab2dced0b81d62615caa2c0a0b6b93d2.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_sampler.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_sampler.py
@@ -39,7 +39,7 @@ def gather_sampler(blender_shader_node: bpy.types.Node, export_settings):
 
 
 def __filter_sampler(blender_shader_node, export_settings):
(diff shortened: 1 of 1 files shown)


julienduroure:
@scurest do you agree with this change
@bitowl Do you have some test case file that demonstrate the change?

scurest:
Yes. It's exactly like bitowl said, I missed updating this 'CLIP' in 34e23a8.

You can test it by creating a texture node that uses Linear/Extend. Exporting and reimporting in master makes Extend turn into Repeat. This branch fixes it.

master1

bitowl:
[quad.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4788885/quad.zip)
Before this change exporting to glTF and importing changes from `Extend` to `Repeat`.
Left correct, Right after export
After this change, this file exports and imports correctly.

PR #1107, opened 2020-06-29
Import: bugfix for occlusion in specgloss

scurest:
Broken by #1065. The signature of make_settings_node changed, but this callsite wasn't updated.


donmccurdy:
Thanks!

PR #1116, opened 2020-07-11
Export/Perf: accept numpy arrays for primitive attribute arrays

scurest:
Accept numpy arrays, in addition to lists, as the arrays for internal primitives.

All lists are converted to numpy arrays to have tobytes called on them, instead of converting them to array.arrays and calling tobytes.

Also for position accessors, the min/max is computed with numpy (eg. [np.amax](https://numpy.org/doc/stable/reference/generated/numpy.amax.html)), which lets us delete gltf2_blender_utils.py.

Also cleans up some code duplication by abstracting out the `array_to_accessor` function.

The main point of this is to allow future enhancements to extract_primitives to use ndarrays for attribute arrays. However, even by itself it already gives a non-trivial perf boost:

* Suzanne subdivided 4 times (126K verts): 16.9s -> 12.9s


donmccurdy:
Looks good here, thanks!

julienduroure:
@Scurest 

Some comment from Blender developer team:

> Hi @Julien DUROURE (julien) it is advised to put imports inside function body to avoid inflating Blender startup time.
Many imports load very fast and have little impact, but numpy takes 0.14 seconds to load on my machine, and that is a significant portion of Blender's 2 second startup time.

scurest:
Can you ask which import statement runs at startup? AFAICT numpy isn't imported until you do a glTF import/export.

The main `__init__.py` file does import the blender/exp modules from inside a function body, eg for export

https://github.com/KhronosGroup/glTF-Blender-IO/blob/98078a0edd5e7bba109737d7495fbc238fd8e86c/addons/io_scene_gltf2/__init__.py#L391-L394

and import

https://github.com/KhronosGroup/glTF-Blender-IO/blob/98078a0edd5e7bba109737d7495fbc238fd8e86c/addons/io_scene_gltf2/__init__.py#L948-L951

If I open up a fresh Blender and check if numpy is imported, it isn't.

````py
import sys
'numpy' in sys.modules  # => False
````

Now do a glTF import/export...

````py
'numpy' in sys.modules # => True
````

scurest:
Also gltf_blender_image.py [already imports numpy exactly the same way](https://github.com/KhronosGroup/glTF-Blender-IO/blob/98078a0edd5e7bba109737d7495fbc238fd8e86c/addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py#L18) so there shouldn't be any change...

julienduroure:
Well, agreed with you.
I asked more info, if you and I missed something (or this was a wrong warning from them)

PR #1117, opened 2020-07-11
Export: use mesh.loop_triangles

### LINKED ISSUE #1091: Triangulate Faces on non square faces with Square UVS breaks the UV's. Related to glTF export
**Describe the bug**
The default export of Quads triangulation is distorting the UVs when the mesh is non square

**To Reproduce**
+ Make a plane
+ scale up one of its edges
+ unwrap all faces with option "Lightmap Pack" or Just "Reset"
+ add an image
+ Triangulate Faces

Or in attached blender file
+ Select the left most mesh that still has quads in edit mode
+ Press Ctrl + t or find the "Triangulate Faces" function via search tool
You'll notice the faces distort the mapping.
+ Select the "Fixed" option in the new options menu in the 3d view to fix

If you export the left most quads mesh to glTF and view in glTF viewer you will see how the AO map looks like the tris version in the screen shots

**Screenshots**
before
before tris
after
after tris
workaround
using fixed option

**Version**
OSX Windows

**Additional context**
If I can patch the addon code I will try. I did search for "quad" and quads_convert_to_tris but found nothing of value I could flip a toggle on

Blender file
[debuggingbox02.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4768127/debuggingbox02.zip)

scurest:
This is faster, less code, and seems to fix #1091.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
@@ -17,7 +17,6 @@
 #
 
 from mathutils import Vector, Quaternion, Matrix
-from mathutils.geometry import tessellate_polygon
 
 from . import gltf2_blender_export_keys
 from ...io.com.gltf2_io_debug import print_console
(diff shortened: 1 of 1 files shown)


scurest:
After merging both this and #1116, another 40% or so is knocked off the export times for 01_vertex_count_32bit.blend (4.3s -> 2.6s).

donmccurdy:
This certainly looks cleaner, but I don't think I understand the difference between the two methods here. Do you know?

scurest:
Not exactly, but from https://developer.blender.org/D3539

> Loop triangles are tessellated triangles create from polygons, for renderers or exporters that need to _match Blender's polygon tesselation exactly_

Emphasis added on the point that explains why this would fix #1091.

PR #1120, opened 2020-07-13
Export/Perf: use numpy in extract_primitives (2-8x speedup)

scurest:
~~Work in progress.~~ This is ready.

Rewrite extract_primitives to use numpy and foreach_get. This gives a pretty good perf boost; export times are now comparable to the other Python exporters (eg. FBX).

Some time comparisons.

| Model | Before | After|
|:---|:---|:---|
|CesiumMan, no anim| 0.37s | 0.09s|
|Ruby Rose, no anim (iss848)| 6.0s | 2.7s|
|01_vertex_count_32bit| 2.6s | 0.5s|
|Suzanne subdivided 4 times| 9.6s | 1.2s|
|AliciaSolid (iss967)| 5.8s | 1.7s|
|folded.blend (iss1112)| 194s | 35s|

### Dots

To replace the dict of verts, the idea is to use [numpy.unique](https://numpy.org/doc/stable/reference/generated/numpy.unique.html). We collect a "dot" for every loop in the mesh. Each dot stores the Blender vertex index (which fully determines all its per-vert data: position, vertex groups, etc), and all the per-loop data at that loop. To make a primitive, we slice out all the dots for the primitive's tris, use numpy.unique to dedupe them, and then read out the attribute arrays from the dots.

### convert_swizzle_*

These aren't used by extract_primitives anymore, so I moved them to gltf2_blender_node where they're still called.

The transformations from eg. __convert_swizzle_normal are now part of the __get_normals function.

### Changes

I think there are only two changes

* the bitangent vector is never stored. Instead, only the bitangent sign is tracked. This should only flip if we apply a transform that reverses orientation (determinant < 0). Since the Zup2Yup doesn't change orientation, the only reason it would flip would be the apply_matrix transform from skinning.
* there is no warning for 0-length tangents since convert_swizzle_tangent is gone. I can put one back in if you want though.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
@@ -16,27 +16,12 @@
 # Imports
 #
 
+import numpy as np
 from mathutils import Vector, Quaternion, Matrix
 
 from . import gltf2_blender_export_keys
 from ...io.com.gltf2_io_debug import print_console
-from ...io.com.gltf2_io_color_management import color_srgb_to_scene_linear
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_skins
-
-
-#
-# Classes
-#
-
-class Prim:
-    def __init__(self):
-        self.verts = {}
-        self.indices = []
-
-class ShapeKey:
-    def __init__(self, shape_key, split_normals):
-        self.shape_key = shape_key
-        self.split_normals = split_normals
 
 
 #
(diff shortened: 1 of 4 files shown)


fire:
Is there a good way to test on 2.83? Can you generate a zip for me?

scurest:
@fire You'll need 2.90 for this. Actually the master branch just switched over to targeting 2.91, but this will still work on 2.90.

scurest:
@julienduroure Think this is ready for review now, when you have time.

scurest:
Tested exporting original.glb from #1166 with default export options. The memory use also improves a lot.

* master
    * time: 118s
    * peak mem: 7.4G
* this branch
    * time: 14s
    * peak mem: 2.5G

julienduroure:
Thanks !

Issue #1069, opened 2020-05-14
T76677 Gltf Imports are splitted into chunks


julienduroure:
https://developer.blender.org/T76677

scurest:
It's expected that meshes get split into chunks with continuous UVs/normals/tangents/colors of course. The obvious question is why the exporter splits the sides into individual quads though, since the normal should be constant on each side

![Sides](https://user-images.githubusercontent.com/11024420/81959965-ddcd1880-95d5-11ea-8caa-5204280d5ad3.png)

The reason is tiny differences in the calculated normals. If the normal is rounded off (say with `round(x, 6)` in `extract_primitives`) you'll get this

![With rounding](https://user-images.githubusercontent.com/11024420/81960137-1a007900-95d6-11ea-9e82-229f45fefbe0.png)

master also splits based on tangents even though tangents aren't exported by default. That's fixed by #1023.

scurest:
Do you think the importer should merge verts together that differ only in their UVs/normals/tangents/colors?

emackey:
> Do you think the importer should merge verts together that differ only in their UVs/normals/tangents/colors?

It wouldn't hurt.  There's been a lot of user confusion over why glTF does it this way, and perhaps the importer could hide all that by automatically merging the vertices.  Done correctly, it wouldn't affect the roundtrip tests, which would split apart again on export.

scurest:
I blocked out a prototype for the importer: [weld-verts-experiment](https://github.com/KhronosGroup/glTF-Blender-IO/compare/master...scurest:weld-verts-experiment).  
It's far enough along to play around with.

Here's an example with CesiumMan. The branch manages to import it all as one piece.

![out](https://user-images.githubusercontent.com/11024420/82514050-40447e00-9ada-11ea-8180-2f4cc1359822.png)

Perf looks between 1.2-2.5x slower so far (only tested with small models).

scurest:
So about the merging verts experiment...

I think it's basically done. I got the import time down to around 0.9-1.5x master's import time (by throwing tons caches at it lol).

### Blender->glTF->Blender

The Blender->glTF->Blender roundtrip works. Originially-distinct-but-collocated verts can get merged obviously.

### glTF->Blender->glTF

But the glTF->Blender->glTF roundtrip is... ehhh. The vertex count can sometimes increase or decrease.

[...]

Issue #1119, opened 2020-07-13
Question: Why not just use mesh.loops[i].normal?


scurest:
There's [a bunch of logic](https://github.com/KhronosGroup/glTF-Blender-IO/blob/blender-v2.83-release/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py#L368-L380) to get the normal for a loop, depending on if the poly is smooth etc etc.

What bad thing happens if we just always call `blender_mesh.calc_normals_split()` beforehand and always read it from `blender_mesh.loops[i].normal`?

Also, same question for tangents.

scurest:
Specifically, what breaks with this patch? All the tests still pass

<details>

<summary>Patch</summary>

````diff
diff --git a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
index 5fe719e..95c446f 100644
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
@@ -144,9 +144,7 @@ def extract_primitives(glTF, blender_mesh, library, blender_object, blender_vert
 
     use_normals = export_settings[gltf2_blender_export_keys.NORMALS]
     if use_normals:
-        if blender_mesh.has_custom_normals:
-            # Custom normals are all (0, 0, 0) until calling calc_normals_split() or calc_tangents().
-            blender_mesh.calc_normals_split()
+        blender_mesh.calc_normals_split()
 
     use_tangents = False
     if use_normals and export_settings[gltf2_blender_export_keys.TANGENTS]:
@@ -242,21 +240,6 @@ def extract_primitives(glTF, blender_mesh, library, blender_object, blender_vert
             prim = Prim()
             prims[material_idx] = prim
 
-        if use_normals:
-            face_normal = None
-            if not (blender_polygon.use_smooth or blender_mesh.use_auto_smooth):
-                # Calc face normal/tangents
-                face_normal = blender_polygon.normal
-                if use_tangents:
-                    face_tangent = Vector((0.0, 0.0, 0.0))
-                    face_bitangent = Vector((0.0, 0.0, 0.0))
-                    for loop_index in blender_polygon.loop_indices:
-                        loop = blender_mesh.loops[loop_index]
-                        face_tangent += loop.tangent
[...]

PR #1121, opened 2020-07-15
Import: rewrite mesh importer with numpy

scurest:
Rewrites the mesh import code to use numpy and foreach_set, as discussed in #996.

The first commit adds accessor decoding to numpy arrays. The existing functions are kept for compat with existing callers.

The second rewrites gltf2_blender_mesh.py. Code is pretty straightforward, just ask if there's anything you want to know about.

This is usually either good for performance, or neutral. Sometimes it is slower, but seems worth it on the balance.

|Model|master|this branch|
|:---|:---|:---|
|modelGroup.glb, Smooth Shading|110s|47s|
|MetalRoughSpheres|5.1|3.4s|
|AliciaSolid (iss967)|1.5s|0.5s|
|RubyRose (iss848)|2.1s|**2.4s**|

This code structure also makes it pretty easy to dedupe verts for #1069; PR to follow.


scurest:
I realized I re-implemented the import shading handling wrong (in master, custom split normals is only set when "Use Normal Data" is on; here, I set it whenever we have normals). Let's block this until #1119 is resolved.

julienduroure:
I converted this PR to draft until the normal issue is solved

scurest:
Nvm lol, I just went ahead and fixed it 'cause it was easy. For some reason I always thought the big difference between "Use Normal Data" and "Flat Shading"/"Smooth Shading" was how it set the poly's use_smooths, but it was actually that "Use Normal Data" creates custom normals and the others don't.

Since "Smooth Shading" now doesn't create split normals it's also even faster! modelGroup.glb is now a cool 35s :o

scurest:
I pushed some more optimizations and RubyRose is no longer slower than master.

julienduroure:
Thanks!

julienduroure:
@Scurest 

Some comment from Blender developer team:

> Hi @Julien DUROURE (julien) it is advised to put imports inside function body to avoid inflating Blender startup time.
Many imports load very fast and have little impact, but numpy takes 0.14 seconds to load on my machine, and that is a significant portion of Blender's 2 second startup time.

PR #1122, opened 2020-07-15
Import: add option to glue pieces of a mesh together

scurest:
Done on top of #1121. (So look at the third commit.)  
Makes a start on #1069.

This adds an option "Glue Pieces Together" (off by default) that serves to partially invert the exporter's splitting of a mesh into pieces with continuous normals/UVs/colors/etc. (As always, tell me if you have a better idea for the option name).

It *cannot* glue along discontinuous normals though, since that would require using per-loop normals which, as I've said before, presents difficulties. That means it works best for models with smooth normals. For example, it will import CesiumMan in one piece. It improves the cube in the OP of #1069 to just 6 pieces, but getting to 1 piece would require gluing discontinuous normals.

(This simplification versus the prototype (not gluing along discontinuous normals) also makes it so that all tests would pass if it were turned on.)

The code impact on the new importer from #1121 is quite small, which makes it easy to flip this on or off with an option. For the future, adding per-loop normal support is a bit more annoying, but still fits into the new code structure without too much trouble.

Performance impact (unlike my last prototype) is quite minimal.

@emackey

out

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -862,6 +862,18 @@ class ImportGLTF2(Operator, ImportHelper):
         default=True
     )
 
+    merge_vertices: BoolProperty(
+        name='Merge Vertices',
+        description=(
+            'The glTF format requires discontinuous normals, UVs, and '
+            'other vertex attributes to be stored as separate vertices, '
+            'as required for rendering on typical graphics hardware.\n'
+            'This option attempts to combine co-located vertices where possible.\n'
+            'Currently cannot combine verts with different normals'
+        ),
+        default=False,
+    )
+
     import_shading: EnumProperty(
         name="Shading",
         items=(("NORMALS", "Use Normal Data", ""),
@@ -906,6 +918,7 @@ class ImportGLTF2(Operator, ImportHelper):
         layout.use_property_decorate = False  # No animation.
 
         layout.prop(self, 'import_pack_images')
+        layout.prop(self, 'merge_vertices')
         layout.prop(self, 'import_shading')
         layout.prop(self, 'guess_original_bind_pose')
         layout.prop(self, 'bone_heuristic')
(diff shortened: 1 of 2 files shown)


emackey:
Awesome @scurest, this seems like the best strategy given Blender's architecture.

Maybe the option could be named "Merge Vertices" instead?  Or "Merge Vertices (where possible)".

donmccurdy:
Just to confirm, the pieces are all part of a single mesh primitive, we're just merging vertices here? If so, "Merge Vertices" sounds good to me, and we can put any caveats in a tooltip.

scurest:
I changed the name.

addons/io_scene_gltf2/__init__.py
```diff
@@ -854,6 +854,18 @@ class ImportGLTF2(Operator, ImportHelper):
...
     )
 
+    merge_vertices: BoolProperty(
+        name='Merge Vertices',
+        description=(
+            'On export to glTF, vertices get split up to create pieces '
+            'with continuous normals/UVs/etc.\n'
+            'When this option is on, attempt to "undo" this and merge '
+            'those verts back together.\n'
+            'Currently cannot merge verts with different normals'
```
emackey, line 864: It's not just Blender export that's the issue.  How about...

> The glTF format requires discontinuous normals, UVs, and other vertex attributes to be stored as separate vertices, as required for rendering on typical graphics hardware. This option attempts to combine co-located vertices where possible.

scurest: Changed.

emackey:
This seems like a good option to me.  👍

scurest:
julienduroure pointed out for #1121 that "Smooth Shading" and "Flat Shading" aren't supposed to set custom normals. I had incorrectly written it so they did. I've pushed the fix.

That means there actually is a way to achieve the ideal result after all, provided you're willing to use "Smooth Shading" or "Flat Shading". Pick "Flat Shading" and "Merge Vertices", and since the custom normals are being ignored, it will happily merge verts that had different ones.

ter

julienduroure:
@scurest Can you please rebase?

scurest:
Yeah, rebased.

Issue #193, opened 2018-12-26
Selected object doesn't get exported when its parent is not selected


soadzoor:
I work on a scene, and I wanted to export only a specific mesh from it for testing, so I selected the option `selected objects` in the exporter settings. I noticed that it produced a "corrupt" glb file, which is too small to be valid. The following errors are shown in the validator:
![capture3](https://user-images.githubusercontent.com/10392261/50454916-dc507100-094a-11e9-96b3-0e83cd11bc40.PNG)

If I select all of the object's ancestors as well, it gets exported flawlessly. Here's the test scene, if anyone wants to try:
[test.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/2710757/test.zip)

emackey:
I've seen this too.  And worse, if you place an object inside a collection, and use "only selected", the selected object won't export because it has a parent, namely the collection itself.

AlexMuir:
I looked into this. At the moment, the exporter starts with root objects and iterates over them filtering out the stuff that doesn't match selection filters.  So if a root object isn't selected, then none of its children will ever be evaluated even if they themselves are selected.

This is also related to my 'Export active collection' issue (https://github.com/KhronosGroup/glTF-Blender-IO/issues/192)  which I had to implement in a hacky way.

Basically I think the code needs rearchitecting so that it doesn't start with root nodes and filter out the stuff that's not needed. 

- If 'Selected objects' is checked, then it should retrieve an array of all selected objects in the view.
- If 'Active collection' is checked then it should retrieve all objects in the active collection.  This is done with `bpy.context.view_layer.active_layer_collection.collection.all_objects`

The downside is that parent-child relationships will need to be rebuilt - because now we have a flat array of objects instead of iterating up the hierarchy.  

This is the way the FBX exporter does it, and it seems to be quite well written for 2.80

https://developer.blender.org/diffusion/BA/browse/master/io_scene_fbx/export_fbx_bin.py$3046

I'm happy to help with this but I'm going to need some guidance because this is my first venture into Python and Blender internals.

donmccurdy:
Per https://github.com/KhronosGroup/glTF-Blender-IO/issues/109 + https://github.com/KhronosGroup/glTF-Blender-IO/issues/192 it sounds like we may want to replace the "export selected" feature with exporting active collections, instead. @emackey does that sound right, or were you thinking we'd have both?

AlexMuir:
[...]

Issue #1066, opened 2020-05-13
Export to .glb and .gltf results in near-empty files


cor-rine:
**Describe the bug**
Exporting to .glb or .gltf produces near-empty files that do not contain any geometry, mesh, or scene information.

**To Reproduce**
1. Open the sheep.blend file provided in the link below
2. Select sheep and go to File > Export > glTF 2.0
3. Select either .glb or .glTF export format (Include: Selected Objects, Geometry: Apply Modifiers)
4. Press Export, and observe created export files (also included in the Dropbox link for reference)

**Expected behavior**
Files should contain all scene and geometry and mesh information.  Export is successful so that the files can be loaded into Three.js web application.

**Actual behavior**
Files do not contain all scene and geometry and mesh information.  Export is unsuccessful and files cannot be loaded into Three.js web application.

**.blend file/ .gltf**
Files: https://www.dropbox.com/sh/nigh8m2yzd1i0fg/AAA2OU6lKB58Q_ZztaC9aVNMa?dl=0

**Version**
 - OS: macOS Mojave
 - Blender Version 2.82a 2020-03-12

julienduroure:
Hello,
Seems to be a duplicate of #193 
Currently, to export your sheep, you have to select the armature too.

cor-rine:
Ah! Excellent - thank you very much!

sofiavera24:
I am having an error when I want to export this model in GLTF format (I have followed this tutorial: https://www.youtube.com/watch?v=g4LmR-TYxVo ), I try to export it and it gives me an error, I attach error, I need to solve it as soon as possible, if someone can help me I would really appreciate it!

Error: Traceback (most recent call last):
  File "/Applications/Blender.app/Contents/Resources/2.83/scripts/addons/io_scene_gltf2/__init__.py", line 506, in execute
    return gltf2_blender_export.save(context, export_settings)
  File "/Applications/Blender.app/Contents/Resources/2.83/scripts/addons/io_scene_gltf2/blender/exp/gltf2_blender_export.py", line 42, in save
    json, buffer = __export(export_settings)
  File "/Applications/Blender.app/Contents/Resources/2.83/scripts/addons/io_scene_gltf2/blender/exp/gltf2_blender_export.py", line 55, in __export
    __gather_gltf(exporter, export_settings)
[...]

Issue #1030, opened 2020-04-15
Object can't be exported as .glb if it has a parent


Moonwards:
**Describe the bug**
An object with a parent will export as an empty file. If the file created is imported into a blender file (2.81a), the following errors are generated:

bpy.context.space_data.system_bookmarks_active = 1
Traceback (most recent call last):
  File "/home/kim/3dDev/blender-2.81a-linux-glibc217-x86_64/2.81/scripts/addons/io_scene_gltf2/__init__.py", line 923, in execute
    return self.import_gltf2(context)
  File "/home/kim/3dDev/blender-2.81a-linux-glibc217-x86_64/2.81/scripts/addons/io_scene_gltf2/__init__.py", line 937, in import_gltf2
    if self.unit_import(path, import_settings) == {'FINISHED'}:
  File "/home/kim/3dDev/blender-2.81a-linux-glibc217-x86_64/2.81/scripts/addons/io_scene_gltf2/__init__.py", line 960, in unit_import
    BlenderGlTF.create(self.gltf_importer)
  File "/home/kim/3dDev/blender-2.81a-linux-glibc217-x86_64/2.81/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py", line 30, in create
    BlenderScene.create(gltf)
  File "/home/kim/3dDev/blender-2.81a-linux-glibc217-x86_64/2.81/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_scene.py", line 60, in create
    BlenderScene.set_active_object(gltf)
  File "/home/kim/3dDev/blender-2.81a-linux-glibc217-x86_64/2.81/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_scene.py", line 88, in set_active_object
    vnode = gltf.vnodes[pyscene.nodes[0]]
TypeError: 'NoneType' object is not subscriptable

location: <unknown location>:-1

bpy.context.area.ui_type = 'INFO'

**To Reproduce**
1. Open new file, duplicate the cube
2. Parent one cube to the other
3. Export the child cube as a .glb, selecting the option `Include: Selected Objects`
4. Attempt to import the .glb in a blender file

**Expected behavior**
I would have thought that this parameter would be discarded and the object(s) exported as though they had no parent

This is so easy to reproduce I'll leave it as an exercise for the viewer 

**Version**
 - OS: Ubuntu 18.04
 - Blender Version 2.81a, 2.82a

[...]

PR #1126, opened 2020-07-16
Better filtering of exported nodes

emackey:
This PR deals with parent nodes and ancestor nodes that have been filtered out of glTF export, typically by "use_selected" failing to select the parents of live children.  Previously, such nodes would be completely excluded, leaving a hole in the scene graph with a disconnected section below.

In this PR, if a given node fails the export filter, but that node has descendants that pass the export filter, then the failed node is still a critical node to include in the glTF node hierarchy.  The failed node will not export any mesh, skin, camera, or weights.  But as a parent or ancestor of live descendants, the node will still include its transformations, animations, extras, and extensions, which may have visible effects on the descendants.

This PR could use a bit of additional testing, with animations and skinning in play.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
@@ -50,25 +50,45 @@ def gather_node(blender_object, library, blender_scene, dupli_object_parent, exp
 
 @cached
 def __gather_node(blender_object, library, blender_scene, dupli_object_parent, export_settings):
+    children = __gather_children(blender_object, blender_scene, export_settings)
+
+    camera = None
+    mesh = None
+    skin = None
+    weights = None
+
     # If blender_scene is None, we are coming from animation export
     # Check to know if object is exported is already done, so we don't check
     # again if object is instanced in scene : this check was already done when exporting object itself
     if not __filter_node(blender_object, blender_scene, export_settings):
-        return None
+        if children:
+            # This node should be filtered out, but has un-filtered children present.
+            # So, export this node, excluding its camera, mesh, skin, and weights.
+            # The transformations and animations on this node will have visible effects on children.
+            pass
+        else:
+            # This node is filtered out, and has no un-filtered children or descendants.
+            return None
+    else:
+        # This node is being fully exported.
+        camera = __gather_camera(blender_object, export_settings)
+        mesh = __gather_mesh(blender_object, library, export_settings)
+        skin = __gather_skin(blender_object, export_settings)
+        weights = __gather_weights(blender_object, export_settings)
 
     node = gltf2_io.Node(
-        camera=__gather_camera(blender_object, export_settings),
-        children=__gather_children(blender_object, blender_scene, export_settings),
+        camera=camera,
+        children=children,
         extensions=__gather_extensions(blender_object, export_settings),
         extras=__gather_extras(blender_object, export_settings),
         matrix=__gather_matrix(blender_object, export_settings),
-        mesh=__gather_mesh(blender_object, library, export_settings),
+        mesh=mesh,
         name=__gather_name(blender_object, export_settings),
         rotation=None,
         scale=None,
-        skin=__gather_skin(blender_object, export_settings),
+        skin=skin,
         translation=None,
-        weights=__gather_weights(blender_object, export_settings)
+        weights=weights
     )
 
     # If node mesh is skined, transforms should be ignored at import, so no need to set them here
@@ -105,6 +125,8 @@ def __filter_node(blender_object, blender_scene, export_settings):
             else:
                 # Not instanced, not linked -> We don't keep this object
                 return False
+    if blender_object.visible_get() is False:
+        return False
     if export_settings[gltf2_blender_export_keys.SELECTED] and blender_object.select_get() is False:
         return False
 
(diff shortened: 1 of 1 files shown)


scurest:
FWIW I tested this with some armatures and animation (object parented to animated bone, object skinned by arma in different tree) and I couldn't break it, so kudos!

I did find that it overapproximates what to export sometimes. It appears to always export all bones wherever they exist.

out

So there are some cases where master, when it works, is better

out2

I don't know how hard this would be to fix, but obviously those extra empty nodes are invisible in actual viewers, so better to overapproximate than under.

scurest:
@emackey So is this waiting for something? This fixes #193, #1030, #1066.

scurest:
Tested with linked collections. It doesn't work.

By "collections", do you mean if you click on a collection in the outliner like this and export with "Selected Objects" it should export the Cube?

col

scurest:
For linked collections, is it enough to check if the dupli_object_parent is selected?

<details>

<summary>Patch</summary>

````diff
diff --git a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
index f05f21e..7a405ae 100644
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
@@ -60,7 +60,7 @@ def __gather_node(blender_object, library, blender_scene, dupli_object_parent, e
     # If blender_scene is None, we are coming from animation export
     # Check to know if object is exported is already done, so we don't check
     # again if object is instanced in scene : this check was already done when exporting object itself
-    if not __filter_node(blender_object, blender_scene, export_settings):
+    if not __filter_node(blender_object, blender_scene, dupli_object_parent, export_settings):
         if children:
             # This node should be filtered out, but has un-filtered children present.
             # So, export this node, excluding its camera, mesh, skin, and weights.
@@ -113,7 +113,7 @@ def __gather_node(blender_object, library, blender_scene, dupli_object_parent, e
     return node
 
 
-def __filter_node(blender_object, blender_scene, export_settings):
+def __filter_node(blender_object, blender_scene, dupli_object_parent, export_settings):
     if blender_object.users == 0:
         return False
     if blender_scene is not None:
@@ -125,8 +125,11 @@ def __filter_node(blender_object, blender_scene, export_settings):
             else:
                 # Not instanced, not linked -> We don't keep this object
                 return False
-    if export_settings[gltf2_blender_export_keys.SELECTED] and blender_object.select_get() is False:
-        return False
+    if export_settings[gltf2_blender_export_keys.SELECTED]:
+        if dupli_object_parent and bpy.data.objects[dupli_object_parent].select_get():
+            return True
+        if blender_object.select_get() is False:
+            return False
 
     return True
 
````

</details>

julienduroure:
Was thinking about another way to solve the issue:
* Construct a virtual tree as done in importer, calculating world matrix for each node
* Based on that, using all roots of this virtual tree to recursively go inside tree

scurest:
I think so too, two passes are necessary.

emackey:
@julienduroure That sounds good, but would it involve "skipping" intermediate nodes, leaving them out of the tree?  Sometimes, multiple children depend on a common parent node that is turned off for export, but the parent may have an animation applied that affects all of the children's locations.  Apologies if I misunderstood your suggestion.

julienduroure:
Checking animation on node must be part of checks that need to be done in order to know if we can skip a node or node.
I was mostly thinking about upper nodes (from blender root to the node, if root is not selected), but we can extend the solution to intermediate node too. Will try to write some quick drawing to explain

scurest:
Is this blocked for "the perfect is the enemy of the good" reasons?

emackey:
@scurest This started to touch too many edge cases and Blender features I'm not well-versed on.  We last talked about two-pass processing here with support for linked collections, armatures, etc., which are not tools I normally use in my own simplified workflow.  I'm unsure how best to get this moving again.  If you have ideas for that, you could make a branch off this branch, or a whole separate implementation if that's easier.  Otherwise, I'll try to come back around to this, eventually, but it won't be quick.

scurest:
What I mean is, why can't it be merged as is? Linked collections don't work in master either.

emackey:
Ah, if this doesn't make things worse, then yes, let's merge it.  Not sure if Julien might be away for a couple days though...

Note that Blender 2.91 "Bcon 3" [starts tomorrow](https://developer.blender.org/project/view/117/), meaning bug fixes only for that particular version.  But I do consider this a bug fix, so if it doesn't break new things it can go into 2.91.

julienduroure:
Hello,
I'm here :)
Will merge it soon

scurest:
Neat. I can confirm that #193 and #1030 now work with master. Can someone close?

TheKruspe:
Is ist possible to export an object without its parent nodes?

emackey:
Without parent nodes, the exported object ends up as an orphan in the scene graph, so does not render at all.

scurest:
@TheKruspe If you mean like

````
BLENDER                             GLTF

A         (unselected)     ====>    B
└─ B      (selected)                └─ C
   └─ C   (selected)
````

Then no, the exporter cannot do that; A always gets pulled along.

emackey:
Right, and in the case @scurest mentions, "A" might have transformations and/or animations applied to it, that B and C are required to inherit.  So you can't just declare B to be a new root of the graph.

scurest:
I mean, you could. A's transforms would be folded into the transform for B. You can pretty use whatever hierarchy you want; just compute all local transforms as `(parent world).inverted() @ (self world)`.

julienduroure:
This is one of the purpose of #1244 
I locally performed some experiments recently on caching/hierarchy. I will write a proposition soon

PR #1127, opened 2020-07-16
Export: always export loop normals

### LINKED ISSUE #73: Sharp Edges Ignored
Exporter ignores manually-assigned hard/sharp edges in Blender 2.8, resulting in an 'all-soft' mesh.

Repro in Blender 2.80 Experimental with glTF-Blender-IO: 

1. Create a cube
1. Display>Shading>Smooth Edges
2. In Edit mode, select all edges. Edge>Edge Data>Mark Sharp
4. Select the Object Data tab, Turn on Auto Smooth and set the angle to 180. 
5. Edges of cube should appear hard/sharp in the viewport.
6. With cube selected File>Export glTF 2.0 (glb). Make sure Export normals and Export Tangents are checked.
7. Write file to disk. 
8. Load file in sandbox.babylonjs.com/
9. Hard edges will not be preserved.

Contrast the same workflow with Blender 2.79 using glTF-Blender-Exporter: Hard edges are preserved in .glb written to disk.

Note: In 2.80, setting Display>Shading>Flat faces does export a .glb with hard edges, but all edges are hard. What we want is the ability to define custom hard/soft edges on any given mesh (like we can do in Blender 2.79 with the glTF-Blender-Exporter addon).

Example files attached.
[cube_hard_soft.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/2559855/cube_hard_soft.zip)

### LINKED ISSUE #1119: Question: Why not just use mesh.loops[i].normal?

scurest:
When exporting, always use loop normals instead of sometimes using vertex/face normals. This is simpler and faster.

I made a test of this, see my results [here](https://github.com/KhronosGroup/glTF-Blender-IO/issues/1119#issuecomment-659597910). I found no visual difference in all but one case (flat polys with Auto Smooth on), and in that one case this branch is correct and master is wrong.

Closes  #1119.

edit: also fixes #73.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
@@ -142,9 +142,7 @@ def extract_primitives(glTF, blender_mesh, library, blender_object, blender_vert
 
     use_normals = export_settings[gltf2_blender_export_keys.NORMALS]
     if use_normals:
-        if blender_mesh.has_custom_normals:
-            # Custom normals are all (0, 0, 0) until calling calc_normals_split() or calc_tangents().
-            blender_mesh.calc_normals_split()
+        blender_mesh.calc_normals_split()
 
     use_tangents = False
     if use_normals and export_settings[gltf2_blender_export_keys.TANGENTS]:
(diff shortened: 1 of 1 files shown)


julienduroure:
@scurest Can you please rebase?

scurest:
Yeah, rebased.

PR #1136, opened 2020-07-22
Import/Perf: for "Use Normal Data", numpyify the code that guesses which polys are flat

scurest:
The calculation of whether each poly should be flat/smooth takes up a large part of the import time when using "Use Normal Data". For unskinned meshes, it's often about half the time. This PR numpyifys it.

Just like last time I touched this code, because of tiny numerical differences this is usually but not always the same as master. When it differs, it appears to usually but not always be slightly more strict (ie. it marks fewer tris flat).

### Alternative

The main alternative is to set all polys to smooth (ie. never try to guess a poly is flat). I'm pretty sure this never causes any visual difference, though there are other ways you can tell whether a poly is flat or not (eg. extrude it). It's not totally clear to me what the goal of guessing which polys should be flat-shaded is so I don't really know what this would mess up.

### Perf

The fourth column shows the alternative (all polys smooth) mentioned above. You can see the time difference is (somewhat surprisingly) negligible. The final column gives the time with "Smooth Shading" for a point of reference.

|Model|master|this branch|alternative|Smooth Shading|
|:---|:---|:---|:---|:---|
|modelGroup.glb|109s|47s|45s|31s|
|Suzanne, subdived 5 times, exported|10.3s|4.9s|4.8s|2.0s|
|MetalRoughSpheres|3.3s|1.3s|1.2s|0.7s|

Note that this branch uses more memory for the intermediate arrays though.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
@@ -352,32 +352,8 @@ def do_primitives(gltf, mesh_idx, skin_idx, mesh, ob):
     # ----
     # Normals
 
-    # Set poly smoothing
-    # TODO: numpyify?
-    smooths = []  # use_smooth for each poly
-    f = 0
-    for prim in pymesh.primitives:
-        if gltf.import_settings['import_shading'] == "FLAT" or \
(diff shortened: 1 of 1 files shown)


julienduroure:
Thanks!

PR #1142, opened 2020-07-23
Import: bugfix for skinned mesh normals not being renormalized

scurest:
Bug introduced in #1121.

I forgot to renormalize normals after skinning them into the bind pose. The old code for `skin_normal` re-normalized here

https://github.com/KhronosGroup/glTF-Blender-IO/blob/f5128eb4b72ab9366441bc054b96589d973a6177/addons/io_scene_gltf2/blender/imp/gltf2_blender_primitive.py#L111

The bug is visible as too many polys being marked flat in models that have big scale factors. The code for marking polys flat depends on the dot product of two normals being >=1 only when they are equal, which isn't true if the normals have lengths >=1.

This is a bugfix, so PRed against the 2.90 branch.


julienduroure:
Thanks!

Issue #1124, opened 2020-07-16
Export joint index attributes as unsigned byte


matthew-mcallister:
Currently, vertex joint indices are always exported as 16-bit ints. However, if a model has 256 deforming joints or less, which I think is usually the case, it would be preferable to convert from shorts to bytes. This could conceivably be implemented by external optimizers, but I haven't seen one that does, and it might be easier to do it in the exporter anyways.

Peach1:
That seems to be a few kilobytes of space saving, and byte indexes usually have no impact on performance. Could you explain your use case which benefits from this optimization?

donmccurdy:
> This could conceivably be implemented by external optimizers...

The [gltfpack](https://github.com/zeux/meshoptimizer) CLI tool implements this. I'll probably include it in https://github.com/donmccurdy/glTF-Transform/pull/59 as well.

> That seems to be a few kilobytes of space saving...

Potentially a byte _per vertex_, so in a model with many vertices this would matter. Especially combined with similar optimization of the other vertex attributes.

But I think it would have to be a fairly simple and safe implementation to include in this exporter... skinning is difficult enough to get right without complicating it with optimizations that could be done in a standalone optimizer. I haven't looked enough at this bit of the code to know whether that is the case.

scurest:
Assuming the best case of a model with only POSITIONs, JOINTS_0s, and WEIGHTS_0s, it would save 11% of the size of vertex attribute arrays. If there are also NORMALs and TEXCOORD_0s, it would be 7%.

It would be easy enough to do. Simplest way would be around [here](https://github.com/KhronosGroup/glTF-Blender-IO/blob/98078a0edd5e7bba109737d7495fbc238fd8e86c/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitive_attributes.py#L163), just turn the the joints array into a numpy array of uint8s if there are few enough joints.

matthew-mcallister:
> The [gltfpack](https://github.com/zeux/meshoptimizer) CLI tool implements this.

Ah, that's my fault for not doing enough research. I mistook gltfpack for a simple frontend to meshopt's quantization/optimization routines, but I see it's much more comprehensive than that. Thanks for the info!

If no one objects to including the optimization, I will happily submit a PR. Otherwise, I agree that it would be no great loss.

donmccurdy:
[...]

PR #1144, opened 2020-07-23
Export joint indices as uint8 when possible

matthew-mcallister:
Closes https://github.com/KhronosGroup/glTF-Blender-IO/issues/1124. I tested this on a rig with <256 joints and a rig with >256 joints. Both exported the expected accessor type, had correct joint data, passed validation, and could be imported back into Blender with the correct result. Clearly, making the change was easy; creating the rig with >256 joints turned out to be a much longer detour.


donmccurdy:
Thank you!

scurest:
@julienduroure This change was lost by 05c31d17.

emackey:
Looks restored in f058163bb28f4.

Issue #1031, opened 2020-04-15
Support more inputs for principled BSDF


schiette:
**Is your feature request related to a problem? Please describe.**
Nodes connected to the material slots are only processed when it's a Image Texture node. Simple combinations like "Image Texture Node + Hue Saturation" are not supported (it will export the original texture, not the adjusted value).

Other examples that are not supported as inputs on the bsdf:
- RGB node 
- All color correction nodes (brightness/contrast,hue/saturation, invert, Mixrgb)
- Colorramp

Because of this, even very simple materials have to be baked. 

**Describe the solution you'd like**

I was wondering whether the plugin could support a few extra very common setups for inputs on the principled bsdf shader.

**Describe alternatives you've considered**

Baking is an alternative, but it's less efficient because it also bakes solid colors to maps.

**Additional context**

![nodesetup1](https://user-images.githubusercontent.com/29231922/79340807-df34f380-7ef8-11ea-8977-cb4af2ec63a1.JPG)
![nodesetup2](https://user-images.githubusercontent.com/29231922/79340810-df34f380-7ef8-11ea-838a-cb4120fb3f96.JPG)
![nodesetup3](https://user-images.githubusercontent.com/29231922/79340811-dfcd8a00-7ef8-11ea-91ef-e0372d9967c1.JPG)

PR #1155, opened 2020-07-31
Export: materials: export factors from MULTIPLY nodes

### LINKED ISSUE #68: After merge of PR #48, export of imported file fails
After merge of PR #48 , exporting file that are an imported gltf file is broken.
Failed because of MULTIPLY nodes created for take into account the Base Color Factor.

### LINKED ISSUE #393: BaseColorFactor not exported
BaseColorFactor is not exported with following node setup: (This is the default shader node created by importer)
Note, the model has no texture and only uses vertex color.
Screenshot 2019-03-31 13:08:55 

Edit: 
You can get the model here : https://skfb.ly/6IVX9

### LINKED ISSUE #646: How to export roughness / metalness / aoMapIntensity
Hi,

I did a lot of research, but couldn't find a way to export roughness / metalness / aoMapIntensity factors (in contrast to maps). In the old gltf-Blender-Exporter, these factors were directly accessible through the gtlf node.

For example, here is my material setup:

image

I would like to have more control over the roughness and ambient occlusion maps and also change the factor. Is there a way to do that? I tried inserting math multiply nodes and rgb multiply nodes at various locations without success. As a workaround I am currently using userData / custom properties in Blender.

I would greatly appreciate it if someone could point out if I am missing something. If this feature is currently missing, having it implemented would be a great addition.

Cheers
-Sebastian

scurest:
Fixes #68.  
Fixes #393.  
Fixes #646.

This adds a roundtrip test for emissiveFactor, baseColorFactor, roughnessFactor, and metallicFactor when there are also textures. Then fixes the test. The exporter can now read factors from a MULTIPLY node in front of a socket.

There is one case where it already did this: the RGB part of the baseColorFactor, see #223. It would look for a MULTIPLY anywhere on the path to the texture node. However, there doesn't have to be a texture node. This changes it so it only looks for a MULTIPLY node _right before_ the socket. It also removes a warning about the MULTIPLY having linked sockets.

-----

As a small bonus, the function __get_const_from_socket can also read a constant value from an RGB/Value node so this now works: (this was one of the things requested in #1031)

red


tests/test/test.js
```diff
@@ -908,6 +908,30 @@ describe('Importer / Exporter (Roundtrip)', function() {
                 assert.strictEqual(asset.images[0].uri, '08_tiny-box-rgb.png');
             });
 
+            it('roundtrips baseColorFactor, etc. when used with textures', function() {
+                let dir = '11_pbr_factors';
```
scurest, line 912: Maybe it would be better to name this file 11_factors_with_textures?

donmccurdy:
> This changes it so it only looks for a MULTIPLY node right before the socket.

Is there a concern with using MULTIPLY nodes further back? If you're multiplying baseColor by something and then joining it with an alpha texture, it might be a little unintuitive that the multiplication needs to come after the join. Just a thought, not necessarily a problem.

scurest:
You could I guess. Of course if you plug in anything non-trivial we're gonna export something wrong anyway, so it's mostly a question of the precise way we'd be getting it wrong. Multiply-must-be-right-before-the-socket is easy to understand and all the linked issues did put it there.

> If you're multiplying baseColor by something and then joining it with an alpha texture

Not sure what kind of graph you're thinking of. Color and alpha are always separate in the material graph?

The one case I think we could understand but this doesn't is that if you're multiplying by the vertex color too, you can multiply in either order.

col_000

scurest:
If you wanted to search further back this would do it

````diff
diff --git a/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py b/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
index 27f4ae1..f735f21 100644
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
@@ -223,16 +223,17 @@ def get_node(data_path):
 def get_factor_from_socket(socket, kind):
     """
     For baseColorFactor, metallicFactor, etc.
-    Get a constant value from a socket, or a constant value
-    from a MULTIPLY node just before the socket.
+    Get a constant value from a socket, or from a MULTIPLY
+    node somewhere in the tree that feeds into socket.
+    The tree is searched depth-first.
     kind is either 'RGB' or 'VALUE'.
     """
     fac = __get_const_from_socket(socket, kind)
     if fac is not None:
         return fac
 
-    node = __previous_node(socket)
-    if node is not None:
+    def get_factor_from_node(node):
+        """If node multiplies by a constant, return that constant."""
         x1, x2 = None, None
         if kind == 'RGB':
             if node.type == 'MIX_RGB' and node.blend_type == 'MULTIPLY':
@@ -245,6 +246,24 @@ def get_factor_from_socket(socket, kind):
                 x2 = __get_const_from_socket(node.inputs[1], kind)
         if x1 is not None and x2 is None: return x1
         if x2 is not None and x1 is None: return x2
+        return None
+
+    visited = set()
+    queue = [__previous_node(socket)]
+    while queue:
+        node = queue[0]
+        queue = queue[1:]
+
+        if node is None: continue
+        if node.type == 'TEX_IMAGE': continue  # stop at Image Texture node
+        if node.name in visited: continue
+        visited.add(node.name)
+
+        factor = get_factor_from_node(node)
+        if factor is not None: return factor
+
+        for soc in node.inputs:
+            queue.append(__previous_node(soc))
 
     return None
 
````

donmccurdy:
Ok, this looks good to me as-is. Thanks!

PR #1157, opened 2020-07-31
Export: always write texture samplers

### LINKED ISSUE #307: Empty Samplers node
The Blender glTF exporter writes a file with the samplers node empty.  This causes any text on a texture to render highly aliased in some viewers, particularly the Microsoft Mixed Reality desktop viewer.

Other exporerts, such as Substance Painter, export a glTF file with correct data to render text smooth:
"magFilter": 9729,
"minFilter": 9985,
"wrapS": 10497,
"wrapT": 10497

When the above lines are manually edited into the glTF file, the text renders with a smooth linear filter.
Can we have the option to include this samplers info in the Blender glTF exporter?

rendering-screenshots
samplersinfo-missing
[glb-files.zip](https://github.com/KhronosGroup/glTF-Blender-Exporter/files/2330359/glb-files.zip)

scurest:
Fixes #307.

According to the spec, when texture filtering options are not defined

> **Default Filtering Implementation Note:** When filtering options are defined, runtime must use them. Otherwise, it is free to adapt filtering to performance or quality goals.

But the exporter currently skips writing a sampler for eg. a LINEAR/REPEAT texture node, acting as if the default filtering were linear.

This deletes the __filter_sampler function and always writes a texture sampler.

The exporter already wrote duplicate samplers, but removing __filter_sampler exacerbates this problem since most textures will have the same LINEAR/REPEAT sampler. So the second commit adds a `@cache`d function for deduping identical samplers.

The third commit is just a minor refactor to use the named constants for sampler values instead of numbers + comments.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_sampler.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_sampler.py
@@ -20,9 +20,6 @@ from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extension
 
 @cached
 def gather_sampler(blender_shader_node: bpy.types.Node, export_settings):
-    if not __filter_sampler(blender_shader_node, export_settings):
-        return None
-
     sampler = gltf2_io.Sampler(
         extensions=__gather_extensions(blender_shader_node, export_settings),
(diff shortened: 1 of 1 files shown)


julienduroure:
Thanks!

PR #1159, opened 2020-08-02
Export: materials: support emission strength

### LINKED ISSUE #367: Exporter, does not save Emissive Factor/Emission Strength Value for materials.
Currently the Emissive Factor/Emission Strength is not exported to the .glb file.

-------
Steps to Reproduce:

Following this [example](https://docs.blender.org/manual/en/2.80/_images/addons_io-gltf2_material-emissive.png)

I set up a material using the Principal BSDF and Emission nodes.  I then make some adjustments to the Principal and Emission settings to get something looking like [this scene.](https://i.imgur.com/5Az9fEl.png) 

The nodes on each material [look like this.](https://i.imgur.com/39s8mNc.png)

After exporting an re-importing I get this. [imported nodes](https://i.imgur.com/7qaPrMR.png)

I wasn't sure if it was the import process or the export process so I checked the .glb embedded file.  It appears the emissiveFactor portion is always being saved as 1, 1, 1 (see below)

```
"emissiveFactor" : [
                1,
                1,
                1
            ]
```

Knowing that the Emissive Factor is supposed to be an RGB value I also tried attaching an RGB node and attaching it to the strength of the Emission node, however it was still not saved.

-------
Additional Notes:

I also went in and edited the "emissiveFactor" in the embedded .glb file.  Like so...
```
"emissiveFactor" : [
                0.33,
                0.33,
                0.33
            ]
```
When importing this edited file, I get a node tree [including a multiply node.](https://i.imgur.com/MfHIGt1.png) Where the Color2 of the multiply node matched my edited .glb file.  

The result looks okay, but is too bright since it still takes into account the emission node's strength.  (Which is always set to 1.0)

Upon exporting this new scene the emissiveFactor is still not saved.

-----
Using Blender 2.8 Beta Windows 64 bit, March 18th, 19:10:14 build.

Edit: updated example link

scurest:
~~Done on top of #1155, so do that first.~~ Rebased onto master.

Fixes #367.

Adds support for Emission Strength. The strength is just multiplied into the emissiveFactor when present.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
@@ -109,13 +109,36 @@ def __gather_emissive_factor(blender_material, export_settings):
     if emissive_socket is None:
         emissive_socket = gltf2_blender_get.get_socket_old(blender_material, "EmissiveFactor")
     if isinstance(emissive_socket, bpy.types.NodeSocket):
-        fac = gltf2_blender_get.get_factor_from_socket(emissive_socket, kind='RGB')
-        if fac is None and emissive_socket.is_linked:
+        factor = gltf2_blender_get.get_factor_from_socket(emissive_socket, kind='RGB')
+
+        if factor is None and emissive_socket.is_linked:
             # In glTF, the default emissiveFactor is all zeros, so if an emission texture is connected,
             # we have to manually set it to all ones.
-            fac = [1.0, 1.0, 1.0]
-        if fac == [0, 0, 0]: fac = None
-        return fac
+            factor = [1.0, 1.0, 1.0]
+
+        if factor is None: factor = [0.0, 0.0, 0.0]
+
+        # Handle Emission Strength
+        strength_socket = None
+        if emissive_socket.node.type == 'EMISSION':
+            strength_socket = emissive_socket.node.inputs['Strength']
+        elif 'Emission Strength' in emissive_socket.node.inputs:
+            strength_socket = emissive_socket.node.inputs['Emission Strength']
+        strength = (
+            gltf2_blender_get.get_const_from_socket(strength_socket, kind='VALUE')
+            if strength_socket is not None
+            else None
+        )
+        if strength is not None:
+            factor = [f * strength for f in factor]
+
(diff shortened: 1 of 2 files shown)


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
```diff
@@ -111,13 +111,30 @@ def __gather_emissive_factor(blender_material, export_settings):
...
-        if fac == [0, 0, 0]: fac = None
-        return fac
+            factor = [1.0, 1.0, 1.0]
+
+        # Handle Emission Strength
+        strength_socket = None
+        if emissive_socket.node.type == 'EMISSION':
+            strength_socket = emissive_socket.node.inputs['Strength']
+        elif 'Emission Strength' in emissive_socket.node.inputs:
+            strength_socket = emissive_socket.node.inputs['Emission Strength']
```
scurest, line 126: These two lines should support [D4971](https://developer.blender.org/D4971), but I haven't tested it or anything.

julienduroure:
Ok, so this means that we can't merge this PR right now. What is the solution for now?
Export it only if blender factor is < 1.0 ? Export always, but clamp to 1.0 when necessary?

emackey:
Export when possible, but clamped to a maximum of 1.0.  This is on the Khronos roadmap though, so hopefully someday soon we'll get an extension going that works around this limitation.

emackey:
Actually it looks like the 1.0 clamp is already included here.

tests/roundtrip/11_pbr_factors/11_pbr_factors_report.json
```diff
@@ -0,0 +1,55 @@
+{
+    "uri": "roundtrip/11_pbr_factors/11_pbr_factors.gltf",
```
emackey, line 2: Once you merge or rebase master, this report file might not be needed anymore.

scurest:
This does clamp when >1. I moved the clamp out to a separate block and put a comment on it so it is more obvious.

emackey (approved):

scurest:
Fun fact: you can set the RGBA values above 1 in Blender. I did not know that.

julienduroure:
Thanks!

PR #1162, opened 2020-08-03
Import: select all (and only) the imported objects

### LINKED ISSUE #1160: gltf importer does not select imported objects after import
**Describe the bug**
After importing a glb file using the blender's importer the imported objects are not setted active
**To Reproduce**
1.Import any gltf file using the importer
fp = "filepath_to_the_gltf_file"
bpy.ops.import_scene.gltf(filepath=fp)
2. Check the selected objects
print(bpy.context.selected_objects)

**Expected behavior**
The imported objects should be active, In other words, bpy.context.selected_objects should keep track of the new imported objects. All the others blender's importers do so

**Version**
 - OS: windows
 - Blender Version: 2.82a

scurest:
Fixes #1160.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_scene.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_scene.py
@@ -43,6 +43,7 @@ class BlenderScene():
 
         if bpy.context.mode != 'OBJECT':
             bpy.ops.object.mode_set(mode='OBJECT')
+        BlenderScene.select_imported_objects(gltf)
         BlenderScene.set_active_object(gltf)
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


addons/io_scene_gltf2/blender/imp/gltf2_blender_scene.py
```diff
@@ -43,6 +43,7 @@ def create(gltf):
 
         if bpy.context.mode != 'OBJECT':
             bpy.ops.object.mode_set(mode='OBJECT')
+        BlenderScene.select_imported_objects(gltf)
         BlenderScene.set_active_object(gltf)
```
donmccurdy, line 47: Might be worth including a comment here, indicating that we're doing this for consistency with addons X, Y, and Z.

scurest: It seems natural enough to me.

donmccurdy: If we got a bug report, "Importing glTF model loses my previous selection, please don't", I guess I'd want to remember why we do this — _Not_ selecting the imported objects also seemed natural enough to me, until learning it was a convention already. ;)

scurest: Do you want me to add a comment? You could already have git blamed it, I put the reason in the commit message.

The reason its natural is not just the other importers, even just adding a Cube has this behavior. Everything else is deselected and only the Cube is selected.

donmccurdy: Git blame disappears when code moves around — I tend to prefer having real comments on anything that is likely to be unintuitive to someone who didn't write it. But if you feel this is intuitive enough, we can skip the comment here.

Issue #1156, opened 2020-07-31
Questions about roundtrip tests


scurest:
What's the reason for the validator reports being checked into this repo? Versus generating a report for both the original and roundtripped file at test time.

Also, is there a way to run just a specific test?

emackey:
Validation reports fixed in #1168.

I don't know of a way to run a test just on a specific model, but that would be a nice feature to add.

emackey:
Oh I stand corrected!  The `-g` flag works on model names.

```
$ ./node_modules/.bin/mocha -b --reporter mochawesome -g metallic


  Exporter
    blender_export
      01_metallic_sphere (670ms)
      05_metallic_sphere_light (619ms)
    blender_export_glb
      01_metallic_sphere (590ms)
      05_metallic_sphere_light (623ms)

  Importer / Exporter (Roundtrip)
    blender_roundtrip
      01_metallic_sphere (597ms)
      05_metallic_sphere_light (645ms)
    blender_roundtrip_glb
      01_metallic_sphere (474ms)
      05_metallic_sphere_light (593ms)

```

scurest:
Cool, thanks!

PR #1168, opened 2020-08-05
Tests: don't check in validator reports

scurest:
No one responded to #1156 lol

For roundtrip tests, the validator report for the src file is checked in to the repo. At test time, a report for the dst file is generated and compared to the checked in report.

This generates both the src and dst reports at test time. All _report.json files are deleted. This decreases the effort needed to write a roundtrip test. Regenerating all reports like #875 did would also not be necessary.


julienduroure:
@emackey @pjoe I am not really familiar with test process. Can you please have a look on this PR?

emackey:
Great, thanks @scurest, and sorry for the delayed responses these days.

PR #1169, opened 2020-08-05
Export: store vertex colors as normalized ushorts

### LINKED ISSUE #1166: .glb file size increases by 5x on export
Hi

I have a .glb obtained where the texture information is embedded in the .glb itself. I wanted to use Blender to adjust the origin of the mesh and export it out again into a .glb file without any other change. The exported file's size now increases by 5 times. 

I have attached the original .glb file and the blender settings that I had used.  Can somebody tell me why the blender output is 5 times larger ? 

Files: Original .glb file - 38 MB (inside 
[original.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/5024753/original.zip)
original.zip)

image

scurest:
I don't know if you want to do this or not. It would halve the size of COLOR_n arrays.

Closes #1166 I guess.


scurest:
btw is @julienduroure back yet?

scurest:
I realized that Blender only stores vertex color at 8-bit sRGB precision, see https://devtalk.blender.org/t/full-precision-float-vertex-colors/4706. So there is no reason not to do this.

PR #1170, opened 2020-08-06
Export: detect when a file will be overwritten (#718)

### LINKED ISSUE #718: Exporter: not detecting overwritten file
During export, because we don't write extension in filename field, blender don't detect when a file already exists in OS, so it's displayed with red background.
We have to check if there is a way to have it, keeping feature about exporting gltf or glb based on export options.

scurest:
Fixes #718.

Detects (ie. the picker turns red) when you're going to overwrite a file. The file extension is now shown on the picker screen. It updates whenever you change the filename or change the format option.

I think the behavior for scripting is the same as master except in the obscure case when glTF extensions would "double-up"

````py
bpy.ops.export_scene.gltf(filepath='a.gltf', export_format='GLB')
# master => a.gltf.glb
# this branch => a.glb
bpy.ops.export_scene.gltf(filepath='a.glb', export_format='GLTF_EMBEDDED')
# master => a.glb.gltf
# this branch => a.gltf
````

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -69,6 +69,17 @@ from bpy_extras.io_utils import ImportHelper, ExportHelper
 
 extension_panel_unregister_functors = []
 
+
+def on_export_format_changed(self, context):
+    # Update the file extension when the format (.glb/.gltf) changes
+    sfile = context.space_data
+    operator = sfile.active_operator
+    if operator.check(context):
+        # Weird hack to force the filepicker to notice filename changed
+        bpy.ops.file.filenum(increment=1)
+        bpy.ops.file.filenum(increment=-1)
+
+
 class ExportGLTF2_Base:
     # TODO: refactor to avoid boilerplate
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: addons/io_scene_gltf2/__init__.py [hunk 1]
```diff
@@ -69,6 +69,17 @@ def reload_package_recursive(current_dir, module_dict):
...
 
+
+def on_export_format_changed(self, context):
+    # Update the file extension when the format (.glb/.gltf) changes
+    sfile = context.space_data
+    operator = sfile.active_operator
+    if operator.check(context):
+        # Weird hack to force the filepicker to notice filename changed
+        bpy.ops.file.filenum(increment=1)
+        bpy.ops.file.filenum(increment=-1)
```
scurest, line 80: Check out this hack. Took forever to come up with.

scurest:
CI tests don't seem to be running?

addons/io_scene_gltf2/__init__.py
scurest: It does mean trailing 0s get erased when switching formats though. That sucks :/

scurest:
I pushed fixes to weird names (like `''` or `'.glb'`) so they behave like the other exporters. I also pushed a new hack that handles trailing 0s in the filename. I tested all these names and they switch correctly when changing the format.

````
0.glb     a0.glb
1.glb     a1.glb
9.glb     a9.glb
00.glb    a00.glb
01.glb    a01.glb
10.glb    a10.glb
09.glb    a09.glb
99.glb    a99.glb
.glb      ''
````

PR #1176, opened 2020-08-11
Import/Perf: numpyify morph target animations

scurest:
Uses numpy in gltf2_blender_animation_weight.py. This is easy since we don't need to transform the weight curves or anything (like we do for animation_node).

I don't have any glTFs with large enough morph target animations to give performance numbers though.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation_weight.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation_weight.py
@@ -14,6 +14,7 @@
 
 import json
 import bpy
+import numpy as np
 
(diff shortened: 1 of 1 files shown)


scurest:
Anyone know why CI tests aren't running btw?

scurest:
Hmm, actually I'll retract this. I also [numpyified animation_node.py](https://github.com/KhronosGroup/glTF-Blender-IO/compare/ae4ea71...scurest:numpyify-anim-node) and got some tests run. The gains were either very modest or, on files with lots of curves with a few keyframes (which is common), negative. Also foreach_set seems to be faster with a list here, so we can't avoid decoding to Python data anyway.

Issue #1178, opened 2020-08-13
Bake shape keys controlled by drivers?


will-ca:
I believe it's fairly common and pretty useful to have shape keys animated via drivers tied to armature bones' transforms.

When this is the case, the NLA tracks corresponding to the shape key are those of the armature, not those of the shape key itself.

From a quick test in Blender 2.83.4, it doesn't seem that the shape key's animations get exported usefully in this case.

julienduroure:
Hello,

Please provide a test case in order to check.
The only case that _should_ be taken into account for now is shapekey driven by a bone transformation from an armature that deforms the mesh, where the armature is parent of the mesh.

will-ca:
@julienduroure 

Ah, cool. Setting a parenting relationship to the armature seems to make it work perfectly.

My test case file just didn't have any parenting, modifier, or any kind of dependency relationship between the armature and the mesh aside from the shape key, but let me know if I should still post it.

julienduroure:
Well, seems you can export properly. Closing the ticket.
Thanks !

will-ca:
Hm. For future reference for anyone finding this, it seems to work even when the armature doesn't deform the mesh, as long the armature is the parent of the mesh— I.E., with "Parent" set, but with "Parent Type" set to "Object" and with no Armature modifier.

PR #1180, opened 2020-08-14
Docs: Mention requirements for sampling driver-controlled shape keys.

will-ca:
I ran into a workflow difficulty in #1178. It's conceivable that other people may as well.

I don't think I saw this information anywhere else in the documentation or UI, but I may have missed something.

I also don't actually know how this behaviour is implemented, so I described only the specific case I encountered (and which is probably the most likely to be encountered). Is it special-cased for only `Armature`>`Bone`>`Transform`>`Driver` shape key animations, is it a consequence of more general behaviour around sampling of animations in object hierarchies, or is it something in between?


julienduroure:
Thanks!

Issue #1183, opened 2020-08-17
bpy.ops.import_scene.gltf doesn't work with bpy.ops.wm.read_factory_settings?


vergol:
I'm trying to load gltf after clearing the scene, like shown here:
https://github.com/KhronosGroup/glTF-Blender-IO/blob/7a3532e32fcce61ab08aa0c081282e72e668ea4c/tests/roundtrip_gltf.py
But I'm getting the error:

```
Error: Traceback (most recent call last):
  File "\Text", line 34, in <module>
  File "\Text", line 32, in process_one
  File "\Text", line 12, in process_gltf
  File "C:\Program Files\Blender Foundation\Blender 2.83.4\2.83\scripts\modules\bpy\ops.py", line 201, in __call__
    ret = op_call(self.idname_py(), None, kw)
RuntimeError: Error: Traceback (most recent call last):
  File "C:\Program Files\Blender Foundation\Blender 2.83.4\2.83\scripts\addons\io_scene_gltf2\__init__.py", line 900, in execute
    return self.import_gltf2(context)
  File "C:\Program Files\Blender Foundation\Blender 2.83.4\2.83\scripts\addons\io_scene_gltf2\__init__.py", line 919, in import_gltf2
    return self.unit_import(self.filepath, import_settings)
  File "C:\Program Files\Blender Foundation\Blender 2.83.4\2.83\scripts\addons\io_scene_gltf2\__init__.py", line 937, in unit_import
    BlenderGlTF.create(self.gltf_importer)
  File "C:\Program Files\Blender Foundation\Blender 2.83.4\2.83\scripts\addons\io_scene_gltf2\blender\imp\gltf2_blender_gltf.py", line 42, in create
    BlenderGlTF._create(gltf)
  File "C:\Program Files\Blender Foundation\Blender 2.83.4\2.83\scripts\addons\io_scene_gltf2\blender\imp\gltf2_blender_gltf.py", line 49, in _create
    BlenderScene.create(gltf)
  File "C:\Program Files\Blender Foundation\Blender 2.83.4\2.83\scripts\addons\io_scene_gltf2\blender\imp\gltf2_blender_scene.py", line 40, in create
    BlenderNode.create_vnode(gltf, 'root')
  File "C:\Program Files\Blender Foundation\Blender 2.83.4\2.83\scripts\addons\io_scene_gltf2\blender\imp\gltf2_blender_node.py", line 51, in create_vnode
[...]

PR #1184, opened 2020-08-18
Replace factory reset with select-all and delete.

emackey:
For developers with custom system scripts folders, as described in the DEBUGGING.md document, the factory reset was killing off
the glTF addon itself.

This replaces it with a simple select-all and delete, so the roundtrip tests don't get polluted by the default cube, light, and camera, but any custom folder settings will survive.

This shouldn't have any real effect on CircleCI, but will be of use to developers running tests locally.

### DIFF
--- a/tests/roundtrip_gltf.py
+++ b/tests/roundtrip_gltf.py
@@ -25,7 +25,9 @@ try:
 
     filepath = argv[0]
 
-    bpy.ops.wm.read_factory_settings(use_empty=True)
+    bpy.ops.object.select_all(action='SELECT')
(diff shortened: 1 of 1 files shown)


emackey:
Possibly will fix #1183?

emackey:
I believe this change is harmless: It's only the tests, and CircleCI starts with a clean image each time anyway.  Developers with weird Blender settings might have their ability to run local test impacted (both positively and/or negatively, depending what they've set as their defaults), but in general this should give them more local control, not less.  Users don't run this code.

Issue #1175, opened 2020-08-11
Errors during python script development require a restart


spiraloid:
**Describe the bug**
I am writing python script to export my models as GLTF.  There are frequently errors of one kind or another.  As I fix them, I can generally reload all scripts and proceed with normal development.  However, anything do with the GLTF exporter is frequently plagued by assert errors that are require blender to be restarted to resolve.   Perhaps whatever is happening during the restart to rest the GLTF exporter could also happen during reload all scripts? 

**To Reproduce**

**Expected behavior**

**Screenshots**

**.blend file/ .gltf**
If applicable, a zipped folder containing a .blend file for exporter issues and a .gltf file for importer issues. Please take care to include all resources, such as textures in the .zip file.

**Version**
 - OS: [e.g. macOS, linux]
 - Blender Version [e.g. 2.82, 2.83, or build date]

**Additional context**

scurest:
Instructions to reproduce?

spiraloid:
still getting this.  it's so weird.  I get it often.   restarting and loading the same file and running the same script fixes it. 

seams related to reloading scripts after an error with another script during development.

RuntimeError: Error: Python: Traceback (most recent call last):
  File "C:\Program Files\Blender Foundation\Blender 2.92\2.92\scripts\addons\io_scene_gltf2\__init__.py", line 575, in execute
    return gltf2_blender_export.save(context, export_settings)
  File "C:\Program Files\Blender Foundation\Blender 2.92\2.92\scripts\addons\io_scene_gltf2\blender\exp\gltf2_blender_export.py", line 46, in save
    json, buffer = __export(export_settings)
  File "C:\Program Files\Blender Foundation\Blender 2.92\2.92\scripts\addons\io_scene_gltf2\blender\exp\gltf2_blender_export.py", line 66, in __export
    json = __fix_json(exporter.glTF.to_dict())
  File "C:\Program Files\Blender Foundation\Blender 2.92\2.92\scripts\addons\io_scene_gltf2\io\com\gltf2_io.py", line 1204, in to_dict
    self.materials)
  File "C:\Program Files\Blender Foundation\Blender 2.92\2.92\scripts\addons\io_scene_gltf2\io\com\gltf2_io.py", line 56, in from_union
    assert False
AssertionError

PR #1186, opened 2020-08-22
Export: more descriptive exception when from_union fails

scurest:
Replaces `assert False` with `raise TypeError(x)` so the argument `x` will show up in the trace.

The stack trace to this assert is not currently very useful in diagnosing errors, see eg. most recently #1175. Hopefully this will help.


julienduroure:
gltf2_io.py is automatically generated from glTF specification schema, so I propose to store any hand modification on it somewhere, to be able to re-apply them if we have to generate this file again. (This should happen probably only for any new version of the specification)

scurest:
This function has also been hand-modified already by 6d63492a94aa8c3bbb44f2fe6e4488fbfcc08a4f to add the trace printing stuff.

scurest:
OTOH I don't really care about this so I don't mind closing lol.

PR #1187, opened 2020-08-22
Permit missing texture.source

scurest:
The texture.source property is not required, but the importer barfs if it's missing. gltf2_blender_texture [handles](https://github.com/KhronosGroup/glTF-Blender-IO/blob/fdce5bcf687f3d0dd9b0ec5a4e74faf9f267745c/addons/io_scene_gltf2/blender/imp/gltf2_blender_texture.py#L46) the None case, it's just that the gltf2_io classes won't parse it correctly in from_dict. This fixes it.

I also changed to_dict to permit None here. There was a comment: `# most viewers can't handle missing sources` (ironic, since that included us). But the gltf2_io class is not the right place to enforce this. It should be done in gather_texture, which it [is](https://github.com/KhronosGroup/glTF-Blender-IO/blob/fdce5bcf687f3d0dd9b0ec5a4e74faf9f267745c/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_texture.py#L49-L51).

### DIFF
--- a/addons/io_scene_gltf2/io/com/gltf2_io.py
+++ b/addons/io_scene_gltf2/io/com/gltf2_io.py
@@ -1117,7 +1117,7 @@ class Texture:
         extras = obj.get("extras")
         name = from_union([from_str, from_none], obj.get("name"))
         sampler = from_union([from_int, from_none], obj.get("sampler"))
-        source = from_int(obj.get("source"))
+        source = from_union([from_int, from_none], obj.get("source"))
         return Texture(extensions, extras, name, sampler, source)
 
     def to_dict(self):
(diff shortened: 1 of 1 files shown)


julienduroure:
gltf2_io.py is automatically generated from glTF specification schema, so I propose to store any hand modification on it somewhere, to be able to re-apply them if we have to generate this file again. (This should happen probably only for any new version of the specification)

scurest:
gltf2_io has already been hand-modified extensively. In fact this just removed the hand-modification done in cda46517011a0502ef7733cf48c614b149cc3411.

julienduroure:
Well, I forgot that this was already modified ... one more reason to keep an history of changes. But maybe git history is enough

scurest:
That's how I noticed it was broken.

scurest:
Any PRs going to be merged soon?

PR #1201, opened 2020-09-09
⚡️ Drastically improve performance when creating bufferViews by using…

m0ppers:
… bytearrays

I am exporting big scenes (several hundred objects - resulting glb is 800MB+) and I finally found time to investigate why performance was so bad in my case.

Here is a simple change that boosts performance in my case.

I couldn't quite believe how much performance was gained.

Before:

`INFO: Finished glTF 2.0 export in 3073.9336087703705 s`

After:

`Finished glTF 2.0 export in 59.18472361564636 s`

Resulting file has same sha1 and is fully working :S

I guess this should help others with big scenes as well.

### DIFF
--- a/addons/io_scene_gltf2/io/exp/gltf2_io_buffer.py
+++ b/addons/io_scene_gltf2/io/exp/gltf2_io_buffer.py
@@ -22,21 +22,23 @@ class Buffer:
     """Class representing binary data for use in a glTF file as 'buffer' property."""
 
     def __init__(self, buffer_index=0):
-        self.__data = b""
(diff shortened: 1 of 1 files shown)


m0ppers:
unsure why CI fails :S

```
Looking for: Blender 2.90

Exited with code exit status 1

CircleCI received exit code 1
```

julienduroure:
Hello,
Master is now future 2.91, so we needed to update test version. I just merged the PR.
Can you please rebase with master, it should solve the issue?

julienduroure:
Thanks!

Issue #1189, opened 2020-08-26
Export/UI: "Animation" section of export screen should be reorganized


scurest:
The "Animation" panel when exporting looks like this

![animpanel](https://user-images.githubusercontent.com/11024420/91360922-ef20e600-e7bc-11ea-8e79-845b50621c05.png)

Shape keys and skinning have nothing to do with animation so they shouldn't be under the "Animation" section. And having a subsection of "Animation" that is also named "Animation" is dumb.

donmccurdy:
How about a new "Armatures" section, and moving Shape Keys under "Geometry" section? On that note, I think the Materials/Images options might make more sense under "Include" than "Geometry".

scurest:
Mockup:
![out](https://user-images.githubusercontent.com/11024420/91375970-a6c7ef00-e7e1-11ea-97b7-364ded42e35f.png)

scurest:
What does "Use Current Frame" do?

donmccurdy:
The mockup looks great to me. :)

julienduroure:
Not sure about "Export Deformation bone only", if this should be in Animation or in Armature

Issue #1081, opened 2020-05-30
Materials that are defined but not referenced will not be instantiated.


XanTheDragon:
When importing a GLB or glTF file, only materials that are referenced by a primitive will actually be instantiated in Blender.

Reproduction of this error is easy. The model I was testing and have included below contains 8 textures (each with its own material, so 8 materials), but only one of them is used by the primitive due to the limits of the glTF format and only supporting one material per primitive. Each texture is intended to be used as a variant or alternate skin, which is why they are included despite not being used immediately.

When importing, I expect all of the materials and textures defined in the glTF JSON data to be instantiated even if they have no users. This is not the case. In a personal case, I would not bother to define the materials and include the texture data if I had no plans to use them. I have included this data due to an explicit need for them to be present when importing.

**Screenshots**
Here's some cleanly formatted JSON showing off the material definitions, alongside the model after a fresh import into Blender.
![image](https://user-images.githubusercontent.com/10490540/83315870-58a64e00-a1e8-11ea-9f6f-dff525337126.png)

**.blend file/ .gltf**
All textures are embedded into the binary. This file should be set to go on its own.
[sk_gremlin.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4704876/sk_gremlin.zip)

**Version**
 - OS: Windows 10
 - Blender Version: 2.82a

scurest:
This is easy to do (not sure whether to put the code in BlenderScene or BlenderGlTF though).

<details>
<summary>Tentative Patch</summary>

````diff
diff --git a/addons/io_scene_gltf2/blender/imp/gltf2_blender_scene.py b/addons/io_scene_gltf2/blender/imp/gltf2_blender_scene.py
index 6119eb9..06947e5 100644
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_scene.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_scene.py
@@ -15,6 +15,8 @@
 import bpy
 
 from .gltf2_blender_node import BlenderNode
+from .gltf2_blender_image import BlenderImage
+from .gltf2_blender_material import BlenderMaterial
 from .gltf2_blender_animation import BlenderAnimation
 from .gltf2_blender_vnode import VNode, compute_vnodes
 
@@ -39,6 +41,8 @@ class BlenderScene():
         gltf.display_current_node = 0  # for debugging
         BlenderNode.create_vnode(gltf, 'root')
 
[...]

PR #1206, opened 2020-09-12
export material: new option to keep placeholders

### LINKED ISSUE #1205: Add ability to export meshes with defined but empty material slots
I can't properly configure my meshes in Godot when I export them from Blender without materials.

An option to export empty placeholder material slots instead of the real ones would be the perfect solution.
I could then populate those slots properly with my customized Godot materials later on.

Other alternatives would require Godot doing the work of discarding real materials and only keeping the slots.
But I see this feature in the Blender exporter better.

julienduroure:
Fix #1205 

This PR change "export material" option from boolean (export / not export) to an enum: 
* export
* not export
* placeholder

Not export merges all primitives (as it is done on current master)
Placeholder keep primitives, but does not create a material. This is working for empty slots or filled slots

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -217,12 +217,11 @@ class ExportGLTF2_Base:
     export_materials: EnumProperty(
         name='Materials',
         items=(('EXPORT', 'Export',
-                'Export materials '),
-                ('PLACEHOLDER', 'Placeholder',
-                'Do not export materials, but keep placeholder primitives '),
-                ('NONE', 'No export',
-                'Do not export materials '),
-               ),
+        'Export all materials used by included objects'),
+        ('PLACEHOLDER', 'Placeholder',
+        'Do not export materials, but write multiple primitive groups per mesh, keeping material slot information'),
+        ('NONE', 'No export',
+        'Do not export materials, and combine mesh primitive groups, losing material slot information'),
         description=(
             'Export materials '
         ),
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -221,10 +221,8 @@ class ExportGLTF2_Base:
         ('PLACEHOLDER', 'Placeholder',
         'Do not export materials, but write multiple primitive groups per mesh, keeping material slot information'),
         ('NONE', 'No export',
-        'Do not export materials, and combine mesh primitive groups, losing material slot information'),
-        description=(
-            'Export materials '
-        ),
+        'Do not export materials, and combine mesh primitive groups, losing material slot information')),
+        description='Export materials ',
         default='EXPORT'
     )
 
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -214,10 +214,19 @@ class ExportGLTF2_Base:
         default=False
     )
 
-    export_materials: BoolProperty(
+    export_materials: EnumProperty(
         name='Materials',
-        description='Export materials',
-        default=True
+        items=(('EXPORT', 'Export',
+                'Export materials '),
+                ('PLACEHOLDER', 'Placeholder',
+                'Do not export materials, but keep placeholder primitives '),
+                ('NONE', 'No export',
+                'Do not export materials '),
+               ),
+        description=(
+            'Export materials '
+        ),
+        default='EXPORT'
     )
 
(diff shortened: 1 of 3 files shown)


addons/io_scene_gltf2/__init__.py
```diff
@@ -672,7 +681,7 @@ def draw(self, context):
         layout.prop(operator, 'export_colors')
         layout.prop(operator, 'export_materials')
         col = layout.column()
-        col.active = operator.export_materials
+        col.active = operator.export_materials == "EXPORT"
```
donmccurdy, line 684: Per https://github.com/KhronosGroup/glTF-Blender-IO/issues/1189, could we move this to the *Include* section?

addons/io_scene_gltf2/__init__.py
```diff
@@ -214,10 +214,19 @@ def __init__(self):
...
         name='Materials',
-        description='Export materials',
-        default=True
+        items=(('EXPORT', 'Export',
+                'Export materials '),
+                ('PLACEHOLDER', 'Placeholder',
+                'Do not export materials, but keep placeholder primitives '),
+                ('NONE', 'No export',
+                'Do not export materials '),
+               ),
```
donmccurdy, line 225: Suggested naming:

- **All**
  - _Include all materials, whether used or unused_
- **Active** (default) 
  - _Include all materials used by at least one mesh_
- **Placeholder**
  - _Include only material names, without properties or textures_
- **None**
  - _Do not export materials_

The "All" option does not need to be included in this PR, but could be added in the future if we decide to support https://github.com/KhronosGroup/glTF-Blender-IO/issues/1081.

addons/io_scene_gltf2/__init__.py
julienduroure: Yes, this enum replaces the checkbox, that is already in "include" section in the mockup

addons/io_scene_gltf2/__init__.py
julienduroure: Currently, "placeholder" doesn't export material names (and it can't where there is an empty slot). The only difference with "None" is that None merged all primitives into a single one, where "Placeholder" will keep all primitives created

But OK with the concept of a future "All" option for #1081

donmccurdy: Oops, I see. In that case... I think this is a useful feature to have, but I would not create a material dropdown for it. Or at least not yet, perhaps if we add the "All" option later. How about a checkbox in the geometry section instead? Perhaps:

- **Merge vertex groups** (default = true)
  - When enabled, vertex groups sharing a material are merged for more efficient export.

Or perhaps I'm still misunderstanding. But I do like using Blender's term "vertex groups" rather than glTF's term "primitives", which users may not know.

donmccurdy: Or maybe better:

- **Vertex groups** (default = FALSE)
  - When enabled, vertex groups within a mesh are preserved, whether or not they require separate materials

scurest: This isn't related to vertex groups.

julienduroure: Confirmed, this is not linked to Vertex Group.

addons/io_scene_gltf2/__init__.py
donmccurdy: Yes, it's under the "include" section in the mockup, but not the code yet — let's make this change in code here, so that the new dropdown element is in its intended location from the beginning.

addons/io_scene_gltf2/__init__.py
donmccurdy: Ok, I see. In that case, I can't tell from the description included for this UI option what the option does. Could you suggest a description that would help users to understand when to choose this?

julienduroure:
New proposition :

items=(('EXPORT', 'Export',
                'This will export all materials used by exported object '),
                ('PLACEHOLDER', 'Placeholder',
                'Do not export materials, but meshes will be exported in multiple sub-meshes (primitives), keeping material slot information '),
                ('NONE', 'No export',
                'Do not export any materials. Mesh will be exported without sub-meshes, loosing material slot information '),
               ),

donmccurdy:
@julienduroure I rephrased a little and corrected a typo — if my update is OK with you, +1 here.

julienduroure:
I updated descriptions.
Let me know if everything is OK now

donmccurdy (approved):

PR #1209, opened 2020-09-14
Export: support KHR_materials_unlit

### LINKED ISSUE #554: Unlit materials do not round-trip

scurest:
Fixes #554.  
See also #800.  
cc @donmccurdy

Adds a separate code path for KHR_materials_unlit. The exporter only goes down this codepath if it detects a "shadeless" material graph. This path is cleanly separated off from all the existing functions which don't need to change.

There was a lot of uncertainty about how exactly to detect this. This PR is very conservative. It requires the graph have a specific "shape", namely:

unlit-nodes

This is basically what the importer creates. Most of the recommendations for a ["shadeless" material on the Blender StackExchange](https://blender.stackexchange.com/search?q=shadeless) will also match this pattern. Note that an Emission shader is *not* required. In particular, the simplest graph is just an RGB node plugged into the Material Output.

Also note that all the code for detecting this is all in the function `detect_shadeless_material`, so the detection is very easy to change.

If a Background socket is detected anywhere in the graph, then the graph also counts as unlit. I left this in even though I don't think you should really have a Background node in an Object (as opposed to World) shader graph.

----

[Here](https://github.com/KhronosGroup/glTF-Blender-IO/files/5216830/UnlitTest.zip) is the .blend I used for tests.

tests

Each cube has a different material graph. The exported .glb roundtrips successfully. I also tested a few unlit models I had around and they all roundtripped successfully too.

-----

This also includes a fix for the annoyance of mesh_double_sided being passed to gather_material (remnant of 2.79 support).

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials_pbr_metallic_roughness.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials_pbr_metallic_roughness.py
@@ -61,8 +61,6 @@ def __gather_base_color_factor(blender_material, export_settings):
         base_color_socket = gltf2_blender_get.get_socket(blender_material, "BaseColor")
     if base_color_socket is None:
         base_color_socket = gltf2_blender_get.get_socket_old(blender_material, "BaseColorFactor")
-    if base_color_socket is None:
-        base_color_socket = gltf2_blender_get.get_socket(blender_material, "Background")
     if isinstance(base_color_socket, bpy.types.NodeSocket):
         rgb = gltf2_blender_get.get_factor_from_socket(base_color_socket, kind='RGB')
 
@@ -81,8 +79,6 @@ def __gather_base_color_texture(blender_material, export_settings):
         base_color_socket = gltf2_blender_get.get_socket(blender_material, "BaseColor")
     if base_color_socket is None:
         base_color_socket = gltf2_blender_get.get_socket_old(blender_material, "BaseColor")
-    if base_color_socket is None:
-        base_color_socket = gltf2_blender_get.get_socket(blender_material, "Background")
 
     alpha_socket = gltf2_blender_get.get_socket(blender_material, "Alpha")
     if alpha_socket is not None and alpha_socket.is_linked:
--- a/tests/test/test.js
+++ b/tests/test/test.js
@@ -951,6 +951,29 @@ describe('Importer / Exporter (Roundtrip)', function() {
                 assert.equalEpsilon(pbr.roughnessFactor, 0.75);
             });
 
+            it('roundtrips unlit base colors', function() {
+                let dir = '01_unlit';
+                let outDirPath = path.resolve(OUT_PREFIX, 'roundtrip', dir, outDirName);
+                let gltfPath = path.resolve(outDirPath, dir + '.gltf');
+                const asset = JSON.parse(fs.readFileSync(gltfPath));
+
+                assert.strictEqual(asset.materials.length, 2);
+
+                const orange = asset.materials.find(mat => mat.name === 'Orange');
+                assert.ok('KHR_materials_unlit' in orange.extensions);
+                assert.equalEpsilon(orange.pbrMetallicRoughness.baseColorFactor[0], 1);
+                assert.equalEpsilon(orange.pbrMetallicRoughness.baseColorFactor[1], 0.217637640824031);
+                assert.equalEpsilon(orange.pbrMetallicRoughness.baseColorFactor[2], 0);
+                assert.equalEpsilon(orange.pbrMetallicRoughness.baseColorFactor[3], 1);
+
+                const blue = asset.materials.find(mat => mat.name === 'Blue');
+                assert.ok('KHR_materials_unlit' in blue.extensions);
+                assert.equalEpsilon(blue.pbrMetallicRoughness.baseColorFactor[0], 0);
+                assert.equalEpsilon(blue.pbrMetallicRoughness.baseColorFactor[1], 0.217637640824031);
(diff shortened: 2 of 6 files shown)


donmccurdy:
Thanks! Does "any color socket" also include textures? Or an RGB node connected directly to the Surface output? From the code that appears to be handled, but just checking.

scurest:
Yes, it's any socket with a yellow dot (like yellow=color, gray=value, green=shader, blue=vector, ...).

tests/test/test.js
```diff
@@ -936,6 +936,27 @@ describe('Importer / Exporter (Roundtrip)', function() {
...
+                assert.equalEpsilon(orange.pbrMetallicRoughness.baseColorFactor[0], 1);
+                assert.equalEpsilon(orange.pbrMetallicRoughness.baseColorFactor[1], 0.217637640824031);
+                assert.equalEpsilon(orange.pbrMetallicRoughness.baseColorFactor[2], 0);
+                assert.equalEpsilon(orange.pbrMetallicRoughness.baseColorFactor[3], 1);
+
+                const blue = asset.materials.find(mat => mat.name === 'Blue');
+                assert.equalEpsilon(blue.pbrMetallicRoughness.baseColorFactor[0], 0);
+                assert.equalEpsilon(blue.pbrMetallicRoughness.baseColorFactor[1], 0.217637640824031);
+                assert.equalEpsilon(blue.pbrMetallicRoughness.baseColorFactor[2], 1);
+                assert.equalEpsilon(blue.pbrMetallicRoughness.baseColorFactor[3], 0.5);
```
emackey, line 957: Can this test also check for the presence of the unlit extension itself?

scurest: Ah, that would be important....

emackey:
With this change, does any use of the Emission BSDF node result in an unlit material?  I think we want some way to distinguish emission from unlit, particularly given that the emission node has a strength value.  Emission strength may be coming to the Principled BSDF node in [D4971](https://developer.blender.org/D4971), but that's not merged as of this writing.

tests/test/test.js
scurest: How does that look? I used `assert.ok('KHR_materials_unlit' in mat.extensions)`.

scurest:
Well, not any but like this, yeah.

ms

I see what you mean that people would expect it to export as emission.

How about only accepting an emission node when there is also the lightpath trick to disambiguate it as "unlit"? I do want to allow emission there since it's a pretty popular setup (eg. [1](https://blender.stackexchange.com/questions/148336/blender-2-8-shadeless-material-setup-not-shadeless) [2](https://blender.stackexchange.com/questions/21533/totally-white-shadeless-material-in-cycles/21535#21535)) and the importer does it like that.

You could still forgo the lightpath trick by plugging a Color socket straight into the Shader node.

tests/test/test.js
emackey: New test looks great, thanks!

emackey:
Yeah.  The "lightpath trick" looks like it has it only affect direct camera rays, not shadows and bounces and stuff?  That sounds good to me.

scurest:
Okay, I amended it in. It now detects this. The link on the left can go to any color socket (or you can put a color directly in the Emission node of course)

tree

I also removed handling of emission strength for unlit.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials_pbr_metallic_roughness.py
```diff
@@ -58,8 +58,6 @@ def __gather_base_color_factor(blender_material, export_settings):
         base_color_socket = gltf2_blender_get.get_socket(blender_material, "BaseColor")
     if base_color_socket is None:
         base_color_socket = gltf2_blender_get.get_socket_old(blender_material, "BaseColorFactor")
-    if base_color_socket is None:
-        base_color_socket = gltf2_blender_get.get_socket(blender_material, "Background")
```
emackey, line 62: @donmccurdy Are you OK if we drop backwards compatibility to "Background" unlit nodes at the same time as we embrace the new system?

emackey (approved):
Other than my question for Don, I'm good with this.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials_pbr_metallic_roughness.py
scurest: Background node support is retained, it's just moved off the main path to the unlit path. In master, if `gltf2_blender_get.get_socket(blender_material, "Background")` returns non-None then the material is unlit.

https://github.com/KhronosGroup/glTF-Blender-IO/blob/2423afdec5f3aeed7637638c878b4736f8773b77/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py#L157-L158

In this PR, detect_shadeless_material does the same check

https://github.com/KhronosGroup/glTF-Blender-IO/blob/4e67a996bcc3f816854731c920b27e45418a088f/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials_unlit.py#L28-L30

The reason those lines are deleted is they will never run now; if there had been a Background node, it would have gone down the unlit path instead.

I tested this setup and it exports as unlit with this branch, same as master

unli

donmccurdy:
Looks good!

PR #1179, opened 2020-08-13
Export: don't mess with normals when doing Apply Modifiers

scurest:
Fixes #877.  
Fixes #975.

Deletes the code that applies a temp edge split modifier and messes with use_auto_smooth when "Apply Modifiers" is used. As far as I could tell these are no longer needed. The OBJ/FBX exporter also don't appear to do anything like this. If anyone has a model that this code was supposed to fix it would be great to test it.

PR #1212, opened 2020-09-16
Export: stop caching by name; use a custom cache for meshes

### LINKED ISSUE #1210: Text not exported correctly
**Describe the bug**
Exporting text objects that are parented to an empty get exported weirdly. Some are transformed, some left out on export.
I think the transforms should be applied before exporting.

**Expected behavior**
Blender Scene and Three.js viewer shoul produce the same outcome.

**Screenshots**
image
image

**.blend file/ .gltf**

[Convert_Text_Bug.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/5217615/Convert_Text_Bug.zip)

**Version**
 - OS: Windows
 - Blender Version 2.9

**Additional context**
First of all, thanks for working on that topic. I programmed an add-on to convert curves to meshes to be able to export annotations to gltf. Maybe with your changes I don't need it anymore in the future.

image

### LINKED ISSUE #1211: Export: when linking meshes with the same name, exporter can't tell them apart
Found in #1210.

If you have two meshes with the same name (from linking), the exporter thinks they are the same. Like I said in #1210, the library parameter to gather_mesh should presumably distinguish these since they have different `mesh.library`s, but the library parameter comes from `object.library`, not `mesh.library`.

**To Reproduce**
1. Create an empty .blend
2. Add > Cube
3. Rename the mesh "Cube" to "M"
4. Save as A.blend
5. Create an empty .blend
6. Add > UVSphere
7. Rename the mesh "Sphere" to "M"
8. File > Link and select A.blend/Mesh/M
9. Export glTF

out

scurest:
Fixes #1210.  
Fixes #1211.  
Helps with, but does not fix, #1203.

`@cached` functions no longer cache any datablocks by their name. This prevents us from confusing datablocks with the same name.

However we don't want to cache on a temporary mesh for gather_mesh. We want to _pass_ a mesh, but we want to _cache_ on blender_object.data. This makes it hard to use `@cached` so I just used a manually implemented cache for the mesh/primitive stuff now. The cache key for a mesh is computed in get_mesh_cache_key. The Blender object is now passed around the mesh/primitive functions and to_mesh has been pushed down the call tree to extract_primitives.

This is cleaner is some ways too: the function signatures are less overburdened with a zillion parameters, the caching logic is more localized, and hacks like skip_filter aren't needed.

I did not bother to port over anything removed by #1179.


julienduroure:
I remember adding caching by name because some datablock are not hashable and caching was wrong.
Seems blame the file give me the commit, but not the corresponding fixed issue, need to find an example.

scurest:
It was added in #547, which also changed the mesh/primitive functions.

If it doesn't work to hash on the datablock itself, `cached` could also hash on

````py
(
    data.name,
    data.rna_type.name,
    None if not data.library else data.library.name,
)
````

scurest:
I tested instancing.blend from #526 and it still works.

scurest:
I refactored this a bit. Instead of a manually implemented cache, there's a new decorator

````py
@cached_by_key(key=key_func)
def func(..., export_settings):
    ....
````

This works like `@cached` but you can choose what the cache key to use is with `key_func`. I reimplemented `@cached` in terms of `@cached_by_key`.

I also moved the cache from being stored on `func.__cache` to being stored on `export_settings[('cached_func', func)]` so the stale caches don't stick around after the export is finished.

julienduroure:
Hello,

Removing cache by name for bpy.type let me anxious and fearful about regression.
Caching by name is not an historical code, and was added after we discover an issue with caching directly the blender type.

Two possibilities: 
* Blender API changes, and caching by bpy.types is now ok
* Blender API didn't change, and caching is still a problem

I propose to solve this question before continuing working in this PR, looking into my history test cases etc.., comparing old blender version, on so on.

Maybe you can create an independent PR about moving func.__cache to export_settings
Making small PR in so central and crucial part of the code will help reviewing and test

scurest:
> Maybe you can create an independent PR about moving func.__cache to export_settings

Okay, I dropped that.

> I propose to solve this question before continuing working in this PR, looking into my history test cases etc.., comparing old blender version, on so on.

Do you remember why you added it? Like I said, it was added in #547, but there is no explanation given for it either in the PR text or the commit message. I already tested all the issues mentioned by #547 and the file in #526 and they did not regress.

The sequence of events AFAIUI was

* Before #526 gather_mesh took a Mesh as a parameter.
* #526 changed gather_mesh to add the Object as a parameter for fetching materials
* This breaks caching, so it was changed to pass down a tuple of materials instead. At the same time it switched from passing a Mesh to passing the mesh's name. pjoe commented

    > NOTE: from debugging I noticed that cache_wrapper does NOT work properly with bpy.types objects, which is the reason I had to change several places from passing around blender_mesh: bpy.types.Mesh to blender_mesh_name: str. This issue with cache_wrapper is likely to cause problems elsewhere :S

    It does not say what the issue was however. Maybe he was referring to the fact it would cache on the temp mesh returned by `to_mesh` when exporting with "Apply Modifiers"?
* You [quoted this](https://github.com/KhronosGroup/glTF-Blender-IO/pull/526#issuecomment-501856969), observing that it had broken modifier. I don't know why it broke modifiers though.
* So #547 fixed this by changing `cached` to cache by name for some bpy types, and then changing gather_mesh to pass around a Mesh again instead of a mesh name.

julienduroure:
Adding a next step:

* I found a case where posebone caching has issue too, so adding posebone to cache too. #950 

About theses changes:

> cache_key in func.__cache

sometime returns False, even if the cache_key seems to be in keys()
Seems to happen not always, and we found that bpy.type are linked to this miss-cache.

So ... we need to reproduce it now, at least with the corresponding blender version

scurest:
Okay, well since I don't understand the issues involved with caching by-name vs by-value and why it was changed, I'll drop this.

Issue #372, opened 2018-06-10
ACCESSOR_NON_UNIT error in validator


AlbertoElias:
I'm exporting a blender scene using the latest add on in master and then opening it in https://gltf-viewer.donmccurdy.com/ and I get the error 

`Cannot read property 'joints' of undefined`

The model has an armature and some animations. I'm attaching the gltf and the blender file. Thanks!

[broken_model.zip](https://github.com/KhronosGroup/glTF-Blender-Exporter/files/2088037/broken_model.zip)

julienduroure:
Skin issue seems to be fixed, but now have gltf error : 
> ACCESSOR_NON_UNIT

emackey:
> Skin issue seems to be fixed, but now have gltf error :
> 
> > ACCESSOR_NON_UNIT

I know that one.  The model contains degenerate faces, and the exporter doesn't weed them out prior to export, resulting in zero-length normal vectors, which are invalid in glTF.

Quick fix is go into Edit Mode, menu, "Mesh" -> "Clean up" -> "Degenerate Dissolve".

julienduroure:
Well, this is a duplicate of #1138, let's continue the discussion on the other ticket

PR #1214, opened 2020-09-16
Export: replace zero normals with UP

### LINKED ISSUE #1123: T78926 Materials export blank if unconnected Principled BSDF node is anywhere in a "Background" material during gltf 2.0 export
See https://developer.blender.org/T78926

### LINKED ISSUE #1138: T79104: Exporting a valid GLTF (was valid before import) causes GLTF validation errors
See https://developer.blender.org/T79104
On 2.83

scurest:
Fixes #1138.

Zero normals are turned into the up vector. Since this seems to happen only on degenerate tris, it shouldn't have any visible effect, but it makes the glTF validate. 

I tried to make a test case, but wasn't able to. I couldn't figure out what causes zero normals. Just being degenerate is not enough, I assume you need some condition on the topology too. Also the file from #372 is working in master with no zero normals right now.

So the only file I know of with this problem is the one from #1138. It now passes validation.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
@@ -348,6 +348,12 @@ def __get_normals(blender_mesh, key_blocks, armature, blender_object, export_set
             ns[:] = __apply_mat_to_all(normal_transform, ns)
             __normalize_vecs(ns)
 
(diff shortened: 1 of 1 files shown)


donmccurdy:
#1123 doesn't look like the right issue, typo?

scurest:
My bad, it was supposed to be #1138. (If I type "#1138<period><enter>" Github auto-replaces it with 1123).

PR #1215, opened 2020-09-16
Export: fix for texCoord/texture transform

### LINKED ISSUE #1037: Export: When using multiple UVs on the same texture, all UVs are set to the first.
**Describe the bug**
It's currently not possible to use multiple UVs on the same texture.

I'm using one texture, red channel for AO, green channel for roughness and blue channel for metallic.
Metallic and roughness use **UV1**, and AO uses **UV2**.
When exporting, AO is using **UV1** as well.

When changing the AO to **another texture**, the export is perfect.

**Screenshots**
blender_PA8rstia1M
Code_C0TtvWqfNQ

**.blend file/ .gltf**
[cubes.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4500354/cubes.zip)

**Version**
 - OS: Windows 10
 - Blender Version: 2.82.7

scurest:
Fixes #1037.

Adds a new parameter to gather_texture_info, primary_socket. This is the socket the texCoord/texture transform should come from. So for example, for occlusion and metal-rough, if the blender_shader_sockets are (o,r,m) both times, the primary_socket will be o vs r/m to distinguish which one we want.

For cases where we have to combine like RGB+A and R+M, there are multiple possible primary sockets. This just picks ones arbitrarily.

Tested with multi-uv.zip from https://github.com/KhronosGroup/glTF-Blender-IO/issues/1037#issuecomment-627359040. Also tested different UVMaps with clearcoat, and different texture transforms with ORM.

------

I added a simple roundtrip test file for this. It's just one texture, two UVMaps.

testo


julienduroure:
Looks good to me. Thanks!

Issue #1216, opened 2020-09-17
Support new Emission Strength on Principled BSDF


emackey:
A new Emission Strength input was added to the Principled BSDF node in [D4971](https://developer.blender.org/D4971).

Emission strength from the original Emission BSDF node was added in #1159.  It should be updated to include the new strength input as well.

emackey:
Ah, correction: #1159 already [includes code to look for emission strength](https://github.com/KhronosGroup/glTF-Blender-IO/pull/1159#discussion_r464088810) on Principled BSDF.  We should add test(s) for this once we get a daily build with the new input added.

julienduroure:
So, seems exporter already manage it. 
Importer still need to be updated?

scurest:
Emission Strength is in the latest build. I can confirm exporting works.

The importer can use the strenght socket for grayscale emissiveFactors at least.

astrand130:
Hi! Author of the Emission Strength Socket patch.

https://developer.blender.org/D8777 is needed to access the socket via python using the Principled BSDF wrangler. It is in review but the API should remain the same.

You can find the WIP updated OBJ importer here which contains code to handle importing/exporting emission strength by doing a sort of RGBM encoding/decoding. https://developer.blender.org/D8868

Thanks

julienduroure:
Hello @astrand130 , current implementation of glTF is not using node_shader_utils wrangler.

Exporting is already manage since yesterday, and importer will be via #1217

julienduroure:
Import part is now merge. Closing

PR #1217, opened 2020-09-18
Import: import grayscale emissiveFactor as Emission Strength

scurest:
For the importer side of #1216.

When there is an emissiveTexture and the emissiveFactor is grayscale (R==G==B), put the emissiveFactor into the strength socket. This doesn't actually matter to the glTF exporter, but it will help exporters that use node_shader_utils, and it makes the graph cleaner.

When the emissiveFactor is not grayscale, it still goes in a multiply node.

I also bumped the addon's required Blender version since the importer now requires the latest version of Blender. (You might object, "Yesterday's build was also (2,91,0), and it didn't have an Emission Strength socket". So I checked what OBJ and FBX did, and they didn't bump it at all, so....)


julienduroure:
Looks good to me.
Can you please rebase?

julienduroure (approved):

scurest:
Rebased.

PR #1225, opened 2020-09-25
Import/Export: fixes for material_index/material slots

### LINKED ISSUE #1224: Importing meshes without materials causes exporting issues
**Describe the bug**
When importing a scene with a mesh that doesn't contain materials, the add-on ignores the first material object linked to the mesh when exporting the scene to GLTF. If you add a single texture/material to the imported mesh, it won't show up in the exported file.

**To Reproduce**
1. Import a GLTF scene that contains a mesh with UV coordinates but no materials (import configurations don't matter);
2. Add a single material to the imported mesh;
3. Export the scene to GLTF (make sure to include materials, images, and UV coordinates);
4. The exported mesh should contain no materials whatsoever, even though we added it in Blender.

**Expected behavior**
It's expected that the output GLTF file contains the material added in Blender.

**Files**
[Here](https://gist.github.com/gabriel-dev/04dd558deaf4d65fc9e03f0b0b9cbbf8) is a GLTF file with a mesh without materials to reproduce the issue.

**Version**
- Blender:
    - version: 2.90.0, branch: master, commit date: 2020-08-31 11:26, hash: 0330d1af29c0, type: Release
    - build date: 2020-08-31, 10:00:13
- Platform: Windows 10 (2004 - 19041.508)

**Additional context**
After some debugging, it seems that the problem comes down to the following code ([check the original code here](https://github.com/KhronosGroup/glTF-Blender-IO/blob/3de6de2b995028ccd62e0fd63feacc71a105d270/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py#L327)):
```python
# gltf2_blender_mesh.py
def create_mesh(gltf, mesh_idx, skin_idx):
    ...
    # Assign materials to faces

    # Initialize to no-material, ie. an index guaranteed to be OOB for the
    # material slots. A mesh obviously can't have more materials than it has
    # primitives...
    oob_material_idx = len(pymesh.primitives)
    material_indices = np.full(num_faces, oob_material_idx)
```
The `oob_material_idx` variable is initialized with `len(pymesh.primitives)`, which, in the case of a simple mesh, is `1`. When the imported GLTF meshes are linked to some material, the loop just below the code snippet sets `material_indices` with the right values. However, when the imported meshes don't have a material, `oob_material_index` stays equal to `1`. This causes the first material (index zero) added to the mesh through Blender not to be found, thus causing the issue when exporting the scene to GLTF.

Wouldn't it be the case to default it to `0` (first material), which is the default value for [`MeshPolygon.material_index`](https://docs.blender.org/api/current/bpy.types.MeshPolygon.html?highlight=material_index#bpy.types.MeshPolygon.MeshPolygon.material_index)? After manually changing it to `0`, everything works as expected, though I didn't do extensive tests to cover other scenarios.

scurest:
Fixes #1224.

Exporter:

* any material slots at or after an empty material slot would be screwed up (they are skipped [here](https://github.com/KhronosGroup/glTF-Blender-IO/blob/5d976d3642b5ce885719a29976b83f8d6562713b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py#L317), which screws up the indices (and caching)). This changes material_names to have a None for empty slots. Anything assigned to an empty slots will now be exported with `material: None`.
* OOB material_indices exported as `material: None`. This changes to use the last slot (if it exists), which is what Blender does.

Importer: 

* if no primitive in a mesh has materials, the `# Assign materials` block is now skipped, which gives every poly a material_index of 0, and no material slots.
* Otherwise, an empty material slot will be created and used for prims that have `material: None`.

IOW we only import with an empty material slot when necessary.

-----

In addition to #1224, here's another example of a bug with empty slots:

good

In master, export->reimport results in

bad

It's fixed on this branch.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
@@ -314,7 +314,8 @@ def __gather_mesh(blender_object, library, export_settings):
                 vertex_groups = None # Not needed if no armature, avoid a cache miss
                 modifiers = None
 
-    material_names = tuple([ms.material.name for ms in blender_object.material_slots if ms.material is not None])
+    materials = tuple(ms.material for ms in blender_object.material_slots)
+    material_names = tuple(None if mat is None else mat.name for mat in materials)
 
     # retrieve armature
     # Because mesh data will be transforms to skeleton space,
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitives.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitives.py
@@ -55,14 +55,17 @@ def gather_primitives(
         material = None
 
         if export_settings['gltf_materials'] == "EXPORT":
-            try:
-                blender_material = bpy.data.materials[material_names[material_idx]]
-                material = gltf2_blender_gather_materials.gather_material(blender_material,
-                                                                      export_settings)
-            except IndexError:
-                # no material at that index
-                pass
-
+            blender_material = None
+            if material_names:
+                i = material_idx if material_idx < len(material_names) else -1
+                material_name = material_names[i]
+                if material_name is not None:
+                    blender_material = bpy.data.materials[material_name]
+            if blender_material is not None:
+                material = gltf2_blender_gather_materials.gather_material(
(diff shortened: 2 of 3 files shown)


julienduroure:
Seems OK to me.
Thanks !

PR #1240, opened 2020-10-18
Import: Set extras from mesh.

donmccurdy:
Fixes https://developer.blender.org/T81792.


scurest:
I forgot to put this back in in #1121. It used to be

````py
set_extras(mesh, pymesh.extras, exclude=['targetNames'])
````

The exclude prevents creating a custom property for the extras field used for shape key names.

julienduroure (approved):

PR #1249, opened 2020-10-23
Take care of active output material node

### LINKED ISSUE #1213: T78926: Materials export blank if unconnected Principled BSDF node is anywhere in a "Background" material
See https://developer.blender.org/T78926

Is same problem that in #1114

Maybe we can check if output is linked (directly or not) to Material Output node?

### LINKED ISSUE #1248: Ability to detect "is_active_output"on OUTPUT_MATERIAL node
It could be useful to have multiple Material Outputs on our materials. One node tree for Cycles, another for glTF, another one for texture baking, etc.

I've notice that the glTF exporter doesn't care about the `is_active_output` and/or `mute` property of an OUTPUT_MATERIAL node, and just take the first in the material nodes list.

A nice feature would be to take into account only node tree linked to the current active output.

Screenshot below, from the .blend in attachment, give you an example. My active output use an UVGrid texture, my inactive output use a ColorGrid texture. The exported node tree is the one which is inactive.

image

[material-output.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/5422567/material-output.zip)

julienduroure:
Fix #1248 Fix #1213
Take care of active output material node

Material node tree can have multiple output node.
Only 1 can be active.

This PR:
* Check that nodes are linked to any output material node
* and that node is the active one

Not sure how this will work with background rendering, when there is no user to click on an output node to activate it

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials_unlit.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials_unlit.py
@@ -38,7 +38,7 @@ def detect_shadeless_material(blender_material, export_settings):
     info = {}
 
     for node in blender_material.node_tree.nodes:
-        if node.type == 'OUTPUT_MATERIAL':
+        if node.type == 'OUTPUT_MATERIAL' and node.is_active_output:
             socket = node.inputs[0]
             break
     else:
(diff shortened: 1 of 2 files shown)


julienduroure:
Ok, think this is ready for review :)

Vinc3r:
Works fine on my tiny test scene in #1248 ✌

cstfan:
I have just tested on my current project, also works for me. Thanks to @julienduroure for the quick correction.

cstfan:
Very great news

PR #1250, opened 2020-10-23
Fix #1219 alpha only

### LINKED ISSUE #1219: Alpha is ignored unless BaseColor is assigned an image texture
If an image texture's Alpha output is plugged into the BSDF node's Alpha input, it gets ignored on export.
The only way it seems to work is if the BSDF node has an image texture plugged into the BaseColor input. This can be either the same image as the source of the alpha, or a different image altogether.

**To Reproduce**
1. Create a plane
2. Assign a BSDF shader to it
3. Plug an image texture node (with a PNG containing transparency) from the Alpha output to the Alpha input of the BSDF shader node
4. Export as .glb and view in any glTF viewer
5. RESULT: The alpha is not exported
6. Now go back and plug ANY image into the BaseColor input of the BSDF node.
7. Re-export as .glb and view in any glTF viewer
8. RESULT: The alpha is now showing

**Expected behavior**
Using an image texture's alpha as the source of transparency should work independently of whether the BaseColor has an image texture input. In other words, BaseColor and Alpha should be completely independent.

**Screenshots**

**This exports alpha correctly:**
image

**And so does this:**
image

**_But this does NOT export with alpha:_**
image

**.blend file/ .gltf**
[VertexAlphaIssue.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/5248968/VertexAlphaIssue.zip)

**Version**
 - OS: Windows 10
 - Blender Version 2.90.0

julienduroure:
Fix #1219

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
@@ -182,7 +182,7 @@ def __get_image_data(sockets, export_settings) -> ExportImage:
             dst_chan = Channel.G
         elif socket.name == 'Occlusion':
             dst_chan = Channel.R
-        elif socket.name == 'Alpha' and len(sockets) > 1 and sockets[1] is not None:
+        elif socket.name == 'Alpha':
             dst_chan = Channel.A
         elif socket.name == 'Clearcoat':
             dst_chan = Channel.R
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials_pbr_metallic_roughness.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials_pbr_metallic_roughness.py
@@ -80,13 +80,23 @@ def __gather_base_color_texture(blender_material, export_settings):
     if base_color_socket is None:
         base_color_socket = gltf2_blender_get.get_socket_old(blender_material, "BaseColor")
 
+    primary_socket = base_color_socket
+
     alpha_socket = gltf2_blender_get.get_socket(blender_material, "Alpha")
     if alpha_socket is not None and alpha_socket.is_linked:
-        inputs = (base_color_socket, alpha_socket, )
+        if base_color_socket.is_linked is False:
+            inputs = (alpha_socket, )
+        else:
+            inputs = (base_color_socket, alpha_socket, )
     else:
+        if base_color_socket is None:
+            return None
         inputs = (base_color_socket,)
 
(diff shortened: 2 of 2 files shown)


scurest:
This needs

```diff
diff --git a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
index 7c1fa86..bb7621e 100644
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
@@ -182,7 +182,7 @@ def __get_image_data(sockets, export_settings) -> ExportImage:
             dst_chan = Channel.G
         elif socket.name == 'Occlusion':
             dst_chan = Channel.R
-        elif socket.name == 'Alpha' and len(sockets) > 1 and sockets[1] is not None:
+        elif socket.name == 'Alpha':
             dst_chan = Channel.A
         elif socket.name == 'Clearcoat':
             dst_chan = Channel.R
```

too. Otherwise it only works if the alpha is already in the alpha channel.

julienduroure:
> This needs
> 
> ```diff
> diff --git a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
> index 7c1fa86..bb7621e 100644
> --- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
> +++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
> @@ -182,7 +182,7 @@ def __get_image_data(sockets, export_settings) -> ExportImage:
>              dst_chan = Channel.G
>          elif socket.name == 'Occlusion':
>              dst_chan = Channel.R
> -        elif socket.name == 'Alpha' and len(sockets) > 1 and sockets[1] is not None:
> +        elif socket.name == 'Alpha':
>              dst_chan = Channel.A
>          elif socket.name == 'Clearcoat':
>              dst_chan = Channel.R
> ```
> 
> too. Otherwise it only works if the alpha is already in the alpha channel.

OK thanks. I committed it

scurest:
I also think its easier to follow if you write it like

````py
    inputs = tuple(
        socket for socket in [base_color_socket, alpha_socket]
        if socket is not None and __has_image_node_from_socket(socket)
    )
    if not inputs:
        return None

    return gltf2_blender_gather_texture_info.gather_texture_info(inputs[0], inputs, export_settings)
````

This uses `__has_image_node_from_socket` instead of `is_linked`, which is what `__gather_metallic_roughness_texture` does.

julienduroure:
Done

Issue #1113, opened 2020-07-07
Export is not creating a sampler with Extend/CLAMP_TO_EDGE extension


DiegoAce:
When exporting gltf with a material that has Linear Interpolation and Repeat Extension, no sampler is exported, which I don't think is desired but I suppose is fine assuming those are the default values.
But when exporting with Linear Interpolation and Extend Extension, no sampler is exported. This seems like a bug.
Exporting with Closest Interpolation and Extend Extension does export a sampler with the expected NEAREST filters and CLAMP_TO_EDGE wraps.
Exporting with Linear Interpolation and Clip Extension does export a sampler with LINEAR filters and REPEAT wraps. Which I suppose is ok since Clip doesn't seem to translate to an allowed wrap value, but I think it should be CLAMP_TO_EDGE.

This is somewhat related to issue #307 

**Version**
 - OS: macOS
 - Blender 2.83.0 2020-06-03

scurest:
Should have been fixed by #1082. Try with a 2.90 alpha from https://builder.blender.org/download/.

DiegoAce:
Yeah 2.90 alpha is good. Thanks!

PR #1265, opened 2020-11-20
Export: roundtrip all texture wrap modes

scurest:
This teaches the exporter to recognize and roundtrip the node setup the importer uses for "unusual" wrap modes (see #926, #971).

Also add a roundtrip test that tests Mirror x Mirror and Repeat x Clamp modes

Tangentially, it also now exports Blender's CLIP extension mode (all pixels outside the edges are transparent) as ClampToEdge; this is not correct, but it seems better than Repeat. This suggestion came from #1113.


scurest:
@emackey cf https://github.com/KhronosGroup/glTF-Blender-IO/issues/926#issuecomment-587813773

scurest:
I've pushed commits to allow the KHR_texture_transform/texCoord code to skip over the wrapping nodes too. This means the exporter should now be able to roundtrip all the nodes the importer creates to hang off a texture node.

ex

scurest:
Any progress? @emackey?

emackey (approved):
This looks great to me.  Thanks for including the full test!

scurest:
@emackey So what's this waiting on then?

emackey:
Was just leaving some room for @julienduroure or anyone else to comment or review, I figured they would take a bit more time during the holidays.

But I see you tagged a related issue, so I'll push this one along.  Thanks!

PR #1274, opened 2020-11-28
Add Github Actions CI workflow.

donmccurdy:
Switches the repository from CircleCI to Github Actions. Related to https://github.com/KhronosGroup/glTF-Blender-IO/issues/1242. This is currently the only Khronos Group repository using CircleCI that I know of, and I'd prefer to use Github Actions if it works just as well. Status at https://github.com/donmccurdy/glTF-Blender-IO/actions. ~It's not working, yet.~ Tests are passing.


donmccurdy:
@emackey @julienduroure @pjoe any concerns with this? Tests are passing in <2 minutes on GitHub Actions for me, much less than previously reported on CircleCI (https://github.com/KhronosGroup/glTF-Blender-IO/pull/673) although I can't access CircleCI (because, reasons...) to check if that's still true.

pjoe:
Looks good 👍 

I've been helping other Khronos/related projects (glslang.js, spirv-tools, gpuweb) setup GH actions, so this move 'just makes sense' ™

I guess the main con is that currently GH doesn't really have a good artifact experience, i.e. you have to download the .zip locally. You can of course always upload artifacts somewhere else, if this becomes enough of a problem.

Another thing to look out for if you do a lot of builds is the quota for total artifact storage. Looks like a build takes ~3MB, so probably not an issue. i think the limit is 2GB and artifacts are retained for 90 days.

emackey:
This is a great start!

Question, historically we have "blender export" and "blender roundtrip" separated because CircleCI will parallelize them.  In fact I think there's logic in the CircleCI config to only download one copy of Blender and then spawn the parallel process after that part is done.

Looking at [these results](https://github.com/donmccurdy/glTF-Blender-IO/runs/1466597900?check_suite_focus=true) it looks like maybe Blender was downloaded twice?  If it is double-downloading, and especially if these aren't running in parallel, then we should get rid of the filter (`-g $FILTER` on the end of line 53, and related infrastructure) and just run all the tests in one shot from a single download.  What kind of performance impact would that have?

pjoe:
In GitHub actions jobs run in parallel when they are setup with `matrix` as is the case here. Because each runner is a separate compute instance it will have to download blender on it's own.

emackey:
@pjoe Thanks, good to know.  I think we could still be a good citizen here and only download the one copy, even if it takes our tests a little longer to run in serial.

donmccurdy:
> ... you have to download the .zip locally.

> Another thing to look out for if you do a lot of builds is the quota for total artifact storage.

I think these are both OK. As far as I know we're just using the build artifacts here so that PR authors can download test output for debugging, when something fails in CI.

>  I think we could still be a good citizen here and only download the one copy...

I expect we can set this up with job artifacts: job 1 downloads Blender and maybe even keeps it cached for ~24 hours, job 2 runs our two test suites in parallel. I'll look into that, but wouldn't hold the PR on it since CircleCI already appears to be downloading Blender twice now.

donmccurdy:
Updated — the workflow now [caches the Blender build](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows), keyed by the given major/minor version. As long as CI runs ~1x/week, the build could be cached indefinitely until we bump the version. Is that what we want? Or should we be clearing the cache each time the SHA1 in the download URL changes?

emackey:
> Or should we be clearing the cache each time the SHA1 in the download URL changes?

Unfortunately yes, these are nightly builds, and the API changes from time to time.  Our master branch corresponds to the leading-edge alpha development branch on the Blender Foundation side, so those changes apply immediately.

pjoe:
Cacing blender makes sense 👍 Maybe could just key the cache on date?

Just FYI, compared to other things I've seen with GitHub actions the ~20 secs it takes to download is nothing 😀 So I wouldn't expect GitHub to have much issue with this and IIRC when I checked with the blender guys, they confirmed that builder.blender.org does have CDN.

donmccurdy:
Easy enough to cache by the build's URL, so let's just do that. I think this is ready to go. 👍

PR #1282, opened 2020-12-01
Fix #1281 Avoid error when exporting from background

### LINKED ISSUE #1281: T83188: Can not call bpy.ops.export_scene.gltf when running Blender in the background (no UI)
See https://developer.blender.org/T83188

julienduroure:
Fix #1281


donmccurdy (approved):

mohsen123:
I still get an error when I add a background to my materials :/

donmccurdy:
@mohsen123 this issue refers to running Blender itself "in the background" i.e. without a UI. It sounds like you have another issue, could you open a bug and include a `.blend` so that we can debug this?

PR #1297, opened 2020-12-17
Import: create placeholder images for files that can't be loaded (fix #1296)

### LINKED ISSUE #1296: Improve handling of missed textures?
When textures are stored externally, and when some of them are not available (i.e. not present in the dir with .gltf), the importer will remove the reference to this texture completely.
Can we change this, so when texture is not available, the texture slot will be still there? (and the texture will be displayed as pink by default, then)

I'm currently on 2.90.1 with unmodified gltf addon.

scurest:
Fixes #1296.

CesiumMan with its image missing in master:

master

And in this branch; a placeholder image is created pointing at where the image should be:

this branch

The missing texture will be reported through `File > External Data > Report Missing Files`. You can check what the missing file should be and put the image file there after the import to make it work.

Placeholders are created only for images from a file (ie. not from bufferView/data URIs).

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_image.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_image.py
@@ -43,13 +43,11 @@ class BlenderImage():
         try:
             if img.uri is not None and not img.uri.startswith('data:'):
                 # Image stored in a file
-                img_from_file = True
                 path = join(dirname(gltf.filename), _uri_to_path(img.uri))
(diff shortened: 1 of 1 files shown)


emackey:
Seems like a good idea.

What does the exporter do when it finds this?

scurest:
Both the material setups shown in the OP export the same way: with no base color texture.

mat

julienduroure:
Seems OK.
Thanks !

PR #1300, opened 2020-12-22
🐛 correctly resolve uvset when nodepath is rerouted

m0ppers:
I have a reroutenode in place on the road to my uvset...unfortunately blender will just return 0 (which is wrong because I have a light UV map at position 0) and I get super distorted models out of blender....

this bugfix handles reroute nodes in between...


scurest:
Can you check if your file is fixed by #1265? It uses `previous_node` to get the previous node from a socket which should skip over reroute nodes.

scurest:
Have tested this

reroute

Works in master after merge of #1265.

julienduroure:
(Note that #1265 is not merged yet with Blender repo. I will do it tomorrow)
Seems that we can close this PR, as a recent PR alread fixed it

m0ppers:
thanks...sorry I didn't find time to test it :( good that it works now (my PR really was just a quick hack)

PR #1309, opened 2021-01-07
Import: bugfix: use Separate R node for clearcoat textures

scurest:
This is a bugfix for #1061.  
cc @donmccurdy 

The comment here says the Separate RGB node when wiring up the Clearcoat texture can be skipped because the R component is used when a Color node is connected to a Value node. The reason I believed this is because the exporter acts as if this were true, but what _Blender_ actually does is

> [the color data is converted to its gray scale equivalent](https://docs.blender.org/manual/en/2.91/interface/controls/nodes/parts.html?highlight=gray+scale#conversion)

So this just adds the missing Separate RGB node. Because, like I said, the exporter uses R in this case, there is no change in roundtripping behaviour.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_KHR_materials_clearcoat.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_KHR_materials_clearcoat.py
@@ -16,7 +16,7 @@ from ...io.com.gltf2_io import TextureInfo, MaterialNormalTextureInfoClass
 from .gltf2_blender_texture import texture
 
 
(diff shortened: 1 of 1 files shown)


donmccurdy (approved):

scurest:
@donmccurdy Can you merge?

donmccurdy:
Done, thanks!

julienduroure:
If we wanted it on 2.92, it should have been merged on blender-2.92-release branch, not on master

scurest:
@julienduroure Just cherrypick it onto that branch.

PR #1312, opened 2021-01-09
Export: more fixes for changing the filename when format changed (#718)

scurest:
Further fixes to the `on_export_format_changed` junk.

* When exporting using the Python console in master, it will fail if passing `export_format=` because `context.space_data.active_operator` is None, and `on_export_format_changed` does not handle that. Further defensive checks have been added to prevent this. (I've also checked that #1281 did not regress.)
* The hack with `bpy.ops.file.filenum` was replaced with setting `context.space_data.params.filename`, which is much cleaner. This is thanks to batFINGER's [answer](https://blender.stackexchange.com/a/207665/88681) to my question on Blender SE.

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -70,21 +70,42 @@ from bpy_extras.io_utils import ImportHelper, ExportHelper
 extension_panel_unregister_functors = []
 
 
+def ensure_filepath_matches_export_format(filepath, export_format):
+    import os
+    filename = os.path.basename(filepath)
+    if not filename:
+        return filepath
+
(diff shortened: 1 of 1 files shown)


scurest:
@donmccurdy @julienduroure @emackey

donmccurdy (approved):
Looks good to me, thank you! I'll merge this in ~2 days if there are no objections (and if no one merges it before I do).

emackey (approved):
Thanks so much @scurest for your continued patience with us.

scurest:
Since this is a bugfix, will it be backported to the 2.92 branch?

julienduroure:
I will backport and merge with blender repo tonight

julienduroure:
Done!

PR #1339, opened 2021-02-25
Fix #1317

### LINKED ISSUE #1317: Export: "Basis" shape key vertex positions sometimes exported wrong
This was found in https://github.com/godotengine/godot-blender-exporter/issues/391. Using the example.blend from there...

The vertex positions in the Basis shapekey for Cube.002 are wrong. You can check by reimporting or in the Three viewer.

badcube

Fetching the vertex positions from the Basis keyblock instead of `mesh.vertices` appears to fix this (would have to change normals/tangents too of course)

````diff
diff --git a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
index 5a78ab04..4eb64ac3 100644
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
@@ -290,7 +290,8 @@ def extract_primitives(glTF, blender_mesh, library, blender_object, blender_vert
 
 def __get_positions(blender_mesh, key_blocks, armature, blender_object, export_settings):
     locs = np.empty(len(blender_mesh.vertices) * 3, dtype=np.float32)
-    blender_mesh.vertices.foreach_get('co', locs)
+    source = key_blocks[0].relative_key.data if key_blocks else blender_mesh.vertices
+    source.foreach_get('co', locs)
     locs = locs.reshape(len(blender_mesh.vertices), 3)
 
     morph_locs = []
````

### PR DESCRIPTION
**Pull request by julienduroure**
Fix #1317

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
@@ -374,7 +374,8 @@ def extract_primitives(glTF, blender_mesh, library, blender_object, blender_vert
 
 def __get_positions(blender_mesh, key_blocks, armature, blender_object, export_settings):
     locs = np.empty(len(blender_mesh.vertices) * 3, dtype=np.float32)
-    blender_mesh.vertices.foreach_get('co', locs)
+    source = key_blocks[0].relative_key.data if key_blocks else blender_mesh.vertices
+    source.foreach_get('co', locs)
     locs = locs.reshape(len(blender_mesh.vertices), 3)
 
     morph_locs = []
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
@@ -412,6 +412,7 @@ def __get_normals(blender_mesh, key_blocks, armature, blender_object, export_set
     normals = np.empty(len(blender_mesh.loops) * 3, dtype=np.float32)
     if key_blocks:
         normals = key_blocks[0].relative_key.normals_split_get()
+        normals = np.array(normals)
     else:
         blender_mesh.calc_normals_split()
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by scurest** [comment]
If you're fetching the positions from the keyblock, shouldn't you be fetching the normals from the keyblock too? (Same goes for tangents but shapekey tangents are already busted so I guess it doesn't matter.)

**Comment by julienduroure** [comment]
Just added normals from key_block

**Comment by julienduroure** [comment]

File: addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py [hunk 1]
```diff
@@ -408,10 +409,14 @@ def __get_positions(blender_mesh, key_blocks, armature, blender_object, export_s
 
 def __get_normals(blender_mesh, key_blocks, armature, blender_object, export_settings):
     """Get normal for each loop."""
-    blender_mesh.calc_normals_split()
-
     normals = np.empty(len(blender_mesh.loops) * 3, dtype=np.float32)
```
scurest, line 412: You can push this down into the else block. It doesn't need to be allocated in the if case.

addons/io_scene_gltf2/blender/exp/gltf2_blender_extract.py
```diff
@@ -408,10 +409,14 @@ def __get_positions(blender_mesh, key_blocks, armature, blender_object, export_s
...
-    blender_mesh.loops.foreach_get('normal', normals)
+    if key_blocks:
+        normals = key_blocks[0].relative_key.normals_split_get()
+        normals = np.array(normals)
```
scurest, line 415: This makes a float64 array by default. Pass `dtype=np.float32` to get a float32 array, like the other case.

Issue #1244, opened 2020-10-21
2-pass hierarchy export


julienduroure:
Let's create a ticket to summarize work on 2-pass hierarchy export.
Main purpose is to have a first pass on all object on blender side, to be able to export correctly.

I want to work on it on next weeks.

Main issues that it will solve:

- Better/Fixes on collection export
  * Collection
  * Linked Collection
  * Recursive linked collection
  * Armature proxy
  * Armature linked proxy
  * Overwriting ???
- Better/Fixes "use_selected" export
  * Standard object should already work with merged workaround
  * Collection
  * Linked Collection
  * Recursive linked collection
  * Armature proxy
  * Armature linked proxy

julienduroure:
Better/Fixes on collection export part of this ticket is currently done in #1378 (still WIP for now)

julienduroure:
Fixed by 69746b2b31eeee080d5dfd3c74f1e02742795ed2

PR #1361, opened 2021-03-29
New export options: Visible Objects, Renderable Objects, and Active Collection

### LINKED ISSUE #1139: None of the Blender visibility options affect export of objects
**Describe the bug**
Typical production scenes contain objects that have visibility toggled, either for scene view or rendering or both.
However, none of those settings affect object export; I understand that GLTF does not specify a way to state "visibility" per object, but expectation would clearly be that e.g. "hide in rendering" would prevent an object from getting exported instead of needing to either hunt down and delete all hidden objects, or make selection sets for export that will get outdated once new objects are added.

**To Reproduce**
1. Create a Cube
2. Create a monkey
3. Change all 3 visibility options of the Cube to "hidden"
4. Export as GLB
200723-145159-blender

**Expected behavior**
Cube does not exist in export

**Actual behaviour**
Cube exists in export
image

**.blend file/ .gltf**
[GLB-Visibility-Issues.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/4965958/GLB-Visibility-Issues.zip)

**Version**
 - OS: [e.g. macOS, linux]
 - Blender Version [e.g. 2.82, 2.83, or build date]

**Additional context**
I think there was some discussion around this in #602, however the discussion there was around explicitly exporting hidden objects. Given there's at least 3 different ways of specifying visibility in Blender I'd hope that one of them at least could be used for determining if an object is exported, if necessary with an export option.

Pauan:
Fixes #1139

The options (and the code) are self-explanatory. The behavior is exactly as you would expect:

* Visible Objects: If an object is hidden (either with the Eye or the Hide in Viewport) then it won't be exported.
* Renderable Objects: If an object is not renderable then it won't be exported.
* Active Collection: If an object is not a child of the active collection then it won't be exported.

All options default to false, so this does not change existing behavior, it simply gives new optional features.


julienduroure:
I think we should align options to other exporter, like FBX for example :
* Limit to selected
* Limit to active collection

(FBX also has limit export by object type.)

Pauan:
@julienduroure These features are extremely useful and are needed by many people, so I would advocate for them even if the other exporters didn't support it.

The linked topic goes into detail, but the situation is that every exporter does something different:

* FBX has active collection and selection.
* Collada has selection.
* Alembic has selection, renderable, and visible.
* USD has selection and visible.
* PLY has selection.
* STL has selection.
* OBJ supports selection.
* SVG supports selection.

So if you want to align to other exporters, then supporting renderable and visible *does* align with Alembic and USD.

julienduroure:
+1 about Alembic, that is the most recently developed in Blender.
So OK to include all these options

Note that currently export by selection is quite buggy. This should be fixed with what is called two pass hierarchy export #1244 
With current lockdown / context since 1 year, I don't have many free time to work on glTF-Blender-IO, but I hope to be back (soon ) on a more frequent way

Pauan:
@julienduroure I think that sort of big refactoring should be done in a separate PR. This PR is small, simple, and self-contained.

julienduroure:
@Pauan Yes sure, this note was not here to ask working on two pass export here, but only to remember that current filter is currenlty buggy, and this PR will add some functionnalities based on this legacy system

I will perform some tests soon. My sugestion is to check that this PR is at same level of features (features and known issues) that other filters

gigablox:
Needed the "Visible Objects" feature for a project, I was able to cherry pick this code into the GLTF plugin and it worked - extremely helpful. Many thanks @Pauan for putting this together. Will be happy to see this make it into the main repo.

wilcoschoneveld:
What is the status of this PR?

julienduroure:
I will merge this PR tonight

wilcoschoneveld:
Just tested this with blender 2.93 and it works! Thanks a lot! 👍🏼

bencresty:
Thanks for adding this setting! A real time and frustration saver!

PR #1366, opened 2021-04-07
Fix #1365 - Basis name is reserved

### LINKED ISSUE #1365: Import: morph target named 'Basis' imported wrong
Since we use 'Basis' as the name of the special first shape key, we shouldn't let morph targets use it. You can test by taking, say, MorphStressTest, and renaming one of targetNames in the glTF to 'Basis'.

Fix is just to use `set(['Basis'])` [here](https://github.com/KhronosGroup/glTF-Blender-IO/blob/bb488b245f0c83fdd566b43c2c762d469c044326/addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py#L168), so 'Basis' will always be regarded as "taken".

julienduroure:
Fix #1365 - Basis name is reserved


julienduroure:

PR #1367, opened 2021-04-07
Fix #1364 SK range

### LINKED ISSUE #1364: Import: Negative shape key values clamped to 0.0.
**Describe the bug**
GLTF files with shape keys that are set to negative values are not respected.

When the files are imported, the values below 0.0 are clamped.

Likely as a result of the min/max values present in Blender's UI automatically performing that clamping.

**To Reproduce**
Here is a sample GLTF file to import which has negative shape keys for the size of the dividing screen on the desk:

[Strata 360 Model 2 - Screen.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/6262317/Strata.360.Model.2.-.Screen.zip)

**Expected behavior**

The sample model should look like this:
Capture
That screenshot is from the 3D viewer included with Windows 10.

In Blender it looks like this:
screen height

The file when imported should show a dividing screen on the desk that's very short, but instead it is very tall in Blender.

The screen should be short because it's size is being controlled by a shape key value, which should have a value of **-0.7.**

 But instead of this value, there is a value of 0.0, with the min/max values for the shape key likely preventing the -0.7 value:

screen height ui2

Ideally the importer should automatically set min/max values that allow for at values as negative or positive as required to import the model correctly. In this case, setting the minimum value to -0.7.

**Version**
 - OS: Windows 10
 - Blender Version 2.92

julienduroure:
Fix #1364 SK range

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation_weight.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation_weight.py
@@ -83,3 +83,10 @@ class BlenderWeightAnim():
                     group_name="ShapeKeys",
                     interpolation=animation.samplers[channel.sampler].interpolation,
                 )
+
+                # Expand weight range if needed
+                kb = obj.data.shape_keys.key_blocks[kb_name]
+                min_weight = min(coords[1:2])
+                max_weight = max(coords[1:2])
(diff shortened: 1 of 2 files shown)


julienduroure:

PR #1372, opened 2021-04-12
Fix #1371 importing extra

### LINKED ISSUE #1371: Import: exception importing large integers in extras
If an extra property is an integer which doesn't fit in a C int, (eg. >= 2^31), the whole import fails with an exception. 

Example:

```json
{
  "asset": { "version": "2.0" },
  "nodes": [
    {
      "extras": {
        "number": 9999999999
      }
    }
  ]
}
```

The guard for importing extras

https://github.com/KhronosGroup/glTF-Blender-IO/blob/8f41a73e442c3cf10ac4f605e1fc962927f421ee/addons/io_scene_gltf2/blender/com/gltf2_blender_extras.py#L90-L93

catches TypeErrors, but in this case an OverflowError is thrown. I'll repeat my conclusion from #1353: **the catch should be broadened from TypeError to Exception** to prevent future problems like this from making the whole import fail.

julienduroure:
Fix #1371 
Trying to convert extra to string in case of error, and if really it doesn't want, except

### DIFF
--- a/addons/io_scene_gltf2/blender/com/gltf2_blender_extras.py
+++ b/addons/io_scene_gltf2/blender/com/gltf2_blender_extras.py
@@ -89,5 +89,9 @@ def set_extras(blender_element, extras, exclude=[]):
 
         try:
             blender_element[custom_property] = value
-        except TypeError:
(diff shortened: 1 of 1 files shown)


julienduroure:

Issue #566, opened 2019-06-25
Multiple duplicated skins are exported for each mesh child of the armature


dpalacio:
**Description**
When an armature in Blender has multiple children meshes, each mesh gets its own skin copy in the glTF export. The result is multiple skeletons, one for each child mesh, in the exported file. 

**To Reproduce**
1. Open the attached .blend file and check the hierarchy
2. Export scene to .gltf text
3. Verify that multiple skeletons have been created

**Expected behavior**
A single skin/skeleton should be made for each Blender armature

**Test case**
I have attached a blend file to help reproduce the behavior. I'm no 3D modeler so please excuse the content.
[blender_export_skin_test.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3327531/blender_export_skin_test.zip)

**Version**
 - OS: Linux
 - Blender Version: blender-2.80-c0c1b4542f39-linux-glibc224-x86_64

**Additional context**
This behavior has been referenced also in issue #331, which I have also have seen happen.

marstaik:
I recently ran into this behavior on my own, without the test case, while trying to import to Godot.

This is a very annoying problem to deal with, every mesh has a skeleton copy.

fire:
Will review.

julienduroure:
@donmccurdy @emackey I have a question about inverseBindMatrices: If multiple mesh objects are linked to an unique armature, I don't know how to create only 1 skin object in glTF, because inverseBindMatrices are differents ... Any thoughts / test cases / info about that ?

emackey:
Sorry, I don't know about skins/armature.  @donmccurdy?

donmccurdy:
The child meshes are not just rigidly transformed by the parent bones, I assume, but their vertices are weighted and transformed? In the rigid case, like parts of the ["Robot" example](https://threejs.org/examples/?q=animatio#webgl_animation_skinning_morph), the child meshes do not need skins at all.

Assuming the more complex case where each child mesh's vertices are transformed, I believe one option would be to create a new skin with unique inverseBindMatrices for each mesh, while reusing the same joints (or a subset of the joints) for each skin. Then the skeleton hierarchy is not duplicated, and each mesh has its own matrix.

This test case covers the situation above: https://bghgary.github.io/glTF-Assets-Viewer/?manifest=https://raw.githubusercontent.com/KhronosGroup/glTF-Asset-Generator/v0.6.0/Output/Manifest.json&folder=2&model=8

[...]

PR #1378, opened 2021-04-18
Export vtree

julienduroure:


julienduroure:
Still a WIP.
This is only a first step: Have a virtual Tree for nodes, and then use this tree to construct glTF tree.

Still have some TODO to manage, because managing skin & joints with this current system is not easy

julienduroure:
Seems tests are KO only for pull_request because a cache issue (still caching 2.93 where push caches 3.0)
Anyone know how to solve it?

julienduroure:
2 different new things in this branch:
- manage a tree before export. It uses an uuid for each "real" object, and these uuid are passed to other function (and can be used for caching)
- Using custom caching, as proposed by scurest. Already used in some cases here.

In current version:
- All tested collection instance issues are solved
- No more proxy check is needed (for example Text and Mesh using same name in different linked data are now ok)

In attachment a test case for multi level collection instance, for materials, and meshes

[collections.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/6735246/collections.zip)
[materials.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/6735248/materials.zip)
[mesh.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/6735249/mesh.zip)


Of course lots of work are still needed, but this is a first step.

dysbulic:
¿Will the resolution of this issue fix the fact that the current glTF export flattens all of Blender's collections into a single level?

For example, this .blend scene graph:

Screenshot from 2021-08-09 15-10-08

Becomes this single-tiered structure after a round-trip through the glb export/import process:

Screenshot from 2021-08-09 15-12-09

If this issue fixes that problem, I would like to put a little ETH on it through [Gitcoin](//gitcoin.co/funding/new). The scene graph is so handy for describing nested relationships, and I really wish we could get access to it when rendering glTF files.

dysbulic:
About two weeks ago I posted a question as to whether this code would allow exporting a hierarchical scene collection.

I just wanted to poke again and see if I could get an answer. I looked at the code changes, and I see there's a tree structure being saved. That's promising…

I created a [sample glTF file](https://github.com/dysbulic/stay/blob/react/public/cubes.gltf) that contains nested children. It produces a scene hierarchy like:

Screenshot from 2021-08-24 12-45-30

So, all the nesting is of objects rather than collections. In the [glTF spec](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#:~:text=%20any%20gltf%20object%20can%20have%20an%20optional%20extensions%20property), it says `extensions` are allowed on any glTF object, is it possible to set an extension property that causes the group to open as a collection?

julienduroure:
I open a new issue #1456 to discuss about that.
This PR is a technical refactoring. I don't want to mix with feature requests

julienduroure:
Ok, seems it's now time to deeply test this PR.
Don't hesitate to test, and report any regression.
I will prepare a document next week to explain what are the changes, why we can have some performance decrease in some cases (but for more accuracy in export). I will share this on most know Blender user forum/site, to have a wide test range.

julienduroure:
Here are more information about how people can test it: https://devtalk.blender.org/t/test-future-gltf-exporter-version/22657

emackey:
@julienduroure I tried exporting a model with multiple same-name NLA tracks, using a [vtree build](https://builder.blender.org/download/experimental/gltf_vtree/) from today.  It looks like all my parts came out animated separately.  I have "Group by NLA track" checked, but it didn't perform any grouping, they all came out as loose pieces.

julienduroure:
@emackey Do you have your .blend test case? 
I can't reproduce here for now on simple test case

image

image

emackey:
Took a while to figure out how to reproduce, but it turns out the bug activates when "Include Selected Objects" is checked in the exporter options.  The combined NLA track exports alongside loose animations per-object.

[NLA_Test_vtree.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/8050223/NLA_Test_vtree.zip)

In the above example, there should be only one NLA animation exported, called "CubeDance."  But when "Include Selected Objects" is used, the combined animation appears alongside individual cube animations.

julienduroure:
@emackey Nice catch!
I just pushed a fix, and ask builder to compile a new build. Should be available in a few hours

julienduroure:
@s-light You need at least 3.1 to run this branch, you can't use 3.0

emackey:
@s-light This branch has its own build, too:  https://builder.blender.org/download/experimental/gltf_vtree/

s-light:
sorry... is not related to vtree.
it also happens in the current master - 
and after some digging i found that the used `use_frame_range` is a blender3.1 feature....

Capital-Asterisk:
Hi, I just tried out these changes to fix my problems in #1576. I've noticed that skins are duplicated if multiple objects are parented to a single armature:
```json
"skins" : [
    {
        "inverseBindMatrices" : 6,
        "joints" : [
            1,
            0
        ],
        "name" : "Armature"
    },
    {
        "inverseBindMatrices" : 6,
        "joints" : [
            1,
            0
        ],
        "name" : "Armature"
    }
],
```
I'm not sure if this is intentional, but this did not happen before. It may re-introduce #566, though this issue and its resolution seems complicated. Unlike the mentioned issue, the inverseBindMatrices are exactly the same in this case.

For some context, I'm writing a C++ application that uses gltf, and relies on nodes within the same armature to have the same skin; it would abort otherwise.

julienduroure:
> I've noticed that skins are duplicated if multiple objects are parented to a single armature

Confirmed. I will have a look soon

julienduroure:
@Capital-Asterisk I just Fixed the duplicate skin data bug

PR #1397, opened 2021-05-27
Import: fix issue involving flipped bone Z dir (#1396)

### LINKED ISSUE #1396: Import: bone rotates wrong way with Temperance (bug)
This was reported at https://devtalk.blender.org/t/gltf-export-issue-upd-import-issue/18970.

In this [bed model](https://github.com/KhronosGroup/glTF-Blender-IO/files/6541389/bed.zip), the leg part of the bed rotates down in Blender. In Three and Babylon, it rotates up instead.

out

edit: only happens with "Temperance"
edit: think "Temperance" might just be a coincidence, it seems to be triggered by very specific numeric values

scurest:
Fixes #1396.

Blender issue is confirmed, but don't think we need to around for a fix. This just avoids the issue by setting the bone Z axis after the length. This fixes the file from https://devtalk.blender.org/t/gltf-export-issue-upd-import-issue/18970.

This is also a good candidate for 2.93, either before the release next week, or after since its an LTS.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_node.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_node.py
@@ -145,8 +145,8 @@ class BlenderNode():
             arma_mat = vnode.editbone_arma_mat
             editbone.head = arma_mat @ Vector((0, 0, 0))
             editbone.tail = arma_mat @ Vector((0, 1, 0))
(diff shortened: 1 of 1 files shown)


julienduroure (approved):

julienduroure:
I didn't get the time to talk to blender dev, so I think it's too late for 2.93. I am going to merge in master, and get it in 2.93.1 (LTS next release)

PR #1399, opened 2021-05-27
CI: be less wasteful

scurest:
A push to the (eg) vtree export branch downloads Blender [four times in CI](https://github.com/KhronosGroup/glTF-Blender-IO/runs/2687067559) (two event triggers X two jobs). This fixes it so it only downloads once.

* Run only on pushes to _master_ (still runs on PRs to all branches). Currently two workflows are started if a maintainer pushes to a PRed branch in this repo, one for the push and one for the PR. (This only affects maintainers since the rest of us push to branches in forks.) See https://github.community/t/duplicate-checks-on-push-and-pull-request-simultaneous-event/18012.
* Do all tests in one job, instead of two, as suggested by @emackey in https://github.com/KhronosGroup/glTF-Blender-IO/pull/1274#issuecomment-735243033. This adds about 60s wall clock time.

This also removes setup-node in favor of using the version of node/yarn that comes preinstalled on the CI machine, which is good enough for us.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,15 +17,6 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 20
 
-    strategy:
-      matrix:
-       include:
-         - filter: "blender_export"
-         - filter: "blender_roundtrip"
-
-    env:
-      FILTER: ${{ matrix.filter }}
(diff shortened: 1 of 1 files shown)


emackey (approved):
Thanks @scurest.

Issue #387, opened 2019-03-27
Animation seems to always use frame_start = 0


Vinc3r:
[A BabylonJS user was confronted to an issue](https://forum.babylonjs.com/t/blender-to-gltf-walk-cycle-animation-shortly-pauses-in-the-beginning/2179/6?u=vinc3r), not specific of BabylonJS, and found that the issue came from difference between Blender default starting frame (1) and what seems to be the default gltf export starting frame (0).

So I've made a few checks and it seems custom starting frame is indeed not take into account.
Here a simple Blender animation starting from 5 to 25, the animation loop without break:
![img](https://i.imgur.com/rT4E0ZC.gif)
Here the glb export (into BabylonJS Sandbox) where we can see a break into the animation, which seems to be ghost frames from 0 to 5:
![img](https://i.imgur.com/JrZ7Oue.gif)

It's maybe a gltf standard limitation for animations to start at frame = 0, if so I think it deserves to be mentionned in the Blender Manual.

mauricedoepke:
There is actually an option to deal with that in the export settings. The tab animations contains the option "Limit to playback range", which is true by default. Searching through the code, it seems like the functionality for this option is not implemented yet.

Vinc3r:
ok! (I've forgot to mention I've play with some exporter options too), thanks for the clarify.
So wait&see :)

emackey:
> The tab animations contains the option "Limit to playback range", which is true by default. Searching through the code, it seems like the functionality for this option is not implemented yet.

If we have an un-implemented option still in the mix at this point, that's a bug.  We should yank the option out, at least until someone has time to write the supporting code.

MiikaH:
I also ran into this issue when trying to export low FPS looping scene. There was a noticeable freeze in the exported animation because of this issue.

Here is a very minimal .blend file to better see and test the issue:
[2_point_loop_anim_blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/3085372/2_point_loop_anim_blend.zip)

It is a 3 frame scene played back at 1 fps that loops a cube back and forth with linear interpolation.
![anim_range](https://user-images.githubusercontent.com/1881360/56219235-11561000-606f-11e9-9131-5ba94eebe208.png)

[...]

PR #1408, opened 2021-06-05
Applies FRAME_RANGE allowing animations to start at non-0 frames.

abaire:
Fix for #387

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
@@ -203,6 +203,10 @@ def gather_keyframes(blender_object_if_armature: typing.Optional[bpy.types.Objec
 
         start_frame = min([channel.range()[0] for channel in channels  if channel is not None])
         end_frame = max([channel.range()[1] for channel in channels  if channel is not None])
+
+        if export_settings[gltf2_blender_export_keys.FRAME_RANGE]:
+            start_frame = max(bake_range_start, start_frame)
+            end_frame = min(bake_range_end, end_frame)
     else:
         start_frame = bake_range_start
         end_frame = bake_range_end
@@ -225,7 +229,9 @@ def gather_keyframes(blender_object_if_armature: typing.Optional[bpy.types.Objec
         frame = start_frame
         step = export_settings['gltf_frame_step']
         while frame <= end_frame:
-            key = Keyframe(channels, frame, bake_channel)
+            # gltf2 expects time to start at 0 whereas Blender defaults to frame 1.
+            adjusted_frame = frame - start_frame
+            key = Keyframe(channels, adjusted_frame, bake_channel)
             if isinstance(pose_bone_if_armature, bpy.types.PoseBone):
 
                 mat = get_bone_matrix(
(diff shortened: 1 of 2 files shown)


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
```diff
@@ -203,6 +203,10 @@ def gather_keyframes(blender_object_if_armature: typing.Optional[bpy.types.Objec
 
         start_frame = min([channel.range()[0] for channel in channels  if channel is not None])
         end_frame = max([channel.range()[1] for channel in channels  if channel is not None])
+
+        if export_settings[gltf2_blender_export_keys.FRAME_RANGE]:
+            start_frame = max(bake_range_start, start_frame)
+            end_frame = min(bake_range_end, end_frame)
```
julienduroure, line 209: Why using frame range only when fcurve is not sampled?

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_samplers.py
```diff
@@ -42,6 +42,17 @@ def gather_animation_sampler(channels: typing.Tuple[bpy.types.FCurve],
...
+      # Attempt to adjust the bake range to match the nla track strip that matches the action_name.
+      for track in blender_object.animation_data.nla_tracks:
+        if not track.strips:
+          continue
+        for strip in track.strips:
+          if strip.name == action_name:
+            bake_range_start = strip.frame_start
+            bake_range_end = strip.frame_end
+            break
+
```
julienduroure, line 55: This will break merging tracks into a single gltf animation:
image
[restrict_range.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/7034162/restrict_range.blend.zip)

julienduroure (changes requested):
Hello,
Thanks for your PR :)
Here are some feedback / things that need to be adjusted

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_samplers.py
julienduroure: @abaire Do you plan to work on this again?

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
abaire: No reason apart from general unfamiliarity with the code, which made me err on the side of minimizing the change.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_samplers.py
abaire: I'd be more than happy if you'd like to take it up. I really don't feel like I know enough about the existing code or event the animation parts of gtlf2 to feel confident that I can make this change safely. I mostly threw it in as a demo/proof of concept (that happens to fix my particular workflow) because there was some interest outside of my team.

abaire (commented):
Sorry, I totally missed your comments from a month ago!

julienduroure:
We can now close this ticket, as the feature is now implemented in main branch, with refactoring of animation export code

PR #1409, opened 2021-06-06
Keep original texture files

julienduroure:
Add a new option, when exporting as 'gltf separate', to keep original texture images if possible.

Proof of concept.
Works only for simple cases yet

Using this example :
[test_keep_texture.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/6604867/test_keep_texture.zip)

Do not work for occlusion/roughnessMetalic
Works for base color, emissive, normal

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
@@ -198,7 +198,6 @@ def __get_image_data(sockets, export_settings) -> ExportImage:
         # Assume that user know what he does, and that channels/images are already combined correctly for pbr
         if export_settings['gltf_keep_original_textures']:
             composed_image = ExportImage.from_original(result.shader_node.image)
-            print("-->", result.shader_node.image.name)
 
         else:
             # rudimentarily try follow the node tree to find the correct image data.
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
@@ -46,7 +46,7 @@ def gather_image(
         uri = __gather_uri(image_data, mime_type, name, export_settings)
     else:
         # Retrieve URI relative to exported glTF files
-            uri = __gather_original_uri(image_data.original.filepath, export_settings)
+        uri = __gather_original_uri(image_data.original.filepath, export_settings)
 
     buffer_view = __gather_buffer_view(image_data, mime_type, name, export_settings)
 
@@ -65,7 +65,7 @@ def gather_image(
     return image
 
 def __gather_original_uri(original_uri, export_settings):
-    
+
     def _path_to_uri(path):
         import urllib
         path = os.path.normpath(path)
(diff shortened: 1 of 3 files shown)


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
```diff
@@ -42,7 +42,11 @@ def gather_image(
     mime_type = __gather_mime_type(blender_shader_sockets, image_data, export_settings)
     name = __gather_name(image_data, export_settings)
 
-    uri = __gather_uri(image_data, mime_type, name, export_settings)
+    if image_data.original is None:
+        uri = __gather_uri(image_data, mime_type, name, export_settings)
+    else:
+        uri = image_data.original.filepath #TODO make it relative to gltf file (currently is filepath of image in blender)
```
scurest, line 48: There's [code](https://github.com/KhronosGroup/glTF-Blender-IO/blob/423685f18f55d352e10995ac989fc62b422c4d11/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py#L250-L267) in #453 for getting the relative path to an image you can use.

reduz:
@julienduroure Thanks hugely! Will make sure to test tomorrow. For ORM I would guess that the first texture you find connected to either of them is probably enough.

reduz:
Ok gave it a test. Enabled the option, but I am not sure if it works because its still re-saving the textures together with the model, as can be seen from the date of the files:
image

netpro2k:
I wonder if this even needs to be an option. It seems that unless we need to generate a new image (either because the image needs to be composited from multiple channels, or needs to be exported in a format that differs from the input format) we should always prefer to skip generating a new image. What are examples where you might want to uncheck this option?

reduz:
@netpro2k If I have to be completely honest, I expected the mode I suggested would be how the exporter worked to begin with. I was quite surprised that it bakes textures for export.

I think the rationale behind their decision is that this functionality was designed more thinking about making simple files within blender (as an example, painting the individual textures for ORM with Blender paint tools) and have them composited on export to a bundled file that can be uploaded somewhere. This is great for artists making simple models, but for a more normal game development workflow its just quite frustrating. 

From my usage perspective, I would think this should be the default behavior and then a "bake textures" option be present if you want new textures to be created by merging texture channels. That said, a this point it does not matter much to me how it is implemented, I am more concerned into actually it being able to be used for game development.

julienduroure:
Hello,
I pushed some changes:
- uri are now relative to exported file
- all textures are now managed

Warning:
If if you use more than one texture,where pbr standard requires only one for example ORM, or alpha separate from baseColor), only one texture will be used for export. This can lead to unexpected results, because these cases require baking.

Please test it / review it

About default "keep original"
I don't have strong conviction about it, my I think that we should keep default option as is, because using "keep originals" can lead to unexpected exports, where not keeping original

reduz:
@julienduroure Thanks hugely for doing the work yourself! Will test today and give feedback.

reduz:
Been testing it on huge scenes with lots of materials and textures, works perfectly :+1:

julienduroure:
So, anyone want to review before merging?

julienduroure:
Last chance of reviewing before I merge it!

HumanRobot3D:
Please extend the Keep Original option to .glb as explained here https://github.com/KhronosGroup/glTF-Blender-IO/issues/1787

Issue #1308, opened 2021-01-07
Export textures as WebP


makidrone:
**Is your feature request related to a problem? Please describe.**
We're working on virtual world software where we have many 3D models loading all around from different http servers. Features such as draco and brotli compression help improve file sizes significantly, but for textures we've been using webp. Using our own Blender addon, it tries to convert textures to webp and modify gltf files afterward. However it's not intuitive and doesn't work with glb files. Most people avoid it because it doesn't import back in either.

**Describe the solution you'd like**
When exporting gltf in Blender, under textures you can select webp. A new set of settings appears where you can toggle lossless mode, or otherwise select a quality between 0 to 100% (possible show this for jpeg as well). This should work with glb files too.

There's an extension for webp here: https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Vendor/EXT_texture_webp/README.md
Though we haven't been using it since webp just works.

I'm not sure how we could compress/decompress with webp in Blender, however for my addon I've packaged the `cwebp` executable into the addon for Linux, Windows and macOS. 

**Describe alternatives you've considered**
Manually using `cwebp` and modifying the gltf files to reference the correct texture. Then using a tool to package to glb.

I hope this can be looked into someday! I can help since I've touched the addon before and perhaps I might look into this myself.

donmccurdy:
I'm not sure about adding WebP support to this addon directly ... there are probably two concerns:

1. We aim to support all official (KHR_) extensions that make sense in Blender, but not vendor-specific extensions, and decide on a case-by-case basis for multi-vendor (EXT_) extensions. So this would need to be discussed, particularly given the more recent releases of WebP2 and AVIF.
2. We can't put arbitrary dependencies into this Blender addon. I don't know the exact restrictions, but my understanding is that native code (or Python bindings for native code) have to be added to Blender directly. I believe Blender already depends on OpenImageIO, which has WebP plugins, so that might be the most natural option. I'm hoping we'll be able to contribute KTX2/Basis support to OIIO at some point in the future.

As a workaround, I'm also planning to add WebP compression support to glTF-Transform, along with optimization codecs for PNG and JPEG. You can find my work-in-progress here: https://github.com/donmccurdy/glTF-Transform/pull/148. That PR is currently blocked on an issue in an upstream dependency (https://github.com/GoogleChromeLabs/squoosh/issues/898).

Zingam:
[...]

Issue #1234, opened 2020-10-08
KTX / DDS support


ArneDJ:
Currently only JPG and PNG are supported, this is a problem for games due to the long loading times and memory that it takes in.

It would be nice if GPU compressed images like KTX or DDS would be supported. Artists could simply export the glTF files with image data embedded without having to manually convert them to a GPU compressed format.

donmccurdy:
Blender itself doesn't support these formats (or at least, [doesn't support DDS well](https://blender.stackexchange.com/questions/111115/can-blender-open-dds-files)?) and adding KTX with Basis would require native dependencies that need to go into Blender core. This could be a good feature request for Blender, but it's not an easy fix.

It's also unfortunately the case that compressing textures (1) is slow and (2) requires some tuning to avoid compression artifacts. Because of that, you may have a better time adding compression after exporting with something like [gltfpack](https://github.com/zeux/meshoptimizer#gltfpack) or [gltf-transform](https://gltf-transform.donmccurdy.com/cli.html). More feedback on good workflows here would be welcome.

ArneDJ:
Thanks for the fast answer, I'll give gltfpack and gltf-transform a try.

UltraEngine:
> Blender itself doesn't support these formats (or at least, [doesn't support DDS well](https://blender.stackexchange.com/questions/111115/can-blender-open-dds-files)?) and adding KTX with Basis would require native dependencies that need to go into Blender core. This could be a good feature request for Blender, but it's not an easy fix.

Blender doesn't need to support DDS files for this feature, and it's better to leave the original images in PNG format. The glTF exporter could compress the image files into the optimal format:

| Name | Format | R | G | B | A |
|---|---|---|---|---|---|
| base | BC7 | base red | base green | base blue | base alpha |
| normal | BC5 | nx | ny | -- | -- |
| occlusion / metallic / roughness | BC7 | ambient occlusion | metalness | roughness | -- |
| emission | BC6H or BC7 | emission red | emission green | emission blue | -- |
| occlusion (alone) | BC4 | ambient occlusion | -- | -- | -- |

[...]

PR #1410, opened 2021-06-08
(WIP) More flexible image export handling for extensions.

netpro2k:
There are now several GLTF extensions for alternate image formats like `KHR_texture_basisu` and  `EXT_texture_webp`. It would be nice to add support for these to to other addons built against this extension without them having to duplicate the node traversal and channel combination/remapping logic already implemented in `ExportImage`.

This is a quick attempt at abstracting things just enough to allow that. I am not particularly happy with this solution but I think its a useful starting point for discussion.

Using this in the Hubs Blender Exporter to support .hdr (which blender already natively understands) files on lightmaps like this:
https://github.com/MozillaReality/hubs-blender-exporter/blob/76579e38ad2b3ad6768eef7b1acf575b2fe4302d/gather_properties.py#L181-L193
https://github.com/MozillaReality/hubs-blender-exporter/blob/76579e38ad2b3ad6768eef7b1acf575b2fe4302d/gather_properties.py#L215

For formats Blender does not already natively support you could instead completely override `encode(mime)`:
```python
ImageData.extension_for_mime["image/ktx2"] = ".ktx"
class BasisExportImage(ExportImage):
    def preferred_mime_type(self) -> str:
        return "image/ktx2"

    def encode(self, mime_type: Optional[str]) -> bytes:
        # use self.fills to generate a ktx2 image
        return bytes()
```
Again this feels a bit messy and its stretching the `ExportImage` abstraction quite a bit, but hopefully serves to show what an addon might want to be able to do. It would likely be a better solution to pull out the bits of `gather_image` and `ExportImage` that are generically useful and have the addon call those instead, skipping `gather_image` entirely, but I wanted to start with a more direct approach just to see the minimal set of changes needed.

---

Some related issues I ran into while looking into this: #1234 #1308

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
@@ -27,14 +27,16 @@ from io_scene_gltf2.blender.exp.gltf2_blender_gather_cache import cached
 from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 
 
-@cached
+# @cached
 def gather_image(
         blender_shader_sockets: typing.Tuple[bpy.types.NodeSocket],
-        export_settings):
+        export_settings,
+        export_image_class = ExportImage
+):
     if not __filter_image(blender_shader_sockets, export_settings):
         return None
 
-    image_data = __get_image_data(blender_shader_sockets, export_settings)
+    image_data = __get_image_data(blender_shader_sockets, export_image_class, export_settings)
     if image_data.empty():
         # The export image has no data
         return None
@@ -93,20 +95,14 @@ def __gather_extras(sockets, export_settings):
 
 
 def __gather_mime_type(sockets, export_image, export_settings):
-    # force png if Alpha contained so we can export alpha
-    for socket in sockets:
-        if socket.name == "Alpha":
-            return "image/png"
-
     if export_settings["gltf_image_format"] == "AUTO":
-        image = export_image.blender_image()
-        if image is not None and __is_blender_image_a_jpeg(image):
-            return "image/jpeg"
-        return "image/png"
+        preferred = export_image.preferred_mime_type()
+        if preferred: return preferred
 
     elif export_settings["gltf_image_format"] == "JPEG":
         return "image/jpeg"
 
+    return "image/png"
 
 def __gather_name(export_image, export_settings):
     # Find all Blender images used in the ExportImage
@@ -122,9 +118,7 @@ def __gather_name(export_image, export_settings):
     if len(filepaths) == 1:
         filename = os.path.basename(list(filepaths)[0])
         name, extension = os.path.splitext(filename)
-        if extension.lower() in ['.png', '.jpg', '.jpeg']:
-            if name:
-                return name
+        return name
 
     # Combine the image names: img1-img2-img3
     names = []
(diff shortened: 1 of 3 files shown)


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
```diff
@@ -27,14 +27,16 @@
 from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 
 
-@cached
+# @cached
```
netpro2k, line 30: export_image_class appears to not be hashable which causes this to fail. Not ideal.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
```diff
@@ -93,20 +95,14 @@ def __gather_extras(sockets, export_settings):
 
 
 def __gather_mime_type(sockets, export_image, export_settings):
-    # force png if Alpha contained so we can export alpha
```
netpro2k, line 96: If the user specifically sets their export mode to jpeg, I think things should always export as a jpeg. When set to auto the new `preffered_mime_type()` check would naturally output as a png unless the input image was already a jpg

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
```diff
@@ -122,9 +118,7 @@ def __gather_name(export_image, export_settings):
     if len(filepaths) == 1:
         filename = os.path.basename(list(filepaths)[0])
         name, extension = os.path.splitext(filename)
-        if extension.lower() in ['.png', '.jpg', '.jpeg']:
-            if name:
-                return name
+        return name
```
netpro2k, line 121: Attempting to remove as much as possible that is assuming a particular filetype. This check seems superfluous anyway.

addons/io_scene_gltf2/io/exp/gltf2_io_image_data.py
```diff
@@ -19,6 +19,13 @@ class ImageData:
     # FUTURE_WORK: as a method to allow the node graph to be better supported, we could model some of
     # the node graph elements with numpy functions
 
+    extension_for_mime = {
```
netpro2k, line 22: We end up with multiple spots with very similar mappings... should be centralized somehow

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
```diff
@@ -27,14 +27,16 @@
 from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 
 
-@cached
+# @cached
 def gather_image(
         blender_shader_sockets: typing.Tuple[bpy.types.NodeSocket],
-        export_settings):
+        export_settings,
+        export_image_class = ExportImage
```
netpro2k, line 34: The logic here is that all of the extensions related to image formats right now actually start at the `texture` level, so its likely they will be calling `gather_image` and can provide an override. Its possible this could instead be handled in reverse via something like a `pre_gather_image_hook` that could return some data, but it would likely require duplicating some functionality.

netpro2k:
For comparison, this is the same thing but built against the existing exporter without modification. In many ways its cleaner, but there is some code duplicated (namely `__encode_from_image`) and I am intentionally skipping many edge cases (unsaved image, compositing multiple image channels, input images in a different format than export format) that ExportImage handles.

https://github.com/MozillaReality/hubs-blender-exporter/blob/hdr-lightmaps/gather_properties.py#L177

julienduroure:
This WIP is 5 years old, without any CLA signed, and Webp is now managed, KTX will come.
Closing this ticket.
Feel free to open a new PR or ticket if you want to discuss new features

PR #1420, opened 2021-06-22
Add merge_animation_extensions_hook

### LINKED ISSUE #1419: Add new `merge_animations_hook`
**Problem**
I am trying to write an add-on that exports frame events from pose_markers. However, if there is more than one animation with the same name, the merge animations feature only uses extension data from the first animation it finds, and throws out the rest, so using `gather_animation_hook` is useless at that point.

**Proposed Solution**
I propose a `merge_animations_hook` that passes in the list of actions to be merged and the name of the NLA track that the actions will be merged into. This will allow us to add extension data to the merged animation without it being thrown away.

I have tried looking into workarounds that would allow me to prepare extension data in advance, but since the NLA track name is never passed in to the `gather_animation_hook`, there is no way I can see to associate the incoming animation with the track it will be merged into without forking the project.

lunarraid:
This adds a new hook called `merge_animation_extensions_hook` to allow add-ons to handle the merging of animation extension data when multiple animations use the same name.

Resolves #1419

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather.py
@@ -116,6 +116,10 @@ def __gather_animations(blender_scene, export_settings):
 
             to_delete_idx.append(anim_idx)
 
+            # Merging extensions
+            # Provide a hook to handle extension merging since there is no way to know author intent
+            export_user_extensions('merge_animation_extensions_hook', export_settings, animations[anim_idx], animations[base_animation_idx])
+
             # Merging extras
             # Warning, some values can be overwritten if present in multiple merged animations
             if animations[anim_idx].extras is not None:
--- a/example-addons/example_gltf_extension/readme.md
+++ b/example-addons/example_gltf_extension/readme.md
@@ -32,4 +32,5 @@ gather_scene_hook(self, gltf2_scene, blender_scene, export_settings)
 gather_skin_hook(self, gltf2_skin, blender_object, export_settings)
 gather_texture_hook(self, gltf2_texture, blender_shader_sockets, export_settings)
 gather_texture_info_hook(self, gltf2_texture_info, blender_shader_sockets, export_settings)
+merge_animation_extensions_hook(self, gltf2_animation_source, gltf2_animation_destination, export_settings)
 ```


julienduroure:
I have no issue with current code of this PR, but some questions:
- Do you have an example of file where you want to change default behaviour ? Because this new hook is inside a loop, not sure every case can be handled with this hook

lunarraid:
Sure. This shouldn't change the default behavior, it just adds an extra listener. Here is part of the extension I've written for my project that uses the hook in question:
```

    def __init__(self):
        # We need to wait until we create the gltf2UserExtension to import the gltf2 modules
        # Otherwise, it may fail because the gltf2 may not be loaded yet
        self.Extension = Extension
        self.properties = bpy.context.scene.AnimationMarkerExtensionProperties
        self.fps = bpy.context.scene.render.fps

    def ensureExtension(self, gltf2_animation):
        if gltf2_animation.extensions is None:
            gltf2_animation.extensions = {}

        if not glTF_extension_name in gltf2_animation.extensions:
            gltf2_animation.extensions[glTF_extension_name] = self.Extension(
                name = glTF_extension_name,
                extension = {},
                required = extension_is_required
            )

        if not 'markers' in gltf2_animation.extensions[glTF_extension_name].extension:
            gltf2_animation.extensions[glTF_extension_name].extension['markers'] = []

    def gather_animation_hook(self, gltf2_animation, blender_action, blender_object, export_settings):

        if not self.properties.enabled:
            return

        if len(blender_action.pose_markers) == 0:
            return

        self.ensureExtension(gltf2_animation)

        markers = gltf2_animation.extensions[glTF_extension_name].extension['markers']

        for marker in blender_action.pose_markers:
            markers.append({ 'name': marker.name, 'time': marker.frame / self.fps })


    def merge_animation_extensions_hook(self, gltf2_animation_source, gltf2_animation_destination, export_settings):

        if not self.properties.enabled:
            return

        if gltf2_animation_source.extensions is None or not glTF_extension_name in gltf2_animation_source.extensions:
            return

        self.ensureExtension(gltf2_animation_destination)

        gltf2_animation_destination.extensions[glTF_extension_name].extension['markers'] += gltf2_animation_source.extensions[glTF_extension_name].extension['markers']
```

The added hook allows me to listen for each merge and control how the merge happens, which is what allows me to gather all of the pose markers from the animations at merge time.

PR #1429, opened 2021-06-30
Don't mangle custom properties on export

Ralith:
Modifies the json export process to leave extras as-is. This is the product of a long debugging session trying to work out why my custom property with a value of `{}` was being silently erased.


julienduroure:
Hello,
Can you please share an example of file you are trying to export, in order to everyone to check if this change can have an impact of their current workflow?

Ralith:
It's just a file with an object with a custom property whose value is intended to be the empty dictionary. This is used for communication with an external application that accepts a number of optional parameters in a dictionary, and for which a property being present with an empty dictionary has *very* different semantics than being absent entirely.

julienduroure (approved):

PR #1433, opened 2021-07-10
Fix #1432 : remove some channel animation if is not animated

### LINKED ISSUE #1432: Always Sample Animations option creates unwanted keyframes on bones
**Describe the bug**
When exporting multiple actions of the same armature, the actions should only contain keyframes on the animated bones.

When using the _Always Sample Animations_ option, the sampling process will keyframe ALL bones in the armature, regardless of the original action containing any keyframes for those bones or not.

The effect of this is that multiple actions do not mix well together anymore when used at the same time.

**To Reproduce**
[GTLF_Sample.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/6779878/GTLF_Sample.zip)

Load the provided GLB in https://sandbox.babylonjs.com/

**Expected behavior**
The GLB contains 2 actions: _Left_ and _Right_

image

You should be able to control the exact position of each action. However when going back to the Left action, the right bone will always reset because it contains a keyframe.

In Blender those keyframes do not exist.

The only workaround is to do a painful manual sample of all actions in Blender in order to be able to disable the _Always Sample Animations_ option and get proper actions without extra bone keyframes.

**System Information**
Operating system: Windows-10-10.0.19041-SP0 64 Bits
Graphics card: NVIDIA GeForce GTX 1080 with Max-Q Design/PCIe/SSE2 NVIDIA Corporation 4.5.0 NVIDIA 471.11

**Blender Version**
Broken: version: 3.0.0 Alpha, branch: master, commit date: 2021-06-28 22:09, hash: `rBe1a719c0433b`
Worked: (newest version of Blender that worked as expected)

julienduroure:
Check is done after sampler
If channel is not directly animated and data are not animated, do not 
create animation for this node property

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_channels.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_channels.py
@@ -126,7 +126,17 @@ def gather_animation_channels(blender_action: bpy.types.Action,
             if len(channel_group_sorted) == 0:
                 # Only errors on channels, ignoring
                 continue
-            channel = __gather_animation_channel(channel_group_sorted, blender_object, export_settings, None, None, bake_range_start, bake_range_end, blender_action.name, None, False)
+            channel = __gather_animation_channel(
+                        channel_group_sorted,
+                        blender_object,
+                        export_settings,
+                        None,
+                        None,
+                        bake_range_start,
+                        bake_range_end,
+                        blender_action.name,
+                        None,
+                        False)
             if channel is not None:
                 channels.append(channel)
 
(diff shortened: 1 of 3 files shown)


scurest:
Yeah, constant + no fcurve is better.

If numpy worth it here? Anyway, [numpy.ptp](https://numpy.org/doc/stable/reference/generated/numpy.ptp.html) (max-min) would be cheaper than numpy.std for detecting constant curves.

julienduroure:
Thanks @scurest for you feedback. I will have a look on ptp

I think we can get a step further for constant + fcurve. Currently we keep all sample keyframes, but I think in this case, we can keep only first one (or maybe first and last one).

This become:
* If animation is not constant:
  * keep all sampled keyframes
* if animation is constant:
  * if no fcurve, skip this animation
  * if fcurve, keep only first keyframe (or first and last)

This will lead to smaller files.
Any thoughts?

julienduroure:
ok, implemented :)

Using robot, file down from 1.3Mo to 551Ko :)

image


... Any reason to not apply this to object animation and morh/SK anim too?

Issue #1440, opened 2021-08-03
.gltf file huge after export


anzmaaan:
Hello!

Exporting my selected objects from Blender (some of the objects came from Cinema4D, rest was created in Blender), I get a file that is surprisingly large to what I expected it to be  - in terms of texture file sizes (all textures are jpgs, total 5 Megabytes) and polygon count (82k for the whole scene), there are no animations, blend shapes etc. just static meshes. Upon opening the .gltf in VS Code, the file reveals a large chunk of data at the end (see attached file) which goes on and on and represents the binary embedded as of my understanding. I can't really make sense of the cause for the size, maybe somebody can help. Ultimately I need a .glb file <10 Megabytes. Currently I'm looking at about 22 Megabytes for a glb-export of my scene. 

Kind regards and thanks for looking into it!
[issue.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/6924725/issue.zip)

scurest:
* Never ever use glTF Embedded. It includes all binary data as base64 text, which inflates its size by 33%.
* Your images alone are 11MB, you have to downsize them or save them at lower quality.
* You have a bunch of unused UVMaps. In particular, "geo_aussenzaun.001_Plane.005" has 7 unused UVMaps taking up 6MB. Delete them.
* Run files through [gltf-pack](https://github.com/zeux/meshoptimizer/blob/master/gltf/README.md) for smaller filesize.

anzmaaan:
Thanks a lot for your tips and your quick reply!

PR #1448, opened 2021-08-13
Change glTF format order

julienduroure:
As proposed by Alexey Panteleev ( @more_fps on twitter), change order of option for gltf format
Embedded is probably the less used, because generate bigger file. So put it at end


donmccurdy:
I'm in favor of (at least) this reordering. 

I'd also be OK with removing this option, or perhaps marking it deprecated — see https://github.com/KhronosGroup/glTF-Blender-IO/issues/1440#issuecomment-892126951.

Issue #1313, opened 2021-01-16
Unselected objects get exported


Ali-RS:
Hi!

I am encountering a model exporting issue.

**Describe the bug**
I have a scene with one tree model and one animated horse model. I want to export only the tree model, so I select it only and upon export, I enable the "**Selected Objects**" option but it also exports an empty node for each bone of the horse.

**To Reproduce**
1. Open the attached file with blender **2.91.0**
2. Select tree model and export it to gltf (with "**glTF Separate**" format and "**Selected Objects**" enabled)
3. Open the exported gltf file and you will notice it contains extra nodes with the name "Bone0XX" which belong to the horse rig.

**Expected behavior**
It should only export the selected tree model.

**.blend file/ .gltf**

[model.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/5824220/model.zip)

**Version**
 - OS: linux mint
 - Blender Version 2.91.0

Ali-RS:
Anyone can confirm this, please?

scurest:
This is #1298.

Ali-RS:
Ok, thanks. For now, going to put them in separate blend files, until this bug is fixed.

julienduroure:
Merged

PR #1453, opened 2021-08-20
Using "selection only" option, better handle armature checking

### LINKED ISSUE #1298: Exporter always exporting every armature present in the blend file
**Describe the bug**

I'm currently working on a blend file containing multiple rigged assets. When ready, each assets then gets exported using the GLTF exporter.

I noticed that when they get imported, the exported asset contains the expected output plus a  "dummy" armature for each armature present in the blend file. This also happens whether or not "selection only" is enabled in the export settings.

A blend file containing multiple rigged meshes 

**To Reproduce**
1. Create a blend file containing several armatures and rigged meshes
image

2. Export one of the rigged meshes using the gltf exporter (i tried multiple settings combination and all of them gave the same behaviour)
image
image

3. Import the resulting file somewhere (here in the Godot game engine)
image

The object is imported correctly and the skinning is also right, but it also contains the other armatures that were in the blend file.

4. Import the file back in Blender using the GLTF importer to check if it's not a bug on the game engine's side
image

The imported file clearly contains "dummy" armatures made of empties.

**Expected behavior**
The exporter should only export the selected armatures when "selection only" is enabled. In its current state, it exports more data than it should.

**Screenshots**
(see the reproduction steps)

**.blend file/ .gltf**
[demo_file.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/5711745/demo_file.zip)

**Version**
 - OS: Windows 10
 - Blender Version 2.91

julienduroure:
Waiting #1378 to be fully implemented, trying to check if armature should be exported or not hen using "selection only" option.
Checking if children of armature are only bone or not

Fix #1298
Fix https://developer.blender.org/T84506
Fix https://developer.blender.org/T87093

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
@@ -49,7 +49,7 @@ def gather_node(blender_object, library, blender_scene, dupli_object_parent, exp
 
 @cached
 def __gather_node(blender_object, library, blender_scene, dupli_object_parent, export_settings):
-    children = __gather_children(blender_object, blender_scene, export_settings)
+    children, only_bone_children = __gather_children(blender_object, blender_scene, export_settings)
 
     camera = None
     mesh = None
@@ -64,6 +64,12 @@ def __gather_node(blender_object, library, blender_scene, dupli_object_parent, e
             # This node should be filtered out, but has un-filtered children present.
             # So, export this node, excluding its camera, mesh, skin, and weights.
             # The transformations and animations on this node will have visible effects on children.
+
+            # Armature always have children node(s) (that are bone(s))
+            # We have to check if children are only bones or not for armatures
+            if blender_object.type == "ARMATURE" and only_bone_children is True:
+                return None
+
             pass
         else:
             # This node is filtered out, and has no un-filtered children or descendants.
@@ -152,6 +158,7 @@ def __gather_camera(blender_object, export_settings):
 
 def __gather_children(blender_object, blender_scene, export_settings):
     children = []
+    only_bone_children = True # True by default, will be set to False if needed
     # standard children
     for _child_object in blender_object.children:
         if _child_object.parent_bone:
@@ -167,6 +174,7 @@ def __gather_children(blender_object, blender_scene, export_settings):
             blender_scene, None, export_settings)
         if node is not None:
             children.append(node)
+            only_bone_children = False
     # blender dupli objects
     if blender_object.instance_type == 'COLLECTION' and blender_object.instance_collection:
         for dupli_object in blender_object.instance_collection.objects:
@@ -179,6 +187,7 @@ def __gather_children(blender_object, blender_scene, export_settings):
                 blender_scene, blender_object.name, export_settings)
             if node is not None:
                 children.append(node)
+                only_bone_children = False
 
     # blender bones
     if blender_object.type == "ARMATURE":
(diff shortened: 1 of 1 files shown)


julienduroure:
Fix #1313

julienduroure:
@donmccurdy @emackey @scurest 
Any feedback / tests are welcome on this PR.
This is a quick hack/fix, waiting #1378 to be ready & stable.
Thanks !

scurest:
If an armature has any objects parented to one of its bones, it will still be exported even if unselected, since `direct_bone_children` will be nonempty. Probably it needs to actually `gather_node` them like it does for standard children so it knows if they're actually going to be exported or not.

If you select an object which is a descendant of an unselected arma, all the arma's bones get exported even if unused.

If you select a mesh with an arma modifier, but don't select the arma, and the arma isn't an ancestor of the mesh, weird things happen. Maybe better to save that for vtree though.

julienduroure:
Agree that this PR is not perfect.
You spotted all cases where the situation do not change with or without this PR

For information, the only case that I managed is:
- when there is multiple armature and skinned meshes, and user select only 1 skinned meshes. Other armature are no more exported.

About cases you spotted:
1. You'r probably right, but not sure it's worth it to try making it work or spend time on vtree
2. You mean if descendant mesh is not skinned to this armature?
3. Yes, agree to keep this case away for now. Not parenting the skinned mesh to armature is not something you should do in Blender :) ("Standard" rigging process parents it) 

Thanks for your feedback!

soelen:
> @donmccurdy @emackey @scurest
> Any feedback / tests are welcome on this PR.

-[x] Downloaded fix_1298_armature_check_selection branch and installed io_scene_gltf2 addon for blender 2.93.1.
-[x] Exported only selected, simple objects like cubes from a blender file which includes armatures -> No armature in gltf file found.
- Exported complex object with armature -> Everything exported as expected, bone manipulation in my threejs based project still works as expected.

Neat!

julienduroure:
Seems this PR solve the issue when selection does not include armature or any descendant of an armature.
For any other case, there is still issue, but no regression regarding master (= same behaviour).
So I think we can merge this, to at least solve 1 issue, even if other ones will probably need to wait #1378 ( or gather_node, that will complexify once again current code)

rageratwork:
If I understand correctly, this PR does not cover simple objects that do not have an armature, but one or more other objects do.

I am exporting a simple cube that I just added to the scene. When I export it, the file also contains 2 armatures from other unselected objects.

I'm not seeing the behavior @soelen is describing in their second bullet item. Based on other comments, this case is not actually fixed in this PR. Is this correct?

[test.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/7231957/test.zip)

julienduroure:
@rageratwork Please provide the .blend file
I can't say anything without the file you are trying to export

rageratwork:
Here is a test blend file with a simple cube and my rigged character. I'm also including the exported gltf and a screen cap of the export settings. Thanks for having a look!

[test.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/7232410/test.zip)

julienduroure:
You are in case described by scurest, that is not managed by this PR:
> If an armature has any objects parented to one of its bones, the arma will still be exported even if unselected, since direct_bone_children will be nonempty

acatdc:
Guys, I checked and it worked well in 2.90 release , [here is the link](https://download.blender.org/release/Blender2.90/)
And was broken from 2.91 release and further.

julienduroure:
Note that there is no simple solution to solve all these cases with the current architecture.
This will be solved by #1378

Issue #1466, opened 2021-09-13
Cleanup animation when sampling, and object is not animated


julienduroure:
Based on 6dc68610e249f276cfd82c9f6638b81b08d2745b, we could clean some channel on objects too

PR #1467, opened 2021-09-15
Fix #1466

julienduroure:
If a channel has only constant values, keep only first and last one

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
@@ -325,6 +325,11 @@ def gather_keyframes(blender_object_if_armature: typing.Optional[bpy.types.Objec
         else: # bone is not animated (no fcurve)
             # Not keeping if not changing property
             return None if std < 0.0001 else keyframes
+    else:
+        # For objects, if all values are the same, we keep only first and last
+        std = np.ptp(np.ptp([[k.value[i] for i in range(len(keyframes[0].value))] for k in keyframes], axis=0))
+        return [keyframes[0], keyframes[-1]] if std < 0.0001 and len(keyframes) >= 2 else keyframes
+
 
     return keyframes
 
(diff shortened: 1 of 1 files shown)


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
```diff
@@ -325,6 +325,12 @@ def gather_keyframes(blender_object_if_armature: typing.Optional[bpy.types.Objec
         else: # bone is not animated (no fcurve)
             # Not keeping if not changing property
             return None if std < 0.0001 else keyframes
+    else:
+        # For objects, if all values are the same, we keep only first and last
+        # TODO We need to differenciate morph anim / TRS anim
+        std = np.ptp(np.ptp([[k.value[i] for i in range(len(keyframes[0].value))] for k in keyframes], axis=0))
+        return [keyframes[0], keyframes[-1]] if std < 0.0001 and len(keyframes) >= 2 else keyframes
```
scurest, line 332:
```py
ptps = np.ptp([[k.value[i] for i in range(len(keyframes[0].value))] for k in keyframes], axis=0)
```

This is an array of the ptp for each component. It's zero where one component is constant over time.

```py
np.ptp(ptps) < 0.0001
```

This part is wrong. It checks if the ptp of the ptps is zero, ie. all the ptps are the same. It should be checking if all the ptps are (approximately) zero. For example

```py
np.ptp(np.ptp( [[0,4], [3, 7]], axis=0))
```

is zero but the curves aren't constant.

scurest: Seems #1433 had the same problem. Why don't you pull this "is constant" check out into a function?

scurest: Seems #1433 had the same problem. Why don't you pull this "is constant" check out into a function?

julienduroure: Thanks for your feedback. 
I got the same conclusion during my tests.

PR #1480, opened 2021-10-27
Add placeholder materials on export

pkowal1982:
Add placeholder materials on export.
Create very simple material with only name and diffuse set so if mesh has more than one material applied they can be distinguished.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
@@ -40,6 +40,9 @@ def gather_material(blender_material, export_settings):
     if not __filter_material(blender_material, export_settings):
         return None
 
+    if export_settings['gltf_materials'] == 'PLACEHOLDER':
+        return __gather_placeholder_material(blender_material, export_settings)
+
     mat_unlit = __gather_material_unlit(blender_material, export_settings)
     if mat_unlit is not None:
         return mat_unlit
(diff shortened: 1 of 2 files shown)


julienduroure:
Current master has 3 options:
- Export: Exports all used materials, 1 primitive by materials
- Placeholder: Exports no materials, but primitives split by blender materials are kept
- No export: Exports no materials, primitives are merged into a single one

This PR change the Placeholder option, by including basic material (only diffuse color).

My question is: Should we change this placeholder behavior, or add this feature under a 4th option?

julienduroure:
We also should discuss what is the "minimal" material that could be used for placeholder : only diffuse color, or full Blender material, but without any textures?

optimus007:
> Add placeholder materials on export.
> Create very simple material with only name and diffuse set so if mesh has more than one material applied they can be distinguished.

Maybe a  "PBR no images" option which exports principled bsdf with whatever numeric values are available for export.


My use case will be:
- I create the model in blender with textures for visualisation .
- I export the imageless glb and load it on three js 
- Based on the download speed and fps on the client system ,  i load 512 or  1k or 2k versions of the textures on the glb 



So my main requirements are just to have a glb with the principled material set ( which gets converted to [meshStandardMaterial](https://threejs.org/docs/index.html?q=mesh#api/en/materials/MeshStandardMaterial)  automatically on three js ) and UV maps maintained

julienduroure:
What I propose:
- rename current "placeholder" to "primitive split"
- Create a 4th option "placeholder" with primitive split + minimal material (that can be a PBR no image for example)

Any thoughts?

julienduroure:
Note that #1481 does it another way, setting "None" in Images dropdown menu.

julienduroure:
I propose #1546 to solve this requirement 
- No new option for materials
- But a new option "None" to image export
  - Materials are exported without textures
  - Default socket color are kept

julienduroure:
I will probably close this PR in favor of #1546

pkowal1982:
If #1546 will export "full" material omitting only textures it's fine for me. :)

GRAFIX21:
I was very happy that the solution #1480 works in Blender 3.0. Unfortunately, this does not work in Blender 4.0 because the script structure looks different here.

Unfortunately, the idea with the placeholder in Blender does not seem to have been thought through to the end, because I would assume that this is exactly what the result should be.

My application scenario is: I export various models from Blender as GLB files and only need simple materials in them, or rather only the material names as PLACEHOLDERS. This is because the materials are configured with the same name in my web viewer and are only loaded automatically when they are needed. This saves a lot of memory and allows me to design the materials with different levels of complexity depending on the output device (e.g. on smartphones without complex maps).

Can someone document the necessary changes for Blender 4.0?

Issue #770, opened 2019-11-13
Option to ignore textures at export


ZanQdo:
I need to export the geometry + materials but with no textures, since the textures are added back later on in the pipeline. The problem is some of the materials come with textures assigned to them, which means I have to constantly go into the shader view and delete the textures in all materials in order to avoid exporting them.

**Describe the solution you'd like**
My proposal is to either add a new Images option below the Materials option or to add a None option in the Images enum menu.

![image](https://user-images.githubusercontent.com/999476/68794226-4120c280-0614-11ea-8c0c-500e11df121c.png)

This way it would be easy to export geo with material groups but making sure no image data is packed in.

DolphinIQ:
I think this option would be very useful, because lots of people use .basis compressed textures in web apps nowadays. Since the exporter doesn't support image copression, users have to either bloat the file size (and uncompressed textures increase it by A LOT) or manually remove all the image nodes from every material before export.

AndrewRayCode:
Another issue is if a mesh has multiple materials on it, the exporter creates one geometry (primitive) per material, but then doesn't associate any material information with it, so you can't re-create it in a downstream consumer.

There apparently was a PR made to do just this about a year ago: https://github.com/KhronosGroup/glTF-Blender-IO/pull/453

@yoanlcq I don't suppose you ever ended up revisiting that PR, did you?

donmccurdy:
>  the exporter creates one geometry (primitive) per material, but then doesn't associate any material information with it, so you can't re-create it in a downstream consumer.

I don't think I understand this part... each primitive will still have materials associated. If that's not what you're seeing you may have a separate bug?

yoanlcq:
My PR (#453) isn't really a solution for OP, but rather for the "texture sharing" case described by @DolphinIQ , which IMHO would need a separate ticket, given how simple OP's request is.

Indeed I have not kept up with the latest changes, so I suspect it could not be merged today without major changes; I'm still using the feature in my personal project though, and it works great for my purposes.  

"texture sharing" was suggested as a post-processing step, but IMO this is too late in the pipeline to do it (you'd need to work out which texture files are "equal" after exporting your multiple GLTFs), especially since all file paths are readily available in Blender, along with whether or not they are relative or absolute. At the very least, they could be provided in JSON via `extras`.  

... And, while we're at it, we could choose not to export the textures, and just leave them where they are, assuming the application would use the provided file path.

[...]

PR #1481, opened 2021-10-28
Add an export option to omit textures

jtnicholl:
Was requested in #770


julienduroure:
I proposed some enhancement in #1546

PR #1486, opened 2021-11-05
Fix #1484 Vertex group

### LINKED ISSUE #1484: T92808 : import skinning is broken (vertex group X%)
See https://developer.blender.org/T92808

julienduroure:


julienduroure:
@scurest can you please have a look on it?

scurest:
Looks right to me.

You should also bump the blender version in `bl_info` up to 3.0 since the addons won't work in <3.0 anymore.

PR #1498, opened 2021-12-11
Fix T93704

julienduroure:


julienduroure:
This PR will probably not be usefull, as root cause is now fixed in Blender.
Closing.
I will double check that the root fix will be included in 3.0.1 (This is already list in potential candidates)

PR #1512, opened 2021-12-20
Fix #1508 - Better jpeg image detection

### LINKED ISSUE #1508: Export: import embedded JPEG -> exports as PNG
If you import a GLB with embeded JPEGs (say, DamagedHelmet.glb), they will now be re-exported as PNGs, not as JPEGs. This is a regression after merge of #1499.

The reason is the exporter does not correctly detect that certain packed images are JPEGs. This problem already existed, but it is worse now because it affects images created by the glTF importer too.

----

Proximal cause is `__is_blender_image_a_jpeg` does not handle images that are packed without having a filename.

https://github.com/KhronosGroup/glTF-Blender-IO/blob/3bc2858c714d5b517aa521a632906934fafdd4a8/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py#L261-L265

Simplest fix is if there's a packed file, check the first three bytes of the packed data for the JPEG magic.

julienduroure:
Fix #1508 - Better jpeg image detection

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
@@ -261,7 +261,7 @@ def __get_tex_from_socket(blender_shader_socket: bpy.types.NodeSocket, export_se
 def __is_blender_image_a_jpeg(image: bpy.types.Image) -> bool:
     if image.source != 'FILE':
         return False
-    if image.filepath_raw == '':
+    if image.filepath_raw == '' and image.packed_file:
         return image.packed_file.data[:3] == b'\xff\xd8\xff'
     else:
         path = image.filepath_raw.lower()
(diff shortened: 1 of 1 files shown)


julienduroure:

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_image.py
```diff
@@ -261,5 +261,8 @@ def __get_tex_from_socket(blender_shader_socket: bpy.types.NodeSocket, export_se
 def __is_blender_image_a_jpeg(image: bpy.types.Image) -> bool:
     if image.source != 'FILE':
         return False
-    path = image.filepath_raw.lower()
-    return path.endswith('.jpg') or path.endswith('.jpeg') or path.endswith('.jpe')
+    if image.filepath_raw == '':
```
scurest, line 264: Think you need `and image.packed_file`. Otherwise it would throw when you try to do `.data` on None.

PR #1513, opened 2021-12-20
Fix #1509 images may be needlessly re-encoded in some cases

### LINKED ISSUE #1509: Export: images may be needlessly re-encoded in some cases
Look at this code for getting the already-encoded data for an image

https://github.com/KhronosGroup/glTF-Blender-IO/blob/3bc2858c714d5b517aa521a632906934fafdd4a8/addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py#L192-L200

This assumes `image.file_format` describes the content of the file on disk/packed file. But that's wrong, `image.file_format` is only the format to use when saving, cf #986. The file_format check should be removed.

To avoid an unnecessary file read in the case of a file on disk, a check for the file extension could be added.

Re-encoding would be especially bad for JPEGs, since it would be lossy.

julienduroure:
Fix #1509

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_image.py
@@ -189,8 +189,7 @@ class ExportImage:
     def __encode_from_image(self, image: bpy.types.Image) -> bytes:
         # See if there is an existing file we can use.
         data = None
-        if image.source == 'FILE' and image.file_format == self.file_format and \
-                not image.is_dirty:
+        if image.source == 'FILE' and not image.is_dirty:
             if image.packed_file is not None:
                 data = image.packed_file.data
             else:


julienduroure:

scurest:
Yeah, I guess it should never re-encode now. You didn't want to try using the filepath to avoid unnecessarily fetching image data?

julienduroure:
> You didn't want to try using the filepath to avoid unnecessarily fetching image data?

Sorry, I am not sure to follow you, and a bit lost on what check/fetch exactly you want to avoid

scurest:
The point of the file_format check was if the file isn't the right type, you can avoid reading the entire file just to find out it's the wrong type and you can't use it.

Ideally even an unneeded `packed_file.data` access would be avoided, since every time that's accessed the data is copied over into a new object on the Python side. But that's less bad than unneeded file IO.

PR #1514, opened 2021-12-20
Fix #1510 - use relative path when possible

### LINKED ISSUE #1510: Import: respect "Default To Relative Paths" preference
When the .blend file has been saved, and `context.preferences.filepaths.use_relative_paths` is on, the importer should load images using relative paths (`//foo.png`), instead of absolute.

[Some relevant OBJ code](https://github.com/sobotka/blender-addons/blob/c60fef38175ad989ee0c45e924cb27e1417c8667/io_scene_obj/__init__.py#L147-L149)
[An old issue for the same thing in OBJ](https://developer.blender.org/T34864)

julienduroure:
Fix #1510 - use relative path when possible

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_image.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_image.py
@@ -48,11 +48,15 @@ class BlenderImage():
             if img.uri is not None and not img.uri.startswith('data:'):
                 # Image stored in a file
                 path = join(dirname(gltf.filename), _uri_to_path(img.uri))
+                path = os.path.abspath(path)
+                if bpy.data.is_saved and bpy.context.preferences.filepaths.use_relative_paths:
+                    path = bpy.path.relpath(path)
+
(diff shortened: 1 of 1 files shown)


julienduroure:

scurest:
Just one thing: on Windows, a relative path may not exist (eg. between `C:\file.blend` and `D:\image.png`). In that case, `bpy.path.relpath` raises a ValueError. So in the if-body I would do

```py
try:
    path = bpy.path.relpath(path)
except ValueError:
    # May happen on Windows if on different drives, eg. C:\ and D:\
    pass
```

This still uses the absolute path if no relative path exists, which is what happens if you load the image by hand.

scurest:
Oh, you should really get rid of the outer `try` that swallows all exceptions too though. It existed to guard the tempdir with a finally. Since there's no tempdir it's not needed and swallowing all errors is bad.

scurest:
The relpath code is messed up. There's no abspath and there's two relpaths now.

Issue #1232, opened 2020-10-05
Export Vertex Groups


canadaduane:
**Is your feature request related to a problem? Please describe.**
In a three.js game, we would like to be able to tell the "hair vertices" from the "skin" vertices of an avatar (clothing, shoes, etc.) so that we can modify the colors of the vertices in-game. To accomplish this, we'd like to be able to see the vertex groups as they exist in Blender; however, it appears vertex groups are not currently exported. 

**Describe the solution you'd like**
When an object's vertices are grouped via Object Data Properties -> Add Vertex Group, we would like to see these vertex groups in the exported glTF file:

<img width="857" alt="image" src="https://user-images.githubusercontent.com/129/95141743-39d43d00-072f-11eb-95e2-da9f5e27e253.png">

**Describe alternatives you've considered**
Multiple UV maps. Dead end due to the way three.js uses UV maps.

**Additional context**
This file has a vertex group called "vertexgroup" and when exported as a glTF, the vertex group is not present:

[vertexgroup.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/5330594/vertexgroup.blend.zip)

scurest:
How do you want to see it in the glTF?

canadaduane:
I'm open to just about anything, but one way that seems congruent with a previous blender/three.js exporter is to export as an array attribute:

https://github.com/repsac/io_three/blob/56b2f35b7d56aab6df131c7a3aecb43cda915ca3/addons/io_three/exporter/geometry.py#L409

(extra_vertex_groups defined here: https://github.com/repsac/io_three/blob/master/addons/io_three/exporter/api/mesh.py#L182)

(See also a long ago discussion at https://stackoverflow.com/questions/38451346/blender-buffergeometry-access-vertices-from-vertex-group)

It was a long time ago though (5 years). Does that seem reasonable today?

scurest:
[This branch](https://github.com/scurest/glTF-Blender-IO/tree/custom-data) should export a SCALAR attribute array for every non-skinning vertex group.

canadaduane:
Wow, that was quick! I'm unfamiliar with testing glTF-Blender-IO but I will give it a shot and report back. Thanks!

canadaduane:
Thanks again for this addition @scurest!

After looking at the output, I think the export could be used in its current state. However, is it exporting data in a format that uses more memory than it needs to?

[...]

Issue #1505, opened 2021-12-16
Add mesh generic attribute as custom attributes in glTF


julienduroure:
With Geometry Nodes, we now have some generic data that are linked to vertex/faces/etc...

Can be retrieved with mesh.attributes. That would give any attribute created by geometry nodes or elsewhere
The spreadsheet displays the attributes, along with a few others like position that are really stored in "legacy" places like MVert

This can be added as custom accessor in attributes (on primitive). See https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#meshes-overview

julienduroure:
This will be managed in #1676 
Need to check that GN attributes are also exported

Issue #1245, opened 2020-10-21
Better Cache system


julienduroure:
Let's use this ticket to summarize bugs/enhancement/open questions regarding cache system

* by-name / by-value for any bpy.types objects (see #1212 discussion)
* Mesh issue that have same name, when linked (see #1210, #1211 )
* library is currently part of the cache key, but maybe #1244 will change that
* cache on material even if material export is disable
* modifiers and vertex group are currently not well cached.
  * Instances with modifiers can be wrongly caches (no way to make difference on 2 objects with same mesh but different modifiers or different parameters on modifiers)
  * Vertex Group are not pass as argument when "not needed" currently, but are needed in some cases (see #900 #747)

julienduroure:
After some experiments:
- Seems that caching by id solve lots of issues (and we don't need check on library anymore for all issues on link)
- Cache on object is more difficult, because of instances. Seems id system is not stable yet (some recent commit on it on blender side), but ... Why do we really need caching on object ? Because of TRS matrix and parent, every object is different. Currently, this is only needed for animation, in order to retrieve the animated object jdon node index. Maybe we can find another way (than caching) for retrieving this info?

julienduroure:
Regarding object caching --> Not that simple for bones, because cache is also used for skin / joint
Investigation still in progress

julienduroure:
Some of these issues are fixed in #1378 (still in WIP)

julienduroure:
Fixed by 69746b2b31eeee080d5dfd3c74f1e02742795ed2

PR #1515, opened 2021-12-22
Export: write custom attribute arrays for non-skinning vgroups

canadaduane:
I've combined the facemap exports and vertex group exports from @scurest's work last year, and tested that this works on 3.0.0.

I haven't followed all of the changes and improvements since Nov 2020, however, so I'm not sure if this is still the ideal way to do things.

There are also a couple of places where a line of code has been commented out (e.g. gltf2_blender_gather_nodes.py has `#vertex_groups = None` in a couple of places). We should probably decide what we want to happen--either remove it, or improve it?

#1232

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
@@ -286,9 +286,12 @@ def __gather_extensions(blender_object, export_settings):
 
 
 def __gather_extras(blender_object, export_settings):
+    extras = {}
     if export_settings['gltf_extras']:
-        return generate_extras(blender_object)
-    return None
+        extras = generate_extras(blender_object) or {}
+    if blender_object.face_maps:
+        extras['facemaps'] = [face_map.name for face_map in blender_object.face_maps]
+    return extras or None
 
 
 def __gather_matrix(blender_object, export_settings):
@@ -339,12 +342,12 @@ def __gather_mesh(blender_object, library, export_settings):
         skip_filter = False
         # If no skin are exported, no need to have vertex group, this will create a cache miss
         if not export_settings[gltf2_blender_export_keys.SKINS]:
-            vertex_groups = None
+            #vertex_groups = None
             modifiers = None
         else:
             # Check if there is an armature modidier
             if len([mod for mod in blender_object.modifiers if mod.type == "ARMATURE"]) == 0:
-                vertex_groups = None # Not needed if no armature, avoid a cache miss
+                #vertex_groups = None # Not needed if no armature, avoid a cache miss
                 modifiers = None
 
     materials = tuple(ms.material for ms in blender_object.material_slots)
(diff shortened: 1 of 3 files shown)


julienduroure:
> There are also a couple of places where a line of code has been commented out (e.g. gltf2_blender_gather_nodes.py has #vertex_groups = None in a couple of places). We should probably decide what we want to happen--either remove it, or improve it?

We can't remove it, it will lead to miss cache. 
The solution is to use a custom cache, as I started to implement (based in scurest suggestion) in vtree branch.
This will be nedded also for other requests (see #900 #747 #1245)

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_nodes.py
```diff
@@ -339,12 +342,12 @@ def __gather_mesh(blender_object, library, export_settings):
...
         # If no skin are exported, no need to have vertex group, this will create a cache miss
         if not export_settings[gltf2_blender_export_keys.SKINS]:
-            vertex_groups = None
+            #vertex_groups = None
             modifiers = None
         else:
             # Check if there is an armature modidier
             if len([mod for mod in blender_object.modifiers if mod.type == "ARMATURE"]) == 0:
-                vertex_groups = None # Not needed if no armature, avoid a cache miss
+                #vertex_groups = None # Not needed if no armature, avoid a cache miss
```
julienduroure, line 350: This can lead to miss cache

julienduroure (changes requested):

canadaduane:
Ok, cool, so this PR should wait until your work there on custom caches is done? If I understand #1245 is the ticket to watch.

scurest:
It does not require a custom cache. It will also not cause a cache miss: if there are non-skinning vgroups, this branch _always_ exports them, so since the vgroups are always used, it is correct that they are always part of the cache key. But I don't think it is good that they are always exported. If it's desirable to merge these features into mainline, I think they should be behind an export option, cf https://github.com/KhronosGroup/glTF-Blender-IO/issues/1232#issuecomment-705290610.

julienduroure:
BTW, some other custom attribute can be exported, as requested by some blender devs. See #1505

Anyway, now that vertex groups are stored on mesh, no more on object, all this code needs to be refactored.

scurest:
> Anyway, now that vertex groups are stored on mesh, no more on object, all this code needs (probably) to be refactored.

Oh, that's right. `vertex_groups = None` is always unnecessary now, since two instances of the same mesh always have the same vertex groups. It still has to be passed down though since there's no API for getting it off the mesh.

iuriiiurevich:
> Plus one to the idea of adding this feature to the official exporter (as an export option)!

@julienduroure Are there any plans to do so?
By the way, thanks for your amazing work!

julienduroure:
This PR is quite old, CLA is not signed, facemaps is not exists anymore, and the PR has many conflicts
Closing this ticket.
Feel free to open a new ticket or PR if you want

PR #1524, opened 2022-01-01
refactor: move export_settings gather to seperate function

pepperoni505:
I'm working on an addon and I need to override the `execute` function in `ExportGLTF2_Base`, but I don't want to repeat 100 lines of code, so I moved the code to generate `export_settings` to a separate function, which turns the `execute` function into only 3 lines. This may not be ideal, so if there is an easier way to override the execute function without this PR, I will close it.

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -506,10 +506,10 @@ class ExportGLTF2_Base:
             del export_props['export_selected'] # Do not save this property, only here for backward compatibility
         context.scene[self.scene_key] = export_props
 
-    def execute(self, context):
+
+    def get_export_settings(self, context):
         import os
         import datetime
-        from .blender.exp import gltf2_blender_export
 
         if self.will_save_settings:
             self.save_settings(context)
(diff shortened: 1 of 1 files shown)


julienduroure:
I have no issue with the changes, but some questions about the root of these changes.
You spoke about an addon that need to override _execute_: do you talk about a glTF user extension addon, or a regular Blender addon?
What are you trying to do than is currently not possible with glTF exporter - and can't be enhance with a user extension addon?

Thanks!

pepperoni505:
It's a glTF user extension addon. I want to add batch export capabilities, so the user can specify which root objects in the scene to export. I first tried a completely separate menu in File -> Export, but because I still needed the user to be able to choose the glTF options, it wasn't possible without copying all panel code. The reason I decided to override the export was to loop over all the objects the user manually chose and call the exporter object by object, instead of having the exporter export everything.

pepperoni505:
Another possible solution would be to add another export setting that defines files that are being exported, like maybe a dictionary of export paths with objects in the scene to export as the value. Just a quick idea

pepperoni505:
@julienduroure I think I have a better way of doing this - closing this PR in favor of #1530

PR #1530, opened 2022-01-08
feat: add export plan

pepperoni505:
This proposal adds the boilerplate code needed to add batch exporting.

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -521,12 +521,20 @@ class ExportGLTF2_Base:
 
         export_settings['timestamp'] = datetime.datetime.now()
 
-        export_settings['gltf_filepath'] = self.filepath
-        export_settings['gltf_filedirectory'] = os.path.dirname(export_settings['gltf_filepath']) + '/'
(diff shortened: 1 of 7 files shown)


julienduroure:
Hello,
Seems this solution is better than the previous one.
But loop on objects is only done on "root objects", so I don't know how you are managing plan where objects have parents?
Same for animated objects

Not that a big refactoring of export will come in next months, where tree is first store in a virtual vtree, and then exported. 

(See #1494 for what it will be)
I will also add some hook to be able to change the vtree filtering.

pepperoni505:
I'll fix that. What should I do in a situation where a root object has a grandchild, but the parent of that object isn't exported?

e.g.
- Object 1 (export)
    - Object 2 (don't export)
        - Object 3 (export)

julienduroure:
In current master, object 2 is exported as empty. 

In future vtree version, object 3 is reparented to object 1, and object 2 is not exported

pepperoni505:
So in this instance, would it be best to just wait until the vtree PR gets merged ?

pepperoni505:
Closing this to wait for the vtree PR.

PR #1546, opened 2022-01-22
Option to not export textures

julienduroure:
Add some enhancement to jtnicholl PR

- moving check to filter
- add default socket values if texture is not exported


julienduroure:

PR #1553, opened 2022-02-02
feat: add more hooks to gather animations

pepperoni505:
This PR adds two hooks to the gather animations process:

- `pre_gather_animation_hook(self, gltf2_animation, blender_action, blender_object, export_settings)`
- `gather_actions(self, blender_object, blender_actions, blender_tracks, action_on_type, export_settings)`
- `pre_gather_actions_hook(self, blender_object, export_settings)`

It is aimed at giving add-ons the ability to gather material animations


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
```diff
@@ -258,6 +259,34 @@ def __get_blender_actions(blender_object: bpy.types.Object,
                         blender_tracks[strip.action.name] = track.name # Always set after possible active action -> None will be overwrite
                         action_on_type[strip.action.name] = "SHAPEKEY"
 
+    # Material animation - not actually used by the base exporter, but useful for addons that take advantage of it.
```
julienduroure, line 262: I don't like so much to have "dead code" that is not used by the vanilla exporter. Maybe we can add a hook here to be able to manage this kind on export data, where you can retrieve these needed information for your extension?

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
```diff
@@ -122,14 +122,15 @@ def __gather_animation(blender_action: bpy.types.Action,
         print_console("WARNING", "Animation '{}' could not be exported. Cause: {}".format(name, error))
         return None
 
-    if not animation.channels:
-        return None
-
-    # To allow reuse of samplers in one animation,
-    __link_samplers(animation, export_settings)
+    if animation.channels:
```
julienduroure, line 125: linking sampler here, before the hook, make the hook data available different of what we have before. Not sure we can change this, now that people know (and use?) this hook.
Solution: create another hook (so one before, and one after the link_samplers process?)

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
pepperoni505: Yeah, you make a great point. I just updated the PR to switch to utilizing hooks. One question though, on the `gather_actions` hook I added, if the add-on modifies the `blender_action`, `blender_tracks`, or `action_on_type` parameters, would it change those values in the export?

pepperoni505:
@julienduroure is it too late for this to ship in the 3.1 release?

julienduroure:
Hello,
@pepperoni505 No, not too late. I will check it and merge during next week if everything is OK

pepperoni505:
Alright, thanks

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animations.py
```diff
@@ -109,21 +109,26 @@ def __gather_animation(blender_action: bpy.types.Action,
...
-        print_console("WARNING", "Animation '{}' could not be exported. Cause: {}".format(name, error))
-        return None
-
-    if not animation.channels:
-        return None
+    animation = None
+
+    export_user_extensions('pre_gather_animation_hook', export_settings, animation, blender_action, blender_object)
+
+    if animation is None:
```
julienduroure, line 116: Not sure this last change is good.
If the object is TRS animated and you add a custom animation, TRS will not be exported

julienduroure: @pepperoni505 ?

pepperoni505: Hi @julienduroure. I was thinking of different ways to do this - essentially the issue is the addon I'm writing doesn't use the `channels` part of the animation, so this will always return `None` for those. I see what you mean about the TRS issue - I can undo that commit, but we still have the issue of checking if `channels` are present

pepperoni505: I've figured out how to fix this on my end and have undone that commit

julienduroure:
Looks good to me

PR #1556, opened 2022-02-06
Fix #1549 Add support of factor in clearcoat via Multiply math node

### LINKED ISSUE #1549: clearcoatRoughnessFactor  value not exported from multiply node if image is present
**Describe the bug**
image + math multiply node 
exports as 
texture + clearcoatRoughnessFactor =1  
If image is not there , it exports the value from the principled bsdf slider correctly 

**Expected behavior**
Similar to roughness and metalness the math multiply value should act as clearcoatRoughnessFactor value in case of clearcoat

**Screenshots**
err
gltf

**Version**
 - OS: Windows
 - Blender Version [3.2.0]

julienduroure:
Some basic implementation via Multiply Math node.
Is there any other node that need to be checked or other checks that need to be done?

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
@@ -258,6 +258,17 @@ def __has_image_node_from_socket(socket):
     return True
 
 def __gather_clearcoat_extension(blender_material, export_settings):
+
+    def __gather_factor(primary_socket, export_settings):
+        # Look for a Mix node that mixes with pure white in front of
+        # primary_socket. The mix factor gives the factor.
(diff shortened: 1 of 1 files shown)


julienduroure:
Test failed because proxy attribute is now removed from blender master (Blender 3.2)

scurest:
The comment is incorrect (it was copied from the occlusion strength code I guess). There is already a function to get a factor, see eg. [how the metallicFactor works](https://github.com/KhronosGroup/glTF-Blender-IO/blob/3ab459dfc3797fdd3a634d66a06dff803d0362ac/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials_pbr_metallic_roughness.py#L111-L112).

julienduroure:
Thanks for your feedback @scurest !

PR #1560, opened 2022-02-06
Armature all animations export

julienduroure:


julienduroure:
When blend file contains only 1 armature, it exports all animations, not only active+NLA.
It can lead to export non wanted actions if some objects are also animated in file.

donmccurdy:
/cc https://blender.stackexchange.com/questions/267556/how-to-export-only-nla-animations-to-glb

^It looks like some users may be looking for a way to opt out of this behavior.

j-conrad:
It appears that even animation in the form of orphan data (animation with a 'fake user' assigned) gets exported. I would not expect the default behavior of the exporter to clutter up the .glb with potentially dozens of animations that a user may not even know existed. I've read the history of how this came to be, but I'm curious about (and am questioning) the decision to make it ON by default.

julienduroure:
Hello,
NLA is probably the less know and used Editor in Blender. This was an quite often requested feature to export all armature animation without having to Stash them or push down them into NLA.
Even if the NLA workflow is written in documentation, I received lots of ticket or support request asking me why animations are not exported.

Do you have an example of workflow where the _Fake User_ is assigned automatically, not by user hand, and where the user can't be aware of existence of these animations?

donmccurdy:
I tend to agree with @julienduroure, it was a constant stream of feedback and confusion from users (here and in three.js forums) struggling to figure out why their animations didn't export, and requiring that users stash into NLA was a non-obvious and awkward workaround for Blender not having direct associations between actions and objects. 

Perhaps there's something we can do to "pre-screen" animation actions, before exporting them with the armature? For example if the action doesn't match up to the bone structure, we skip it unless stashed into NLA. I don't have an example at hand, but I think that would eliminate many of the junk animation exports I've seen.

julienduroure:
> Perhaps there's something we can do to "pre-screen" animation actions, before exporting them with the armature? For example if the action doesn't match up to the bone structure, we skip it unless stashed into NLA. 

Currently, the only check that is done is to verify that the animation is an armature action (so that there are some bone fcurve(s) ).
This prevents to export random object animation attached to armature nodes.

Issue #1559, opened 2022-02-06
problems with exporting one frame animations


weckr:
I have a character file with skeleton. Exporting "normal" animation like a walkcycle works fine, but I want to have some one frame long animations/poses, and those just do not work. I thought it might be godot engine problem but they do not work after reimporting the exported glb back into blender, the animation is just not there. 

I thought its a sampling issue and tried to get around it by duplicating the keyframes and making it several frame long animations, but that also doesnt work, i suspect because there is no delta detected. Only when i put keyframes in the middle that i move and rotate a bit, it exports mostly properly - but for example in "slide" animations, there are some differences in the export from the original file. 

I try to hard keyframe every bone at the beginning of animation. Im at my wits end with this one, spent hours. Alternatives (even paid fbx plugin), also dont really work with godot. Also this is rather complex rig, but I had this exact issue with extremely simple 10-or so bone rig with basic IK.

Also tried every possible export setting combination possible i think.

**To Reproduce**
1. Export glb file with static or one frame animations

**Expected behavior**
Correct keyframes will be sampled and exported

**.blend file/ .gltf**
[ronin.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/8010733/ronin.zip)

**Version**
 - OS: manjaro linux
 - Blender Version 3.0

julienduroure:
Related to #1541

weckr:
Thank you so much, got it working by deleting the if clauses in https://github.com/KhronosGroup/glTF-Blender-IO/pull/1433/commits/22df001409f94a19b37e1f9a70fbd68d61d65c59  and just always returning the keyframes. It might be good idea to put toggle in the export settings for this, in the game engine case, they almost always have compression settings for imported animations, and could save somebody some time. I could do pull request for this but it would be my first and dont know if you already have proper solution planned for next release.

julienduroure:
Hello,
Please be free to propose a PR, I didn't started to work on it yet.
Deadline is to merge it before end of February, to get it into 3.1

julienduroure:
Fixed by #1575

PR #1563, opened 2022-02-07
animation export settings toggle to force export all animation tracks without optimization

weckr:
https://github.com/KhronosGroup/glTF-Blender-IO/issues/1559
tested it with blender 3.0 where it works

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -388,6 +388,15 @@ class ExportGLTF2_Base:
         default=False
     )
 
+    export_all_tracks: BoolProperty(
+        name='Export all tracks',
+        description=(
+            "Export all animation tracks without optimization for filesize."
+            "Can help with one frame long and static animations"
+        ),
+        default=False
+    )
+
     export_current_frame: BoolProperty(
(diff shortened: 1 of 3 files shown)


julienduroure (changes requested):
About vocabulary: I think we can't name it "track", as this is the name of another concept in Blender (NLA)

weckr:
Yeah, I thought naming could be an issue, honestly no idea what to is appropriate here, 
Export without filesize optimization
Force export all keyframes
?

julienduroure:
@weckr Do you plan to work on this PR?

* You have to validate the CLA
* Update the PR

I'd like to have it merge during the following week. If we don't plan to, don't worry, I will manage it

weckr:
I can rename it to optimize animation size then. My email in git wasnt setup properly tho because i was just using it for personal stuff, so im not sure i can validate the license easily... should i open a PR from a new branch or something like that... Cant I just write here that I agree with the license ...

julienduroure:
@weckr I can't accept and validate work without the CLA validation
Closing this ticket.
I will manage the change on my side, except if another PR is proposed (and CLA validated) until then

PR #1567, opened 2022-02-12
Fix #1540 Use active UVMap when needed

### LINKED ISSUE #1540: UV (or texture transform) mismatch
[glb-textures.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/7884699/glb-textures.zip)

In blender-3.1.0-alpha+master.34d424fd64bb (2022-01-17), this looks good, but in the GLB export (displayed in https://gltf-viewer.donmccurdy.com/ or no sketchfab), the textures are way too big.

good:
image

wrong:
image

floor texture (from blenderkit):
image

julienduroure:
If no explicit UVMap is used by the texture, store information, and then override material to use active UVMap of mesh


scurest:
This doesn't seem to work when objects _are_ supposed to share a texture. For example, import [TwoCubesSharedMaterial.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/8054207/TwoCubesSharedMaterial.zip), then export. I get two materials, two textures, two images, two samplers, two buffer views.

julienduroure:
@scurest . We have to ignore when active UVMap index is 0.
Fixed

julienduroure:
Well, problem will be the same with 2 objects with active UVMap at 1 for example.
This need to be cached somewhere, as you suggested.

julienduroure:
This seems to work.

- Create a new material if node tree is using active UVMap
- Use existing one if node tree is not using active UVMap, even if meshes have different active UVMaps

julienduroure:
@scurest any feedback?

scurest:
This



produces textures/images/bufferviews that are not shared, even on the emission texture.

Out

scurest:
Also the emissiveTexture is using texCoord 1, it should be using 0.

julienduroure:
@scurest Ok for the not shared, deepcopy is a little too much. Should copy all data except binary.

About emissiveTexture using texCoord at 1, what is your mesh setup? I can't get it in my test.

scurest:
>About emissiveTexture using texCoord at 1, what is your mesh setup? I can't get it in my test.

That was my bad, I was testing what happens if you use a UVMap that doesn't exist.

julienduroure:
Should be OK now

julienduroure:
@scurest Can you please test it? Should be ok now

julienduroure:
Any other feedback before I merged it?

PR #1569, opened 2022-02-14
Fix #1301 - emission node needs pbr not default

### LINKED ISSUE #1301: Export: emission materials export too bright
A basic emission material like this

emis

exports as

```
  "materials": [
    {
      "emissiveFactor": [1, 0, 0],
      "pbrMetallicRoughness": {},
      ...
    }
```

Since pbrMetallicRoughness is omitted, the defaults are baseColor=[1,1,1,1], metallic=1, roughness=1. This is much too bright.

out

Messing around in Blender to find an equivalent-looking Principled node, baseColor=[0,0,0,1], metallic=(anything), roughness=1 is much better.

> _Note:_ This material setup was discussed at https://github.com/KhronosGroup/glTF-Blender-IO/pull/1209#issuecomment-693649108.

julienduroure:
Fix #1301 
When emission node is used, pbr needs non default values

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
@@ -60,10 +60,10 @@ def gather_material(blender_material, export_settings):
         pbr_metallic_roughness=__gather_pbr_metallic_roughness(blender_material, orm_texture, export_settings)
     )
 
-    # If emissive is set, but no orm_texture,
-    # "Emission" node is used. We need to set manually default values for 
+    # If emissive is set, from an emissive node (not PBR)
+    # We need to set manually default values for 
     # pbr_metallic_roughness.baseColor
-    if material.emissive_factor is not None and orm_texture is None:
+    if material.emissive_factor is not None and gltf2_blender_get.get_emissive_node_socket(blender_material) is not None:
         material.pbr_metallic_roughness = gltf2_blender_gather_materials_pbr_metallic_roughness.get_default_pbr_for_emissive_node()
 
     export_user_extensions('gather_material_hook', export_settings, material, blender_material)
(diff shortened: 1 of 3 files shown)


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
```diff
@@ -60,6 +60,18 @@ def gather_material(blender_material, export_settings):
         pbr_metallic_roughness=__gather_pbr_metallic_roughness(blender_material, orm_texture, export_settings)
     )
 
+    print(material.pbr_metallic_roughness)
+    print("->", orm_texture)
```
emackey, line 64: Did you intend to leave these `print` statements in here?  There's another one on line 70.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
```diff
@@ -60,6 +60,18 @@ def gather_material(blender_material, export_settings):
...
+
+    # If emissive is set, but no orm_texture,
+    # "Emission" node is used. We need to set manually default values for 
+    # pbr_metallic_roughness.baseColor
+    if material.emissive_factor is not None and orm_texture is None:
```
emackey, line 69: Is testing for orm_texture the correct answer here?  What if they have a Principled BSDF node without textures?

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
julienduroure: Fixed

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
julienduroure: Exact. I just pushed another (better) check

emackey (approved):
Looks good to me.

PR #1574, opened 2022-02-17
Fix #1465 - add UI for glTF settings node creation

### LINKED ISSUE #1465: Make glTF Settings (for AO) available by default
Currently, people that want to export AO need to manually create the glTF Settings node.
This can be done automatically.

Some notes:
- Code for creation is already done, as the node is created at import when needed (when AO in glb file imported)
- Not sure every Blender user when the node to be displayed in UI, so maybe we need to add this option behind an addon parameter?

julienduroure:
Option need to be actived in addon preferences, avoiding adding UI for people that don't want to use glTF export

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -1168,6 +1168,24 @@ class ImportGLTF2(Operator, ImportHelper):
             self.loglevel = logging.NOTSET
 
 
+def update_gltf_settings_node_ui(self, context):
+    import io_scene_gltf2.blender.com.gltf2_blender_ui as blender_ui
+    if bpy.context.preferences.addons['io_scene_gltf2'].preferences.settings_node_ui is True:
+        bpy.utils.register_class(blender_ui.NODE_PT_GLTF_PANEL)
+    else:
+        blender_ui.unregister()
+
+class GLTF_AddonPreferences(bpy.types.AddonPreferences):
+    bl_idname = __package__
+
+    settings_node_ui : bpy.props.BoolProperty(default= False, update=update_gltf_settings_node_ui)
+
+
+    def draw(self, context):
+        layout = self.layout
+        row = layout.row()
+        row.prop(self, "settings_node_ui", text="Settings node interface")
+
 def menu_func_import(self, context):
     self.layout.operator(ImportGLTF2.bl_idname, text='glTF 2.0 (.glb/.gltf)')
 
@@ -1185,14 +1203,18 @@ classes = (
     GLTF_PT_export_animation_skinning,
     GLTF_PT_export_user_extensions,
     ImportGLTF2,
-    GLTF_PT_import_user_extensions
+    GLTF_PT_import_user_extensions,
+    GLTF_AddonPreferences
 )
 
 
 def register():
+    import io_scene_gltf2.blender.com.gltf2_blender_ui as blender_ui
     for c in classes:
         bpy.utils.register_class(c)
     # bpy.utils.register_module(__name__)
+
+    blender_ui.register()
 
     # add to the export / import menu
     bpy.types.TOPBAR_MT_file_export.append(menu_func_export)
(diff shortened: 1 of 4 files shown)


julienduroure:
@emackey Can you please have a look?
You can activate new UI in addon preferences, and then add the glTF Settings node with this new panel.

addons/io_scene_gltf2/__init__.py
```diff
@@ -1168,6 +1168,24 @@ def set_debug_log(self):
...
+class GLTF_AddonPreferences(bpy.types.AddonPreferences):
+    bl_idname = __package__
+
+    settings_node_ui : bpy.props.BoolProperty(default= False, update=update_gltf_settings_node_ui)
+
+
+    def draw(self, context):
+        layout = self.layout
+        row = layout.row()
+        row.prop(self, "settings_node_ui", text="Settings node interface")
```
donmccurdy, line 1187: Suggested naming, and tooltip if possible:

- text: "Shader Editor Panel"
- tooltip: "Displays glTF 2.0 panel in the Sidebar region of the Shader Editor (shortcut for glTF Settings node)."

Based on capitalization elsewhere in preferences, and terminology used in the Shader Editor documentation: https://docs.blender.org/manual/en/latest/editors/shader_editor.html#sidebar

addons/io_scene_gltf2/blender/com/gltf2_blender_ui.py
```diff
@@ -0,0 +1,66 @@
...
+    bl_category = "glTF 2.0"
+
+    def draw(self, context):
+        layout = self.layout
+        row = layout.row()
+        row.operator("node.gltf_settings_node_operator")
+
+class NODE_OT_GLTF_AO(bpy.types.Operator):
+    bl_idname = "node.gltf_settings_node_operator"
+    bl_label  = "Add Settings glTF node"
```
donmccurdy, line 39: Suggested naming:

- "Add glTF Settings"

donmccurdy (commented):
Is it possible to cause the "glTF Settings" node to appear under the Group dropdown when this setting is enabled? As far as I can tell it only appears once a glTF Settings group has actually been used.

Screen Shot 2022-03-08 at 9 51 11 AM

julienduroure:
About adding in menu: 
Adding node groups in "Groups" menu is done only when a group is already instanced. I didn't find a way to perform it without instancing it. But ... 
I just found that we can add a completely new entry in menu, like "Templates" in this screenshot.
I suggest to add a "glTF 2.0" entry.
In that case, panel is no more required.

image

addons/io_scene_gltf2/__init__.py
julienduroure: Done, but need to be rechecked, as this is now in add menu

addons/io_scene_gltf2/blender/com/gltf2_blender_ui.py
julienduroure: Done

julienduroure:
Ok, I found how to put it in Add > Output Menu

addons/io_scene_gltf2/__init__.py
```diff
@@ -1168,6 +1168,20 @@ def set_debug_log(self):
             self.loglevel = logging.NOTSET
 
 
+class GLTF_AddonPreferences(bpy.types.AddonPreferences):
+    bl_idname = __package__
+
+    settings_node_ui : bpy.props.BoolProperty(
+            default= False, 
+            description="Displays glTF 2.0 node in Shader Editor (Menu Add > Ouput)"
```
donmccurdy, line 1176:
```suggestion
            description="Displays glTF Settings node in Shader Editor (Menu Add > Ouput)"
```

addons/io_scene_gltf2/__init__.py
```diff
@@ -1168,6 +1168,20 @@ def set_debug_log(self):
...
+            )
+
+
+    def draw(self, context):
+        layout = self.layout
+        row = layout.row()
+        row.prop(self, "settings_node_ui", text="Shader Editor")
```
donmccurdy, line 1183:
```suggestion
        row.prop(self, "settings_node_ui", text="Shader Editor Add-ons")
```

donmccurdy (approved):
Looks excellent — thanks @julienduroure! Two small wording suggestions.

emackey (approved):
Awesome!

PR #1582, opened 2022-03-01
Fix #822 Export Armatures without skined mesh

### LINKED ISSUE #822: Export Armature and Animations Only - Exclude Mesh
I would like to be able to export only the animations. THe idea is that I export the mesh+armature in one file, then export animations+armature in as other files. I do not want them together in one large file for specific reasons.

julienduroure:
Fix #822 Export Armatures without skined mesh


fire:
I tested by selecting the skeleton and exporting as glb with `selected` only.

Works.

Tested on Blender 3.2 dailies from steam and your pr branch.

[Fox_anims.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/8178031/Fox_anims.zip)

fire:
Selecting more than one armature works too.

fire:
The fox animation is broken in the zip but when exported again it should look like:

image


[fox.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/8190797/fox.zip)

julienduroure:
@fire I don't get what you mean in your last comment

fire:
I meant that in my sample zip, I uploaded the wrong file. Your pr works as intended.

PR #1589, opened 2022-03-09
feat: add import animation hooks

pepperoni505:
This adds two more hooks to the import animation cycle:
```
gather_import_animation_before_hook(self, gltf, anim_idx, import_settings)
gather_import_animation_after_hook(self, gltf, anim_idx, track_name, import_settings)
```

If these changes are ok, is it possible for it to be included in 3.1.X releases?


julienduroure:
Hello,
Too late for 3.1
Code freeze was one week ago, and stable release should happen today

pepperoni505:
Should the hooks be moved to `/blender/imp/gltf2_blender_animation.py` instead?

julienduroure:
In blender_animation, it will make hooks for generic, as it will manage woth TRS animations and shapekeys animations

pepperoni505:
So I should move it there?

julienduroure:
yes please. More generic case will avoid to create a new one later for shapekeys animation

pepperoni505:
Ok, it's moved. Can you check to make sure this is good?

pepperoni505:
Merging into #1591

PR #1591, opened 2022-03-12
feat: add more import hooks

pepperoni505:
Adds a few new importer hooks:

- `gather_import_decode_primitive(self, gltf_mesh, gltf_primitive, import_settings)`

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
@@ -146,6 +146,8 @@ def do_primitives(gltf, mesh_idx, skin_idx, mesh, ob):
             print_console('INFO', 'Draco Decoder: Decode primitive {}'.format(pymesh.name or '[unnamed]'))
             decode_primitive(gltf, prim)
 
+        import_user_extensions('gather_import_decode_primitive', gltf, pymesh, prim)
+
         if prim.indices is not None:
             indices = BinaryData.decode_accessor(gltf, prim.indices)
             indices = indices.reshape(len(indices))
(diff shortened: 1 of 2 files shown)


addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
```diff
@@ -69,6 +69,17 @@ def do_primitives(gltf, mesh_idx, skin_idx, mesh, ob):
...
+    class IMPORT_mesh_options:
+        def __init__(self, vertex_colors: bool = True, skinning: bool = True, skin_into_bind_pose: bool = True, use_auto_smooth: bool = True):
+            self.vertex_colors = vertex_colors
+            self.skinning = skinning
+            self.skin_into_bind_pose = skin_into_bind_pose
+            self.use_auto_smooth = use_auto_smooth
+
+    mesh_options = IMPORT_mesh_options()
+    import_user_extensions('gather_import_mesh_options', mesh_options, gltf, pymesh, skin_idx)
+
```
pepperoni505, line 82: Would like some feedback on this approach

julienduroure: There is already some option to export (or not export) vertex colors, skinning and normals. Maybe a new hook is not needed?
Not sure what are your use cases exactly.

pepperoni505: This is for importing.

julienduroure: Oops, sorry :)
For normals, isn't 'shading' option what your are looking for?

pepperoni505: Not sure what you mean. Did I accidentally change something with normals?

julienduroure: No, 
I mean that there is already a dropdown list option at import to choose how normals are taken into account. I don't know if you already see that, and why the choices are not fitting your need

pepperoni505: In short, the glTF files I'm trying to import already have the verts position changed based off the skin. As this importer also applies it at import time, it causes errors in the model

julienduroure:
@pepperoni505 Can you please rebase with 3.2/master?
After an API change, there is no more vertex_colors, but API is now based on color attributes

julienduroure:
@pepperoni505 Can you please rebase with 3.2/master?
After an API change, there is no more vertex_colors, but API is now based on color attributes

pepperoni505:
> @pepperoni505 Can you please rebase with 3.2/master? After an API change, there is no more vertex_colors, but API is now based on color attributes

Rebased

PR #1592, opened 2022-03-16
feat: add setting to restore first animation at import

pepperoni505:
This adds an option to the import settings to enable or disable restoring the first animation in the glTF file. In my use case, I do not want the first animation to be restored. I don't quite know the purpose of this, so if it is needed, feel free to close this PR.

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -1062,6 +1062,12 @@ class ImportGLTF2(Operator, ImportHelper):
         default=True,
     )
 
+    restore_first_animation: BoolProperty(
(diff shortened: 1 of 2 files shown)


julienduroure:
I'm not sure the solution is to add more and more and more option in UI.
We need to keep it as simple as possible.

Possible solution:
- add an "advanced options" checkbox in addon preferences, where more option are available
- Add more hooks, to be able to check this kind of feature
- Other?

pepperoni505:
I think a hook might be a good idea. Not sure how you would implement that however. Maybe add a class with the option and pass to hook?

pepperoni505:
Closing in favor of #1591

PR #1596, opened 2022-03-21
fix: improve objectcache logic

pepperoni505:
This PR changes how the `objectcache` logic works. I was running into very slow performance on a scene that had ~1500 nodes each with unique animations. The issue was every time `get_object_matrix` was called, it would get the matrix for every object in the VTree, even though it is completely unnecesary. In the `objectcache`, there was also a case where it would again get the matrix for every object just to only use the info from one entry. 

Previous export time with scene that had 1500 nodes w/ animations: 72 minutes
New export time: 5 minutes

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
@@ -143,7 +143,8 @@ def get_object_matrix(blender_obj_uuid: str,
                       bake_range_end: int,
                       current_frame: int,
                       step: int,
-                      export_settings
+                      export_settings,
+                      only_gather_provided=False
                     ):
 
     data = {}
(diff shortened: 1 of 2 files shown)


julienduroure:
Hello,
Can you please share your example file?

Caching object speed up mostly all cases I tested, but there are probably some corner cases.
Even if in your case this PR make better performance, it can lead to worse cases with some other files.

pepperoni505:
Am I able to privately send you the file?

julienduroure:
@pepperoni505 Sure, email me the url where I can download it : contact AT firstnamename DOT com

I think I get what case you want to avoid. I let a comment on your PR

pepperoni505:
Sent you the file.

julienduroure:
> Sent you the file.

Received. Will have a look soon

julienduroure:
Tests on my computer with your test file:
- master : 523 seconds (caching all objects / all frame)
- your original PR : 291 seconds (caching one object at a time, but caching frames) : 
- current PR (caching all objects / all frame for first run, then only object where animation changed): 290 seconds

Not sure how we can have a such so big difference, where I have better performance, but only 50%, where you go from 72 minutes to 5 minutes

Note : at long term, after animation refactoring (not planned yet, not for 2022), we may change multiple action on multiple objects. So we will be able to take this case into account.

Anyway, This PR make it faster, but there are still some stuff to understand (way quite no difference between solution 2 and solution 3 ?)

PR #1599, opened 2022-03-26
Manage not initial data buffer

julienduroure:
No change in Blender, but this file can be used to change glb data
directly in python without Blender


julienduroure:
Seems today Blender build in KO.
I open a bug report here : https://developer.blender.org/T96796

Running local tests (with Blender 3.1) are ok

julienduroure:
Note: Blender bug is now solved

PR #1601, opened 2022-03-28
🐛 resolve uvmap index using the correct mesh

m0ppers:
previously it was just trying to find the uvmap name globally via its name in ANY mesh.

This could result in broken glTF files. For example if there was a mesh having 2 UVMaps named "First UVMap" and "UVMap" and there is a second mesh with just one UVMap "UVMap" then the resulting glTF file would export the textures of the second mesh with a texCoord 1 (which is not present and not exported in this mesh) because the logic was just searching globally.

I am not sure if this is the correct way to fix this but this works for me. I am especially unsure about the id_data thing and if that is really a stable way to do it.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_texture_info.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_texture_info.py
@@ -165,12 +165,15 @@ def __gather_texture_transform_and_tex_coord(primary_socket, export_settings):
     use_active_uvmap = True
     if node and node.type == 'UVMAP' and node.uv_map:
         # Try to gather map index.
-        for blender_mesh in bpy.data.meshes:
-            i = blender_mesh.uv_layers.find(node.uv_map)
-            if i >= 0:
-                texcoord_idx = i
-                use_active_uvmap = False
-                break
+        node_tree = node.id_data
+        for mesh in bpy.data.meshes:
+            for material in mesh.materials:
+                if material.node_tree == node_tree:
+                    i = mesh.uv_layers.find(node.uv_map)
+                    if i >= 0:
+                        texcoord_idx = i
+                        use_active_uvmap = False
+                        break
 
     return texture_transform, texcoord_idx or None, use_active_uvmap
 


julienduroure:
Hello,
Can you please share a simple .blend file ? It can help people to test / check your PR

m0ppers:
[uvmapfail.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/8361751/uvmapfail.zip)

I am not 100% sure that the order of data elements is consistent so I embedded the exported gltf as well:

```
[mop@ludwig-otto uvmapfail]$ cat uvmapfail.gltf | jq .meshes
[
  {
    "name": "Cube",
    "primitives": [
      {
        "attributes": {
          "POSITION": 0,
          "NORMAL": 1,
          "TEXCOORD_0": 2,
          "TEXCOORD_1": 3
        },
        "indices": 4,
        "material": 0
      }
    ]
  },
  {
    "name": "Cube.001",
    "primitives": [
      {
        "attributes": {
          "POSITION": 5,
          "NORMAL": 6,
          "TEXCOORD_0": 7
        },
        "indices": 4,
        "material": 1
      }
    ]
  }
]
[mop@ludwig-otto uvmapfail]$ cat uvmapfail.gltf | jq .materials
[
  {
    "doubleSided": true,
    "name": "Material.001",
    "pbrMetallicRoughness": {
      "baseColorTexture": {
        "index": 0
      },
      "metallicFactor": 0,
      "roughnessFactor": 0.4000000059604645
    }
  },
  {
    "doubleSided": true,
    "name": "Material",
    "pbrMetallicRoughness": {
      "baseColorTexture": {
        "index": 1,
        "texCoord": 1
      },
      "metallicFactor": 0,
      "roughnessFactor": 0.4000000059604645
    }
  }
]
```

Cube has two uvmaps. The one at index 1 is named like the one at index 0 of Cube.001. Cube.001 only has one UVMap.

After exporting Cube.001 references texCoord 1 which does not exist for this mesh.

The UVMap logic searched globally by name across all meshes and returned index 1 because Cube had a uvmap with the same name at index 1

PR #1602, opened 2022-03-30
Export: Allow custom name for merged animation

Macleodsolutions:
Currently merged NLA strip animations are exported with the hardcoded name of "Animation". This would allow the user to select a custom animation name instead.


julienduroure:
Thanks!

PR #1606, opened 2022-03-30
Added SpecularGlossiness export

elmajime:
This PR handles importing a material using the KRH_materials_pbrSpecularGlossiness extension.

The node tree generated follows the setup needed for exporting a specular/glossiness material to gltf.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials.py
@@ -22,6 +22,7 @@ from io_scene_gltf2.blender.exp import gltf2_blender_gather_texture_info, gltf2_
 from io_scene_gltf2.blender.exp import gltf2_blender_search_node_tree
 
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_materials_pbr_metallic_roughness
+from io_scene_gltf2.blender.exp import gltf2_blender_gather_materials_pbrSpecularGlossiness
 from io_scene_gltf2.blender.exp import gltf2_blender_gather_materials_unlit
 from ..com.gltf2_blender_extras import generate_extras
 from io_scene_gltf2.blender.exp import gltf2_blender_get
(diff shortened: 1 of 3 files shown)


CrazyHairGuy:
The addon seems to import the specular material just fine, but upon export the material generated just has the occlusion and emission channels. 
Sample model: image
After import and immediate export: image

fire:
Guessing, I think SpecularGlossiness is replaced by https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_specular/README.md spec.

julienduroure:
As said by fire, KHR_materials_pbrSpecularGlossiness is set as archived. (see [here](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Archived/KHR_materials_pbrSpecularGlossiness/README.md ) )
So, creating new files with this extension is not recommanded (See [here](https://github.com/KhronosGroup/glTF/tree/main/extensions#archived-extensions-for-gltf-20) )

This is Superseded by [KHR_materials_specular](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_specular/README.md), that is currently under implementation in #1646

I so suggest to close this PR

donmccurdy:
Especially with the upcoming development of [Principled BSDF v2](https://developer.blender.org/T99447), I'd be inclined to agree we should probably not ship support for `KHR_materials_pbrSpecularGlossiness` here. This feature could (I expect?) be implemented through this add-ons extension API as well.

I would also just flag that `KHR_materials_pbrSpecularGlossiness` is mutually exclusive with other material-related glTF extensions. Features like clearcoat, iridescence, IOR, volume, transmission, etc. are supported on the metal/rough workflow only. With the addition of KHR_materials_specular, there is no longer need for `KHR_materials_pbrSpecularGlossiness`.

julienduroure:
So ... We agree that this will not be included in official exporter.
Closing.
Thanks!

PR #1609, opened 2022-04-02
Fix #1304 Normalisation of weights when all influences

### LINKED ISSUE #1304: Export: skin weights not normalized when using "Include All Bone Influences"
The WEIGHTS_n arrays are normalized [only if "Include All Bone Influences" is off](https://github.com/KhronosGroup/glTF-Blender-IO/blob/04e26bef903543d08947c5a9a5fea4e787b68f17/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitive_attributes.py#L197-L203). If "Include All Bone Influences" is on, and a mesh has non-normalized weights in Blender, the exported .gltf will have non-normalized weights and be invalid.

julienduroure:
Fix #1304

First draft, need to be tested

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitive_attributes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitive_attributes.py
@@ -179,6 +179,10 @@ def __gather_skins(blender_primitive, export_settings):
         max_bone_set_index += 1
     max_bone_set_index -= 1
 
+    # If no skinning
+    if max_bone_set_index < 0:
+        return attributes
+
     if max_bone_set_index > 0 and not export_settings['gltf_all_vertex_influences']:
         gltf2_io_debug.print_console("WARNING", "There are more than 4 joint vertex influences."
                                                 "The 4 with highest weight will be used (and normalized).")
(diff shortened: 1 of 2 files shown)


julienduroure:
@scurest External eyes are welcome. Can you please have a look?

scurest:
You can do it in a batch with numpy

```py
    # Convert weights to numpy arrays
    weight_arrs = []
    for s in range(0, max_bone_set_index+1):
        weight_id = 'WEIGHTS_' + str(s)
        weight = blender_primitive["attributes"][weight_id]
        weight = np.array(weight, dtype=np.float32)
        weight = weight.reshape(len(weight) // 4, 4)
        weight_arrs.append(weight)

    # Sum weights for each vertex
    for s in range(0, max_bone_set_index+1):
        sums = weight_arrs[s].sum(axis=1)
        if s == 0:
            weight_total = sums
        else:
            weight_total += sums

    # Normalize weights so they sum to 1
    weight_total = weight_total.reshape(len(weight_total), 1)
    for s in range(0, max_bone_set_index+1):
        weight_arrs[s] /= weight_total

    # weight_arr[s] is now ready to go into array_to_accessor
```

ninjz:
@julienduroure I made a comment here: https://github.com/KhronosGroup/glTF-Blender-IO/issues/1304#issuecomment-1093029964 that I think still exists in this PR. Would like to know what you think of this

PR #1612, opened 2022-04-02
feat: make extension hooks returnable

pepperoni505:
This will make it easier to add conditions to an extension.

For example, let's say a hook is added to skip materials. This would give extensions the ability to return whether or not to do so


pepperoni505:
Closing as after further consideration this doesn't seem practical. If multiple extensions are returning values, which do you choose from?

PR #1619, opened 2022-04-09
Fix #1618 No texture export when original can't be retrieved

### LINKED ISSUE #1618: "Keep original" + packed images --> Invalid file
Using Keep original, when original source file can't be retrieved (for example because there are packed images), this generate invalid file (image without an uri)

Better to not export texture at all, will be better than export an invalid glTF file

julienduroure:
Fix #1618 No texture export when original can't be retrieved


julienduroure:

Issue #1349, opened 2021-03-09
Export then import glb, not same data (skinning)


Snouf-ID:
Hello,

When I export this model to glb and re-import it, if I delete armature, I observe that the mesh is no longer in its standard correct position.
If I export / import in fbx for example it works as expected.

Green : mesh when I delete armature
Red : mesh when I delete armature after export/import in glb

![bug_a_1](https://user-images.githubusercontent.com/47319640/110486088-97a8ee80-80ec-11eb-9c50-72b5fcf0fa05.png)

https://drive.google.com/file/d/1R31zmWWK6gdUfLsTCSPOzDqcbUV_vd3I/view?usp=sharing

julienduroure:
This is now solved by using rest pose as default values for joint nodes

PR #1621, opened 2022-04-09
Use rest pose for armature when 'use current frame' is off

julienduroure:
Use rest pose ( = edit mode of bones) as default TRS for joint nodes.
If you want to have the current pose as default TRS for joint nodes, use 'use current frame' option

No change for objects (not joints), this will use frame 0 (if option is off) or current frame (if option is on)


julienduroure:

scurest:
Does this fix #1349?

julienduroure:
@scurest Yes, it does!

PR #1624, opened 2022-04-11
Check bad image

julienduroure:
- using image.channels is time consuming
- using image.channels load buffers that was never freed
This is very noticeable with very big texture images

I try to find another way to detect bad images without having to check if image has channels.
Currently we can detect if image has no size. Not sure if we have case with image with size, but without channels? (weird)

See https://github.com/godotengine/godot/issues/60063


julienduroure:

fire:
No objections.

PR #1627, opened 2022-04-14
Fix #1626 Set skinned mesh at root

### LINKED ISSUE #1626: Skin/armature export produces NODE_SKINNED_MESH_NON_ROOT
There's been discussion of this before, for example https://github.com/KhronosGroup/glTF-Blender-IO/issues/1538#issuecomment-1015403354

> So perhaps GLTF exporter should apply transforms and clear skinned mesh parent during export.

I've come to understand the correct fix is to **NOT** apply transforms to the skinned mesh.  The design of glTF's skinning system is such that any transforms applied to the skinned mesh itself are completely ignored, only joint transforms are considered.  Specifically, the final location of each vertex coming out of a glTF skinned vertex shader has influence from joints adding up to 100% strength, with exactly 0% coming from the mesh's own model matrix.

So the correct answer here sounds pretty reckless, but it's true:  Just clear the parent of the skinned mesh.  As in, just lose the reference to the parent.  Don't apply any of the parent's transforms or animations (and you may also clear any transforms or animations that target the skinned mesh itself; they're unused as well).

Essentially glTF's skinning system treats the mesh as always being at the origin, and uses the joints (not parent node transforms) to move each vertex where it belongs.  The joints' transforms do the work of inheriting from their own parents, so these are not lost.

julienduroure:
Fix #1626 
Set skinned mesh at root

Note : Need to be fully tested


scurest:
Even if the transform is not needed for the skinned mesh, there may be child objects that need the transform. The importer handles this by possibly creating a new node to attach the skinned mesh to.

Example

1. Import CesiumMan.
2. Parent a cube to the CesiumMan mesh.
3. Drag the armature so its off-center.
4. Export and view in viewer. On this branch, the cube is at the origin instead of off-center with the CesiumMan mesh.

donmccurdy:
Copying from https://github.com/KhronosGroup/glTF-Blender-IO/issues/1626#issuecomment-1099236644 —

> *Understanding that applying transforms to the skinned mesh node (as opposed to the armature / joints) does not have an effect in glTF ... don't we still need to apply those transforms somewhere, if they were contributing to the final position of the vertices in Blender? Applying the transform to the inverse bind matrices might be the best approach if that's the case.*

julienduroure:
Trying a fix to the case described by @scurest 
by resetting not only the TRS of glTF mesh, but also the internal world matrix that is used for children calculation

julienduroure:
So, any other cases that breaks or are not solved?
Don't hesitate to tell me

julienduroure:
Note: This breaks driver detection for now

PR #1669, opened 2022-06-17
Fix 1668 - export def bone only without anim

### LINKED ISSUE #1668: Export deforming bones only without exporting animations
I need to export my meshes and the rig separately. When exporting the meshes, I'd like to not export any animations. However, I also need to export the meshes with deforming bones only. Currently, the option to export deforming bones only is locked behind exporting animations.
image

Have the option to export non-deforming bones be parallel to the option of exporting animations. However, it seems like the option to export non-deforming bones requires that animations are sampled - why is that? Knowing which bones are deforming or not is independent of any animations, so this option should not be dependent on animations.

Ways of not exporting animations despite having the animation box ticked, such as setting sampling rate to 0 (automatically moves to 1). Deleting all animations prior to export and undoing after exporting seems like an option, but should not be the solution here.

**Additional context**
Using Blender 3.2

julienduroure:
Fix #1668 - Export def bone only without exporting animation
This move option under skinning instead of animation

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -385,7 +385,7 @@ class ExportGLTF2_Base:
 
     export_def_bones: BoolProperty(
         name='Export Deformation Bones Only',
-        description='Export Deformation bones only (and needed bones for hierarchy)',
+        description='Export Deformation bones only',
         default=False
     )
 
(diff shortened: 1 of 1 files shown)


PerplexedPeach:
Do you know if this will make it to a Blender patch version (like 3.2.1) or is it only for the next minor version (3.3)? Not sure what the policy is (whether patches are only for fixes while this might not qualify as a fix)

julienduroure:
@PerplexedPeach 
Hello,
Corrective release (3.2.x) is only for crashes and/or regressions.

This PR will be available on 3.3

PR #1676, opened 2022-06-28
Attributes refactoring

### LINKED ISSUE #1658: Better Vertex Color / Attributes management
Cf discussion starting at this comment, and next ones : https://github.com/KhronosGroup/glTF-Blender-IO/pull/807#issuecomment-1136480131

julienduroure:
Fix #1658


julienduroure:
Hello,
Any feedback / tests are welcome!

scurest:
Why func_get/func_get_args? Just pull the data into a numpy array.

Why enable_for_edge/enable_for_point/skip_getting_to_dots? Use the domain to decide what goes in tri/edge/point prims.

Why after_others? Either make morph tangents another generic attribute by computing them in advance, or compute them after the fact as a separate step.

You don't need all the glTF data_type/component_type stuff, you can tell that from the dtype/shape of the numpy array. (You do need to know if it's normalized though.)

Turn extract_primitive into a class so the parts can be factored up into clean member functions.

primitives/primitives_attributes should be refactored to handle generic attributes too, no special cases for position, color, etc.

I assume you know FACE attributes are broken. It's easiest to convert them to CORNER domain (can use loop_triangles to get a numpy array with the polygon index at each loop).

julienduroure:
Thanks!

> Why func_get/func_get_args? Just pull the data into a numpy array.

Here are the main steps of the new workflow:
- 1. Declare attributes that are going to be exported
- 2. Create dots array that are going to store data
- 3. populate dots array
- 4. for each primitive:
  - 4.a extract needed data from dots array to prim_dots array
  - 4.b create attribute data

I use func_get at declaration (1.) , and call this function at population (3.), because at 1., dots is not yet created, and I wanted to avoid to create a tmp array, for perf and memory saving.
Any better idea?

>Why enable_for_edge/enable_for_point/skip_getting_to_dots? Use the domain to decide what goes in tri/edge/point prims.

Ok for custum attribute to use domain, but what about legacy attributes (position, normals, color, etc...) ?

> Why after_others? Either make morph tangents another generic attribute by computing them in advance, or compute them after the fact as a separate step.

I wanted to make it generic, means that I wanted to avoid add a separate step with hard coded morph tangents, because morph tangents needs data already calculated by other attributes. Maybe a solution will be to make calculate data by the first attribute that needs them, and keep them for next ones.

> You don't need all the glTF data_type/component_type stuff, you can tell that from the dtype/shape of the numpy array. (You do need to know if it's normalized though.)

Yes, maybe only normalized info can be stored, will have a look

> Turn extract_primitive into a class so the parts can be factored up into clean member functions.

That a next step, reflecting the steps I mentioned above

> primitives/primitives_attributes should be refactored to handle generic attributes too, no special cases for position, color, etc.

That a next step

> I assume you know FACE attributes are broken. It's easiest to convert them to CORNER domain (can use loop_triangles to get a numpy array with the polygon index at each loop).

Yes, still a TODO. Same work will have to be done on FaceMaps.

> No newline at end of file, inconsistent spacing between functions, x in dict.keys() can be shortened x in dict.

Done

Any idea how to get EDGE data and correctly dispatch on dots?


BTW, not sure I will have time to finish/cleanup this PR in time for 3.3
Maybe I will proposed another PR for 3.3, exporting active color attribute as COLOR_0

scurest:
> dots is not yet created, and I wanted to avoid to create a tmp array, for perf and memory saving.

Well, okay. Why don't you store a closure you can call to get the array instead of function+args. The big issue is the dot code is now coupled to the "get data" code (eg. __get_uvs_attribute). Dots are an internal implementation detail of the "gather triangle primitive" part.

> Ok for custum attribute to use domain, but what about legacy attributes (position, normals, color, etc...) ?

They have a domain too. Position is POINT domain, normal and color are CORNER, etc. etc.

> Any idea how to get EDGE data and correctly dispatch on dots?

It doesn't seem like EDGE data should be exported for triangle primitives, similar to how CORNER/FACE data isn't exported for lines/points. It's possible to export it for line primitives if you really want to (it would involve creating "tips" for edges similar to "dots" for triangles), but I'd just skip them for now.

scurest:
Here's some bugs. Exporting with the default cube and "Attributes" on.

* Create all four combinations of domain/type color attributes -> export -> error.
* Vertex domain INT8 attribute -> export -> error.
* Vertex domain BOOLEAN attribute -> export -> no error, but no attribute.
* Vertex domain INT attribute -> export -> error.

Issue #960, opened 2020-03-05
[Doubt] How is tangents generated for cubicspline animations


jjcasmar:
I am implement EXT_property_animation as part of a custom extension and I am not sure I understand the logic for converting the bezier control points to glTF cubicsplines. Could someone explain it to me, please?

something that annoys me is that in blender you have 6 values to control each bezier keyframe (time and value for the sample and the 2 control points) while in glTF we only have 4 values (timestamp, value, in-tangent and out-tangent). That makes the space of possible curves smaller. Is this correct? If so, how is the conversion done?

donmccurdy:
There can be quite a lot of things in authoring tools, like Blender, that don't translate easily into runtime-ready keyframes. Another major one is IK constraints, which modify data between keyframes in complex ways.

The most common general solution, as far as I know, is to resample the animation (perhaps at 30fps) with linear interpolation. Tools like gltfpack can prune and optimize the animation from there, to remove unneeded frames with minimal impact on visual output.

jjcasmar:
I understand that blender can do much more and some decision must be made, but still, the exporter can output cubicspline animations. As I want to match the output when using the property animation extension, I want to understand how they are generated. By looking at the code is not clear to me the math happening

scurest:
I also want to know. It's this right?

https://github.com/KhronosGroup/glTF-Blender-IO/blob/4936b5b2779c3500e834d7fba253100627c10b33/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py#L296-L299

The glTF out-tangent is just the right derivative of the interpolated curve at the keyframe so the obvious thing to do is to use the right derivative of the Blender curve, which is just the slope of the line from the point on the curve to the handle:

    rise/run = (keyframe_pt.handle_right[1] - keyframe_pt.co[1]) /
               (keyframe_pt.handle_right[0] - keyframe_pt.co[0])

jjcasmar:
@emackey I algo opened that question, but more on the line of how to convert cubic spline to bezier (as it can be done exactly) . Here I want the other way round, how to get the best approximation of a bezier using a cubic spline and how is the exporter generating the tangents (and why you took that decisions if you know them :-) )

emackey:
Having explored glTF's `CUBICSPLINE` interpolation just recently for a new implementation, I don't think it's up to the task of replicating what Blender does without animation sampling.  I think this primarily for two reasons:

[...]

Issue #1436, opened 2021-07-19
Keyframe Interpolation issue of exported file? 


gonnavis:
**Describe the bug**
Hello, I found that if uncheck "Always Sample Animation" ( thus just export keyframes and interpolation function ), the animation will has problem.
![image](https://user-images.githubusercontent.com/10785634/126149889-52d67626-dd9a-422f-b87a-006efbfa5c12.png)

**To Reproduce**
1. Download this blender file.
[Keyframe Interpolation issue.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/6840270/Keyframe.Interpolation.issue.zip)
2. Open a.blend and play the animation. You can see the blue box go through the top of the red box smoothly.
2. Export gltf/glb with "Always Sample Animation" unchecked.
3. Open it with Win10 3D Viewer or online gltf viewers
    https://github.khronos.org/glTF-Sample-Viewer-Release/
    https://gltf-viewer.donmccurdy.com/
    https://modelviewer.dev/editor/
    https://sandbox.babylonjs.com/
    https://playcanvas.com/viewer    
4. You can see the blue box paused on top of the red box for a short period.

**Expected behavior**
Whether check or uncheck "Always Sample Animation", all played smoothly in Win10 3D Viewer and online gltf viewers.

**Screenshots**
![image](https://user-images.githubusercontent.com/10785634/126153368-b2ab74c6-d4c4-4d85-ab98-0a4c97f690df.png)

**.blend file/ .gltf**
[Keyframe Interpolation issue.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/6840270/Keyframe.Interpolation.issue.zip)

**Version**
 - OS: [ Windows10 ]
 - Blender Version [ 2.93.1 ]

**Additional context**
If export gltf/glb with "Always Sample Animation" checked ( thus baked to per frame animation datas ), will looking smoothly with Win10 3D Viewer and all the online viewers, as the origin a.blend file looked in Blender.

julienduroure:
Confirmed.
One more test : Reimporting in Blender -->Seems to be smooth...
![image](https://user-images.githubusercontent.com/1696020/129063042-ad2d467d-d6b9-4474-abd6-373b6436a13c.png)

julienduroure:
Closing, this is now solved

PR #1679, opened 2022-07-01
Fixed bezier control points to cubic spline tangents conversion

Andari85:
Related to:
https://github.com/KhronosGroup/glTF-Blender-Exporter/issues/90
https://github.com/KhronosGroup/glTF-Blender-IO/issues/960
https://github.com/KhronosGroup/glTF/issues/1778
https://github.com/KhronosGroup/glTF-Blender-Exporter/commit/42c46350f06e7401b941474c81bae60837a86ade
The commit above does not seem to have made it to this new repository.
Exported animations now maintain derivatives of curves at keyframes.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_animation_sampler_keyframes.py
@@ -409,8 +409,15 @@ def gather_keyframes(blender_obj_uuid: str,
                     key.set_first_tangent()
                 else:
                     # otherwise construct an in tangent coordinate from the keyframes control points. We intermediately
-                    # use a point at t-1 to define the tangent. This allows the tangent control point to be transformed
-                    # normally
+                    # use a point at t+1 to define the tangent. This allows the tangent control point to be transformed
+                    # normally, but only works for locally linear transformation. The more non-linear a transform, the
+                    # more imprecise this method is.
+                    # We could use any other (v1, t1) for which (v1 - v0) / (t1 - t0) equals the tangent. By using t+1
+                    # for both in and out tangents, we guarantee that (even if there are errors or numerical imprecisions)
+                    # symmetrical control points translate to symmetrical tangents.
+                    # Note: I am not sure that linearity is never broken with quaternions and their normalization.
+                    # Especially at sign swap it might occure that the value gets negated but the control point not.
+                    # I have however not once encountered an issue with this.
                     key.in_tangent = [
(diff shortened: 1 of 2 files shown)


Andari85:
Sorry for the force push, I just quickly corrected pushing with wrong credentials.

julienduroure:
Thanks for your PR
I will have a look soon.

scurest:
FWIW, I didn't test it, but this looks correct to me.

I had to check how the comment about "intermediately use a point at t+1 to define the tangent" works: it's clever, instead of the code needing to know how to transform both the ordinate and the derivative

$$
\begin{align}
T &: f(t) \mapsto Af(t) + B \\
T' &: f'(t) \mapsto Af'(t)
\end{align}
$$

by working with f(t)+f'(t), which is the value of (the tangent line to f at t) evaluated at t+1, the code only needs to know how to transform ordinates

$$
\begin{align}
T &: f(t) \mapsto Af(t) + B \\
T &: f(t) + f'(t) \mapsto A(f(t) + f'(t)) + B
\end{align}
$$

because the transformed derivative can be recovered by subtraction $T(f(t)+f'(t)) - T(f(t)) = Af'(t)$.

The comment for in_tangents claims instead to "use a point at t-1 to define the tangent". In this case it would work with f(t)-f'(t) and recover the transformed derivative by reverse subtraction $T(f(t)) - T(f(t)-f'(t)) = Af'(t)$. It's not clear to me which one of these two forms the old code was actually trying to use, but this patch is definitely using the "t+1" form, not the "t-1" form, in both cases (which seems easier anyway).

So @Andari85, the only thing I'd say is update the comment near key.in_tangent to say "use a point at **t+1** to define the tangent" since that's what it does.

Andari85:
Thanks for the note. I added more info to the comment.

julienduroure:
@Andari85 Seems your didn't signed the CLA with the last github user you used to push

Andari85:
Oh no, not again...

julienduroure:
I didn't test it yet, but here is a (maybe) related report: #1436

scurest:
> here is a (maybe) related report: https://github.com/KhronosGroup/glTF-Blender-IO/issues/1436

This fixes it.

Andari85:
Yes, it is described exactly as the issue I had to solve with this.
When may we expect it to be merged?

julienduroure:
> When may we expect it to be merged?

I plan to review/test (but seems OK) and merge next week

julienduroure:
Thanks a lot for your PR!

Andari85:
Thank you for handling it quickly and accepting it!

PR #1688, opened 2022-07-20
Fix #1687 - export gather_tree: improved performance ...

### LINKED ISSUE #1687: Slow export for scene containing 4000 objects
When exporting a scene with 4000 objects, with only one of them visible, I noticed that performance is problematic - even when exporting only the visible object.

I came across this problem with a scene containing many variants of kitchen furniture elements, I can't provide this file but I provide a synthetic replacement.

1. Open the attached blend file (containing 4000 cubes)
2. Export as GLB using the following script snippet, to export the single visible cube

```python
import bpy
bpy.ops.export_scene.gltf(filepath=r'4000cubes-b3d_3.2.glb', use_visible=True, use_active_scene=True, export_materials='PLACEHOLDER', export_animations=False, export_morph=False, export_skins=False)
```

3. I obtain the following result (2+ seconds to export the single cube)

```
16:40:43 | INFO: Draco mesh compression is available, use library at C:\nllbin\blender-3.3.0-alpha+master.7f4ee97b9ef9-windows.amd64-release\3.3\python\lib\site-packages\extern_draco.dll
16:40:43 | INFO: Starting glTF 2.0 export
16:40:45 | INFO: Extracting primitive: Cube.1643
16:40:45 | INFO: Primitives created: 1
16:40:45 | INFO: Finished glTF 2.0 export in 2.051008939743042 s
```

**Expected behavior**
The export for this single cube should be faster : I get less that 0.1 second when exporting a blend file with a single cube object.

**.blend file/ .gltf**
[4000cubes-b3d_3.2.blend.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/9151109/4000cubes-b3d_3.2.blend.zip)

**Version**
 - OS: Windows 10 x64
 - Blender Version `3.3-alpha+master.7f4ee97b9ef9`

**Additional context**
I've been profiling the export time using `cProfile` :

```py
import bpy
import cProfile
with cProfile.Profile() as pr:
    bpy.ops.export_scene.gltf(filepath=r'4000cubes-b3d_3.2.glb', use_visible=True, use_active_scene=True, export_materials='PLACEHOLDER', export_animations=False, export_morph=False, export_skins=False)

import pstats
from pstats import SortKey
p = pstats.Stats(pr)
p.sort_stats(SortKey.CUMULATIVE).print_stats(30)
```

and I noticed that :

* a majority of the time is spent in `gltf2_blender_gather_tree.py`
* due to many calls to `blender_object.children`
* `children` is a slow method, the [doc](https://docs.blender.org/api/current/bpy.types.Object.html?highlight=children#bpy.types.Object.children) mentions that it "takes `O(len(bpy.data.objects))` time."

rotoglup:
... by not calling blender_object.children, which is slow : takes O(len(bpy.data.objects)) operations.

Related to issue https://github.com/KhronosGroup/glTF-Blender-IO/issues/1687

Instead, I build a `dict` containing parent/child relations, which is used during the traversal.

I measured time spent before/after the patch, using b3d 3.3-alpha+master.7f4ee97b9ef9 ; I see improvement from 2.05sec to 0.06sec - about 40x faster in this case.

## before patch
```
16:40:43 | INFO: Draco mesh compression is available, use library at C:\nllbin\blender-3.3.0-alpha+master.7f4ee97b9ef9-windows.amd64-release\3.3\python\lib\site-packages\extern_draco.dll
16:40:43 | INFO: Starting glTF 2.0 export
16:40:45 | INFO: Extracting primitive: Cube.1643
16:40:45 | INFO: Primitives created: 1
16:40:45 | INFO: Finished glTF 2.0 export in 2.051008939743042 s
```

## timing after patch, b3d 3.3-alpha+master.7f4ee97b9ef9
```
16:43:00 | INFO: Draco mesh compression is available, use library at C:\nllbin\blender-3.3.0-alpha+master.7f4ee97b9ef9-windows.amd64-release\3.3\python\lib\site-packages\extern_draco.dll
16:43:00 | INFO: Starting glTF 2.0 export
16:43:00 | INFO: Extracting primitive: Cube.1643
16:43:00 | INFO: Primitives created: 1
16:43:00 | INFO: Finished glTF 2.0 export in 0.06299805641174316 s
```

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_tree.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_tree.py
@@ -106,10 +106,18 @@ class VExportTree:
         bpy.context.window.scene = blender_scene
         depsgraph = bpy.context.evaluated_depsgraph_get()
 
-        for blender_object in [obj.original for obj in depsgraph.scene_eval.objects if obj.parent is None]:
-            self.recursive_node_traverse(blender_object, None, None, Matrix.Identity(4))
+        # Gather parent/children information once, as calling bobj.children is
+        #   very expensive operation : takes O(len(bpy.data.objects)) time.
+        blender_children = dict()
+        for bobj in bpy.data.objects:
+            bparent = bobj.parent
+            blender_children.setdefault(bobj, [])
+            blender_children.setdefault(bparent, []).append(bobj)
 
-    def recursive_node_traverse(self, blender_object, blender_bone, parent_uuid, parent_coll_matrix_world, armature_uuid=None, dupli_world_matrix=None):
+        for blender_object in [obj.original for obj in depsgraph.scene_eval.objects if obj.parent is None]:
+            self.recursive_node_traverse(blender_object, None, None, Matrix.Identity(4), blender_children)
+
+    def recursive_node_traverse(self, blender_object, blender_bone, parent_uuid, parent_coll_matrix_world, blender_children, armature_uuid=None, dupli_world_matrix=None):
         node = VExportNode()
         node.uuid = str(uuid.uuid4())
         node.parent_uuid = parent_uuid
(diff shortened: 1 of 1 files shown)


rotoglup:
Also, I've run the test suite without getting any regression.

julienduroure:
Looks good to me.
Any reason why you pass blender_children as parameter in recursive function instead of creating it as VExportTree attribute and accessing it using self.blender_children?

rotoglup:
No reason except me being reluctant to add attributes when I don't fully
understand code and objects lifetimes.

I'm on vacation for now, if you'd like me to adjust, I can do so at the end
of the month.

Thanks

Le ven. 5 août 2022 à 15:46, Julien Duroure ***@***.***> a
écrit :

julienduroure:
Hello,
Lifetime is a good point. VExportTree instance will survive during all export, when blender_children is useful only during vtree construction.
I think I will commit it as is.
Thanks again, and enjoy your holidays!

Issue #1694, opened 2022-07-26
Suggestion: "glTF Settings" cleanup


donmccurdy:
The addition of the "glTF Settings" node to the "Output" dropdown is helpful, but does expose a couple small consistency/naming/docs issues —

![Screen Shot 2022-07-26 at 11 27 37 AM](https://user-images.githubusercontent.com/1848368/181047791-b9aa9a19-de2b-4545-8523-d7edc59fbff0.png)

1. Note the missing tooltip for this item, replaced by "(undocumented operator)". All the other nodes say, "Add a node to the active tree", which is not terribly helpful but okay.
2. Should we consider renaming this "glTF Output" or "glTF Material Output" rather than "glTF Settings" for consistency? Or supporting both, with Output as the default? The node is effectively an output for export purposes.

![Screen Shot 2022-07-26 at 11 40 39 AM](https://user-images.githubusercontent.com/1848368/181049825-cceea714-4ca9-4487-872d-d4b94c4b8178.png)

I'm not feeling so confident about the "glTF Original PBR Data" node. Its name does not match the button shown in the UI, possibly it should just be part of the other glTF Settings/Output node. And Blender 3.3 is the upcoming LTS release, which means we might want to be careful about adding features in that release. Especially with the upcoming Principled BSDF v2 work (which sounds very promising!) I'm not sure we should be planning to maintain this additional node in an LTS release. I'd be inclined to remove it from the menu until a later date.

Thank you, and let me know if PRs would be welcome for any of these changes.

/cc @emackey

donmccurdy:
Points from @emackey  in https://github.com/KhronosGroup/glTF-Blender-IO/issues/1686 is also good. 👍🏻

julienduroure:
> undocumented operator

Will fix it after my holidays

> glTF Material Output

Ok to rename. I will included it in 3.3  ( but I find a  way to keep compatibility with older files, created with old name)

> glTF Original PBR Data

I will fix name consistency between node and menu. 
I disagree about remove it from 3.3
People need a way to export exactly what they want about specular textures/data. I already spend so much of my time explaining why some features are not 1 to 1 between Blender and glTF. This will be the case for specular too.
 Basic users will not use it, but having a way to export exactly needed data for advanced users is a good feature I think. 
Principled v2 will of course not be part of 3.3, so using this node is for me the way to go until v2 is ready (not planned yet I think)

donmccurdy:
[...]

PR #1698, opened 2022-08-05
rename glTF Settings  to glTF Material Output

julienduroure:
Part of #1694


julienduroure:
@donmccurdy

donmccurdy (approved):
I prefer this update because it's very consistent with "Material Output", but wanted to check this sounds okay to others  — @emackey  any preference?

Thanks for including backward-compatibility with the previous name as well!

emackey (approved):
Sounds good to me.

PR #1700, opened 2022-08-05
Fix #1686 default thickness is now 0.0

### LINKED ISSUE #1686: Default thickness value
Hi, following #1646 in the updated "glTF Settings" node, the "thickness" value should default to zero, not one.  This matches glTF's own default thicknessFactor of zero.

I think it's appropriate to write `KHR_materials_volume` on export when (a) thickness is non-zero or textured, and (b) transmission is non-zero or textured.  This can have an effect even in cases where absorption/attenuation/density is not being used on the volume.

julienduroure:
Fix #1686 
@emackey @donmccurdy


donmccurdy (approved):
Looks good to me!

PR #1702, opened 2022-08-05
Merge all custom socket into a single node

julienduroure:


julienduroure:
@donmccurdy @emackey

addons/io_scene_gltf2/blender/com/gltf2_blender_material_helpers.py
```diff
@@ -22,23 +22,11 @@ def create_settings_group(name):
...
-
-def get_gltf_pbr_non_converted_name():
-    return "original glTF PBR data"
-
-def create_gltf_pbr_non_converted_group(name):
-    gltf_node_group = bpy.data.node_groups.new(name, 'ShaderNodeTree')
-
     specular = gltf_node_group.inputs.new("NodeSocketFloat", "specular glTF")
     specular.default_value = 1.0
     specularColor = gltf_node_group.inputs.new("NodeSocketColor", "specularColor glTF")
```
donmccurdy, line 27: Suggested naming, for consistency with other nodes —

```suggestion
    specular = gltf_node_group.inputs.new("NodeSocketFloat", "Specular")
    specular.default_value = 1.0
    specularColor = gltf_node_group.inputs.new("NodeSocketColor", "Specular Color")
```

I'm happy to make this change myself in a follow-up PR, if that would save you a bit of time.

addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_materials_specular.py
```diff
@@ -24,8 +24,8 @@
 def export_original_specular(blender_material, export_settings):
     specular_extension = {}
 
-    original_specular_socket = gltf2_blender_get.get_socket_original(blender_material, 'specular glTF')
-    original_specularcolor_socket = gltf2_blender_get.get_socket_original(blender_material, 'specularColor glTF')
+    original_specular_socket = gltf2_blender_get.get_socket_old(blender_material, 'specular glTF')
+    original_specularcolor_socket = gltf2_blender_get.get_socket_old(blender_material, 'specularColor glTF')
```
donmccurdy, line 28: See above:

```suggestion
    original_specular_socket = gltf2_blender_get.get_socket_old(blender_material, 'Specular')
    original_specularcolor_socket = gltf2_blender_get.get_socket_old(blender_material, 'Specular Color')
```

donmccurdy (approved):

addons/io_scene_gltf2/blender/com/gltf2_blender_material_helpers.py
julienduroure: This is not as simple as simple naming change.
Current channel mapping during texture creation (see __get_image_data & __get_image_data_mapping) is based on socket name.
Because there is already a socket named "specular" on Principled shader, not sure we can have a second one. 
We need a way to differentiate the two ones, by checking socket.node.type for example

donmccurdy: Hmm I see – thank you! I'll think about this more and open a PR or issue if I can think of anything, but for now I'd be happy to go ahead with this PR. I think consolidating the sockets onto one node is a good UX improvement help already.

julienduroure: I just pushed the name changes + changing checks to take the "same socket name on two different nodes" issue into account

PR #1723, opened 2022-09-09
Clarify user extensions import variable.

Exairnous:
Currently there are two variations of the `import_settings` variable, one which contains the actual import settings, and another that is the suggested parameter for user import extensions, but actually refers to the gltf data structure.  I suspect that the user import extensions variable was named this way to be consistent with the `export_settings` variable for the user export extensions, but `export_settings` does actually refer to the export settings and so they aren't equivalent.

This PR modifies the examples and the `import_user_extensions` function to more clearly convey what is actually being passed and prevent any confusion.


julienduroure:
Thanks for this clarification, you were right!

Exairnous:
Excellent!  You're welcome :)

PR #1726, opened 2022-09-10
Fix #1725 Texture can now be plugged directly on socket

### LINKED ISSUE #1725: Manage texture plugged directly on socket without separate node
For texture when channel is fixed by specification (ie thickness for volume, metallic for PBR, etc...), currently a sperateColor node is needed.
But texture can be plugged directly on socket. In this case, export is wrong (channel R is used by default for now)

Capture d’écran du 2022-09-10 09-31-23

julienduroure:
Fix #1725 Texture can now be plugged directly on socket


julienduroure:
If No SeparateColor node found, so take the specification channel that is needed
So export is correct if user plug the texture directly to the socket
Currently on master, the R channel is taken by default.

This still need to be fully tested to be sure there is no regression

julienduroure:
TODO :
- [ ] Crash when texture is plugged on Color, but socket is int (and should be linked to Alpha)

PR #1733, opened 2022-09-17
Fix #1677 Objects parented to bones animated

### LINKED ISSUE #1677: Object Parenting To Bone and Bone Constraint ordering issue
When an object is parented to a Bone, which have a constraint that changes it's transforms, the exporter seems to: Reset the bone transform to rest pose > parent the transformed Object to the un-transformed Bone > re-activate the bone's constraint, displacing the bone AND the object AGAIN.  This results in the object having twice the Bone's constraint transformation.

For example, if my constraint rotate the bone 15deg. , the child object will be rotated 30deg. So far I only tested and confirmed this behavior with the 3 "Track" constraints. 

1. Create a piston type Rig (with 2 bones tracking each other)
2. Set the resting position of a bone be different than the result of it's constrained position. (Meaning that switching the Armature from Rest to Pose will change the rotation of the bones).
3. I think that an animation is necessary to see the bug happens, not certain though.
4. Export a .glb with Blender 3.2

Expected Behavior:
In Blende4r, it works fine, but the GLB export give this bug. It worked well in Blender 3.1.1.

**Screenshots**
Rest Position:
image
Pose Position:
image
ModelViewer Result with a 3.2 export:
image
ModelViewer Result with a 3.1.1 export:
image

Blend File with glb exports: 
[Setup piston glTF.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/9004784/Setup.piston.glTF.zip)

Let me know if I can provide more information.

julienduroure:
Fix #1677 Objects parented to bones animated


Balzx:
Gausah spam napa

Pada tanggal Sab, 17 Sep 2022 23.31, Julien Duroure <
***@***.***> menulis:

PR #1758, opened 2022-10-10
Cleanup mesh caching

julienduroure:


julienduroure:
Any feedback is welcome.
I want to be sure that I not forgot any corner case

PR #1762, opened 2022-10-14
Handle ortho camera frame correctly. (Closes GH-944, GH-624.)

### LINKED ISSUE #624: Orthographic Camera Lens and FOV exported wrong
**Describe the bug**
The orthographic camera isn`t exportet right. It Ignores the Resolusion and the orthographic size.

**To Reproduce**
1. Create Blend file with plane
2. Add a texture to the plane (preferably something you can see easy if it is stretched)
3. Create a camera and switch its type to orthographic.
4. Export
5. The issue is seen in several gltF viewers.

**Expected behavior**
I expect that my Resolution and my camera size will be the same as I assigned in blender

**Screenshots**
This is how it looks in perspective
image

This is how it looks with the orthographic camera
image

**.blend file/ .gltf**
https://drive.google.com/file/d/1lOF9cn86ByZjQQ4LEyQHb1xpnSwktaUD/view?usp=sharing

**Version**
 - OS:Windows 10 64 Bit
- Blender V 2.8

**Additional context**

### LINKED ISSUE #944: Possible bug in ortographic camera export
I am not sure if this is really a bug, but in the ortographic camera exporter, there is something like this

```
        orthographic.xmag = blender_camera.ortho_scale
        orthographic.ymag = blender_camera.ortho_scale
```
both have the same value

This looks odd to me, but I am not sure if its really a bug or not. Can someone have a look, please?

will-ca:
Blender:

image

Before:

image

After:

image

Tested with Babylon because I believe Three is currently also non-conformant.
https://github.com/mrdoob/three.js/issues/24800

Spec:

https://github.com/KhronosGroup/glTF/issues/1663

https://github.com/KhronosGroup/glTF/pull/2053

https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/Specification.adoc#31034-orthographic-projection

```ADoc
==== Orthographic projection

Let

- `r` be half the orthographic width, set by `camera.orthographic.xmag`;
- `t` be half the orthographic height, set by `camera.orthographic.ymag`;
- `f` be the distance to the far clipping plane, set by `camera.orthographic.zfar`;
- `n` be the distance to the near clipping plane, set by `camera.orthographic.znear`.
```

Closes #944.
Closes #624.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_cameras.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_cameras.py
@@ -68,9 +68,17 @@ def __gather_orthographic(blender_camera, export_settings):
             zfar=None,
             znear=None
         )
+        
+        _render = bpy.context.scene.render
+        scene_x = _render.resolution_x * _render.pixel_aspect_x
+        scene_y = _render.resolution_y * _render.pixel_aspect_y
+        scene_square = max(scene_x, scene_y)
+        del _render
+        
+        # `Camera().ortho_scale` (and also FOV FTR) maps to the maximum of either image width or image height— This is the box that gets shown from camera view with the checkbox `.show_sensor = True`.
 
-        orthographic.xmag = blender_camera.ortho_scale
-        orthographic.ymag = blender_camera.ortho_scale
+        orthographic.xmag = blender_camera.ortho_scale * (scene_x / scene_square) / 2
+        orthographic.ymag = blender_camera.ortho_scale * (scene_y / scene_square) / 2
 
         orthographic.znear = blender_camera.clip_start
         orthographic.zfar = blender_camera.clip_end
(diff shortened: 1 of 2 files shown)


julienduroure:
Hello,
Really happy to see you proposed some enhancements/fixed on this repository.
Thanks for this PR :)

I didn't check the code yet, but in your screenshot, the checker doesn't look square (even it is much better than before).
This is something like 132x87
I don't know is this is only a screenshot issue, or if there is still an issue in code?

will-ca:
Glad to be of help!

The stretching is an implementation detail of the viewer, to fill the browser window.

Technically the spec says applications aren't supposed to stretch the rendered image when camera aspect ratio doesn't match window proportions, but the Three and Babylon viewers currently do it anyway.

If I were to resize the browser to the right aspect ratio, the stretching would go away. The important thing is that the number of squares and the relative bounds are the same as in Blender.

julienduroure:
OK, thanks for your answer :)

PR #1767, opened 2022-10-16
Fix #1765 Color Attribute import fix:

### LINKED ISSUE #1765: Wrong color index exported (COLOR_0, COLOR_1)
I have a [gltf with COLOR_0 and COLOR_1](https://github.com/KhronosGroup/glTF-Blender-IO/files/9793361/Bro.zip).
I import it in blender and then export it in glTF (default parameter).

Issues:
- On Blender 3.3.1, COLOR_1 and COLOR_1 are swapped on export
- On Blender 3.4.0, there is only one channel exported (and it has the data of COLOR_1), even if I use both channels in the shader graph in shading tab.

(both channels have different data, so an easy repro is to simply export-reload to see the difference in shading tab).

julienduroure:
Fix #1765 Color Attribute import fix:
- The first imported one is now the rendered one
- Color Attribute Node does no more specify the layer, so rendered one is used


julienduroure:
@scurest Can you please have a look? Here is the PR based on your suggestion for better Vertex Color import

PR #1769, opened 2022-10-17
Expand display of most enum properties.

will-ca:
image image

*Entirely* subjective, but personally I like how this display type exposes all the options at a glance.

It might just be me, but personally I go through all the dropdowns anyway whenever I'm exporting a final copy of something, just to see all the options to be sure I have them set right. So it'd be faster to just have them all visible at the same time.

And I suppose from a UX perspective, most of the Enums are already hidden in subpanels that are closed by default anyway. So those subpanels already provide the the space-saving functionality of using a dropdown, and needing to open the dropdowns right after opening the subpanels feels like a redundant click.

Uses #1760 as a base. So if accepted, should not be merged before that one. ....I'll need to rebase for any further changes to #1760, and probably throw on a dummy commit after it's been merged so GH generates a clean diff view. (That one introduces a new enum, so I figured it should be in the tree for this.)

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -874,10 +874,12 @@ class GLTF_PT_export_geometry_material(bpy.types.Panel):
         sfile = context.space_data
         operator = sfile.active_operator
 
-        layout.prop(operator, 'export_materials')
+        col = layout.column()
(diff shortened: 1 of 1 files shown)


julienduroure:
Closing this quite old PR, as the UI system completely changed

PR #1774, opened 2022-10-21
Refactor Panels to remove boilerplate and code duplication.

will-ca:
A PR that *reduces* code line count instead of adding to it. How'z'about that?

Should have no changes in behaviour.

---

Technically I suppose the DRACO and possibly the Extensions behaviour is untested as I don't have those set up in the Blender install.

---

There is this `TODO` in a seemingly unrelated spot— I left it as I'm not sure what it's referring to:

https://github.com/KhronosGroup/glTF-Blender-IO/blob/0bfaa1f1c09bca623bf79a6f4fda67e745a3a2b6/addons/io_scene_gltf2/__init__.py#L127

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -707,27 +707,54 @@ class ExportGLTF2_Base(ConvertGLTF2_Base):
         pass # Is needed to get panels available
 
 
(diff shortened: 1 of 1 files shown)


julienduroure:
Closing this PR, as the complete UI changed from Panel Class to a new UI system (for drag&drop support)

PR #1776, opened 2022-10-23
Ignoring special attributes

julienduroure:


addons/io_scene_gltf2/blender/exp/gltf2_blender_gather_primitives_extract.py
```diff
@@ -148,6 +157,11 @@ def prepare_data(self):
     def define_attributes(self):
         # Manage attributes + COLOR_0
         for blender_attribute_index, blender_attribute in enumerate(self.blender_mesh.attributes):
+
+            # Excluse special attributes (used internally by Blender)
+            if blender_attribute.name in PrimitiveCreator.get_special_attributes():
```
will-ca, line 162: Why not a constant `set()` `SPECIAL_ATTRIBUTES = {'.select_vert', '.select_edge', ...}`?

The list is being reconstructed for each loop I think, and the containment check will be `O(n)`.

…Well, premature optimization. But then again, code cleanliness?

julienduroure: Yes, sure, will do it

PR #1786, opened 2022-11-06
Experiment exporting Geometry Nodes Instances

julienduroure:
Still experimental and has some open questions:
- [x] Nested Instancing instances are all at object level, but some empties are created for intermediate levels. Need to find a way to not export them
- [x] UI
- [x] Geometry Nodes can have both geometry and instances. To export correctly, need to activate both Apply Modifier & Export GN Instances. Do we need to allow export instancing only when Apply Modifier is enabled?
- [x] Merge code with Duplis Instances? => No, keep it sep.
- [x] Issue When export only selection. Some instances are exporting even if instancier is not selected
- [x] Instancing lights
- [x] Instancing curves (Seems API is not able to perform that)
- [x] Tests
- [x] Non Regression tests
- [x] In a next step, option to use EXT_mesh_gpu_instancing ? See https://github.com/takahirox/glTF-Blender-IO-EXT-mesh-gpu-instancing => Is now managed separatly in #1984 
- [x] Animated GN (Use Scene bake export for that)
- [x] UV generic attribute => Done in #1989 
- [ ] https://projects.blender.org/blender/blender/issues/108343 => Still in progress, but on Blender side

Capture d’écran du 2022-11-05 18-49-12


HugoLamarche:
Just have to say, spend the day yesterday to look into source code to see how to make this until I find this PR, this is fantastic !
Workflow from Blender to Godot is fixed 🥇

bluediamond87:
Definitely looking forward to this one.  I was just about to do research into this for game engines.  Currently I’m using Godot.

Ameobea:
Thank you so much for your work on this!

I know this PR is still in progress, but I'm very eager to try this out and had a question.

I see that support for `EXT_gpu_instancing` is marked as complete in the PR description.  When I run an export of a test file with GN instances, I see that the instances do make it into the output GLTF but they are all listed as separate child objects rather than true instances using the extension.

Does that functionality depend on having the https://github.com/takahirox/glTF-Blender-IO-EXT-mesh-gpu-instancing Blender plugin installed?  I tried setting it up, but it seems to be broken in modern versions of blender and fails to install due to a missing import error.

Thanks again for your hard work here; this will be a crucial feature and unlock a ton of exciting usecases.

julienduroure:
@Ameobea Sorry, this was a misclick. gpu instancing is not done yet

Ameobea:
Ah I see - no problem ty for clarifying.  I'll keep watching the progress here

Issue #1785, opened 2022-11-05
"Animation channel cannot target WEIGHTS when mesh does not have morph targets." when exporting from Blender


khalleu:
Hey! I'm fairly new to GitHub, please forgive me.

I posted [this](https://blender.stackexchange.com/questions/278529/unable-to-export-as-gltf-mesh-does-not-have-morph-targets?noredirect=1#comment478428_278529) thread on Stack Exchange and it was suggested that I report a bug here.

To give a rundown of the issue, my .glTF exports cause Microsoft's 3D Viewer to crash before I'm even able to view the file. I opened the file in [this](https://gltf-viewer.donmccurdy.com/) online .glTF viewer to try to pinpoint the issue, and got the error message

`Animation channel cannot target WEIGHTS when mesh does not have morph targets.`

I converted every object into a mesh and ensured that each object had at least one shape key, but the issue prevailed. Is this a bug?
You can find the project files [here](https://github.com/KhronosGroup/glTF-Blender-IO/files/9944749/project.zip).

willstott101:
I'm also having trouble.
I don't think I can share the blender file I'm having trouble with.
But I can say that using Blender `append` from the troublesome file, of one mesh, allows me to export morph targets OK.
Exporting that same mesh from the troublesome file (and only that mesh, using `Export Selected`) the `targets` attribute on the primitive in the JSON GLTF file is missing...

willstott101:
Ok, here is:
* A blender file exhibiting my problem
* A GLTF export from that file missing the `meshes[].primitives[].targets` key
* A fresh 3.3.1 blender file with the object I am testing with appended into it
* A GLTF export from that file with working morph targets

https://drive.google.com/drive/folders/1m1Ng4aFuSYsRCCKyqZRV5wSnaTd7fK_F?usp=share_link

I'm piggy-backing on this issue as I'm seeing a very similar symptom, I apologise if it turns out to be unrelated.

julienduroure:
Can you please also make screenshots of options you are using at export?

julienduroure:
Hello,
This is now fixed in current main branch, after animation refactor 64e3532f48ebdf878a1c9b81ade6d1a19c8ff890

Because of are not exporting actions library, you can use "Scene" animation option at export

PR #1788, opened 2022-11-10
WIP: Try testing morph target exports

willstott101:
I don't seem to be able to export morph targets, I thought I'd try updating the tests - but am having trouble running them locally. So opening a PR to see them on a CI system

Related to #1785


willstott101:
It's something specific to my file... can't reproduce it like this.

PR #1819, opened 2023-01-04
Fix MIX output socket 2

### LINKED ISSUE #1817: import KHR_materials_sheen - node link not created - ShaderNodeMix issue in blender >= 3.4?
**Describe the bug**
For some reason the node link is not created in Blender 3.4.

After some digging it seems `ShaderNodeMix` has _three_ outputs (?!).
Apparently separate outputs for 'Float', 'Vector' and 'Color'. Odd...

My guess is this is a Blender bug.

Linking to outputs[2] `mh.node_tree.links.new(sheenColor_socket, node.outputs[2])` seems to fix the issue.

See [io_scene_gltf2/blender/imp/gltf2_blender_KHR_materials_sheen.py#L60](https://github.com/KhronosGroup/glTF-Blender-IO/blob/master/addons/io_scene_gltf2/blender/imp/gltf2_blender_KHR_materials_sheen.py#L60)

Of course the same issue is at play whenever `ShaderNodeMix` is used. (sheen roughness for example).

**To Reproduce**
1. Import attached velvet.glb
2. Go to 'shading'
3. Select the object
4. Link between mix shader and velvet bdsf's inputs[0] is missing

**Expected behavior**
Link should be created

**Screenshots**
image
image

**.blend file/ .gltf**
[velvet.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/10345640/velvet.zip)

**Version**
 - OS: win64 / windows 11
 - Blender Version 3.4.1

**Additional context**
As said: this is probably a bug of the newly introduced `ShaderNodeMix` in Blender 3.4 and up
Filed a bug at Blender https://developer.blender.org/T103642 (assuming it's a bug and not a feature)

julienduroure:
Fix #1817

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
+++ b/addons/io_scene_gltf2/blender/exp/gltf2_blender_get.py
@@ -52,11 +52,20 @@ def get_node_socket(blender_material, type, name):
     :return: a blender NodeSocket for a given type
     """
     nodes = [n for n in blender_material.node_tree.nodes if isinstance(n, type) and not n.mute]
-    nodes = [node for node in nodes if check_if_is_linked_to_active_output(node.outputs[0])]
+    nodes = [node for node in nodes if check_if_is_linked_to_active_output(node.outputs[__get_output_socket(node)])]
     inputs = sum([[input for input in node.inputs if input.name == name] for node in nodes], [])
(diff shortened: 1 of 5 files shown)


julienduroure:
Replaced by #1821

PR #1831, opened 2023-01-17
Remove usage of deprecated np.bool alias

oreo639:
https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

np.bool was deprecated in numpy 2.20 removed in 2.24.

This fixes importing glTF models with numpy 2.24


julienduroure:
Hello,

Blender 3.5 will be shipped with numpy 1.23.5
There is no absolute need to hurry, the deprecation will not happen on 3.5
But because 3.5 is still in bcon2, let's merge it before bcon3 of 3.5

julienduroure:
Merged, thanks!

PR #1834, opened 2023-01-21
Fix #1833 - uri dir sep doublebackslash

### LINKED ISSUE #1833: Import: URIs containing un-escaped \ behave differently on Windows, Unix
Some misbehaving glTF producers (eg. assimp) may produce invalid URIs containing `\`. When imported, backslashes pass through `urllib.parse.unquote` unchanged and get passed on through `open` to the OS. On Windows the `\` will work as a folder separator and the URI will silently work. On Unix it will be treated as a normal filename character and (probably) fail.

The behavior should be the same on both OSes.

Note that the glTF validator does not flag URIs with a backslash as invalid. It treats it as though it had been properly percent-encoded, `%5C`.

#### Solution

I think the URI should be repaired by replacing backslashes with forward slashes, ie `uri = uri.replace('\\', '/')` before unquoting. The reasons:

1. This is what browsers do when they find a backslash. It's what the WHATWG's URL spec says to do.
2. Backslashes are almost certain to be improperly encoded Windows folder separators (they are not common in Unix filenames).

The alternative is to raise an error and refuse to use the path.

Also the `_uri_to_path` function used for loading images should be moved out so it can be used for .bins as well. The change would go there.

julienduroure:
Fix #1833

### DIFF
--- a/addons/io_scene_gltf2/io/com/gltf2_io_path.py
+++ b/addons/io_scene_gltf2/io/com/gltf2_io_path.py
@@ -14,6 +14,7 @@
 
 from urllib.parse import unquote, quote
 from os.path import normpath
+from os import sep
 
 def uri_to_path(uri):
     uri = uri.replace('\\', '/') # Some files come with \\ as dir separator
@@ -21,5 +22,6 @@ def uri_to_path(uri):
     return normpath(uri)
 
 def path_to_uri(path):
+    path.replace(sep, '/')
     path = normpath(path)
     return quote(path)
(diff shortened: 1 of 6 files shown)


julienduroure:
Added fix of #1836 inside the same PR, as they are both related to uri encoding (one for importer, other for exporter)

addons/io_scene_gltf2/io/com/gltf2_io_path.py
```diff
@@ -0,0 +1,27 @@
...
+
+def uri_to_path(uri):
+    uri = uri.replace('\\', '/') # Some files come with \\ as dir separator
+    uri = unquote(uri)
+    return normpath(uri)
+
+def path_to_uri(path):
+    path.replace(sep, '/')
+    path = normpath(path)
+    return quote(path)
```
scurest, line 27: Please just copy the _path_to_uri code. The first line does nothing, and even if it worked, normpath replaces / with \.

PR #1845, opened 2023-01-29
Fix #1227 filter_glob update when changing export type

### LINKED ISSUE #1227: [Bug] .GLTF files are displayed when GLB format is selected
When you export to gltf, there's an option to select which format you want to use: GLTF or GLB.
But no matter which one you choose, the explorer will always show both GLTF and GLB files.
img
**Expected behavior:** show only .glb files when expoting to GLB;
show only .gltf files when exporting to GLTF

julienduroure:
Fix #1227 
When changing export type (glb / gltf embeded / gltf separated), it changes automatically the filter in file export list.
(When GLB => displays only *.glb, otherwise => displays only *.gltf)

There is currently a bug in Blender preventing the list to be refreshed automatically.
Update is done when clicking on "Refresh File List"
https://developer.blender.org/T104221

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -110,6 +110,8 @@ def on_export_format_changed(self, context):
 
     # Also change the filter
     sfile.params.filter_glob = '*.glb' if self.export_format == 'GLB' else '*.gltf'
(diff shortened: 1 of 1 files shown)


julienduroure:
List can be forced to be update using python. I updated the PR.
Any comments?

Issue #1837, opened 2023-01-23
Use tabs instead of spaces for glTF export, or reduce/eliminate indentation.


lyuma:
**Describe the bug**
A glTF document exported by Blender is over 50% whitespace (25% whitespace if counting the .bin file). While this is small compared to textures, in certain applications where large numbers of nodes are used, it can be significant, especially in environments where transport-layer compression is not possible (given that glTF documents are not gzip or brotli compressed).

**To Reproduce**
1. Open a blend file such as the one below
2. Click on File->Export->glb/gltf
3. Select glTF Separate (.gltf+.bin+textures) and Visible Objects inside Include
4. Check filesize of exported document.

**Expected behavior**
I expect glTF documents to be reasonably efficient, given that these files may also be used for transport.
This seems like a small change which can still be a win for efficiency and file size.

I know whitespace formatting is widely considered a subjective and inflammatory subject, but I think it can be argued logically in this case: The disk space savings are significant, and is not common/expected that users open generated glTF documents, so there is no need to make them look aesthetically pleasing at such substantial cost.

**Screenshots**
![screenshot of VS code showing abundant whitespace](https://user-images.githubusercontent.com/39946030/214177575-b3961720-745b-46ba-ba63-e7377f4e82e5.png)

**.blend file/ .gltf**
Blend file is zipped here.
https://github.com/godotengine/godot/files/10153278/GithubBrokenAnimationsProject.zip
glTF documents exported using glTF Separate as per Godot recommendations, with Include -> "Visible Objects" selected.
[PlayerFPS_testexp.orig.gltf (1599KB)](https://github.com/KhronosGroup/glTF-Blender-IO/files/10484482/PlayerFPS_testexp.orig.gltf.json.github.txt)
[PlayerFPS_testexp.tabs.gltf (930KB)](https://github.com/KhronosGroup/glTF-Blender-IO/files/10484484/PlayerFPS_testexp.tabs.gltf.json.github.txt)
[PlayerFPS_testexp.no_indentation.gltf (707KB)](https://github.com/KhronosGroup/glTF-Blender-IO/files/10484485/PlayerFPS_testexp.no_indentation.gltf.json.github.txt)

**Version**
 - OS: Windows 10
 - Blender Version 3.4.1

**Additional context**

julienduroure:
The  question here is:
- What should be the default value. More compact json, or human readable for debugging?
[...]

PR #1907, opened 2023-04-14
Export option for minified JSON (.gltf).

jackladybird:


julienduroure:
Hello,
This has been discussed already, but not sure we should add one more option for that.
You can either use glb that have already minified json, or use a hook that can minify it.

See https://github.com/KhronosGroup/glTF-Blender-IO/issues/1837
See https://github.com/KhronosGroup/glTF-Blender-IO/pull/1587


``` 
def gather_gltf_encoded_hook(self, gltf_format, sort_order, export_settings):
    	gltf_format.indent =  None
    	gltf_format.separators = (',', ':')
```

julienduroure:
Closing this old PR, as we will not use a (one more )new option to manage it.
See my previous comment for more info

PR #1938, opened 2023-06-12
Fix vertex attribute loading

elrnv:
This commit fixes loading vertex attribute data, which ought to be indexed with the unique indices extracted from each primitive's index set.

Attached is a test gltf file that fails to import into Blender without this fix.
[tet_and_tri.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/11717335/tet_and_tri.zip)

I am happy to include a test for this, but I need some guidance for how to run the tests for this plugin.


julienduroure:
Hello, thanks for the reporting and fix.

For tests:
- You should first fix your exporter, as it generate a not valid glTF file. You can use gltf validator https://github.khronos.org/glTF-Validator/ or check it also here: https://gltf-viewer.donmccurdy.com/
- You can add gltf here: https://github.com/KhronosGroup/glTF-Blender-IO/tree/main/tests/roundtrip
- And then test here: https://github.com/KhronosGroup/glTF-Blender-IO/blob/4115a53160fb56119c387ffe898c5641216d98da/tests/test/test.js#L1839

julienduroure:
BTW, I agree with your fix. I will merge it soon, as the 3.6 will be freeze next week.
If possible, can you please rebase it to be able to merge it on blender-v3.6-release branch?

elrnv:
It seems that the test is failing on mismatched drawCallCount, any advice on how to resolve this? A number of other tests seem to fail as well since the rebase, but it seems unrelated.

julienduroure:
I created #1941 with test fixes:
- Tests checks that gltf have a default scene, so I added one
- I added fake materials. If no material => primitives are merged at import, so only 1 at export
- Adding option to export with attributes, to be able to perform the check

elrnv:
ah, I think I misunderstood before. Thanks for fixing up this PR, I pulled your update and rebased on upstream/main. I'm also happy if you want to merge the other PR you made. You can also make changes to this PR directly if you like.

julienduroure:
Thanks!

PR #1945, opened 2023-06-26
This PR adds the option to run the produced gltf export through gltfpack with various options

deccer:
Hi there,

Juan and I thought of extending this plugin with the option to run the exported glTF through the tool [gltfpack](https://github.com/zeux/meshoptimizer) (aka meshoptimizer). The main motivation is to get ktx2 support.

By default gltfpack outputs regular .glb/.gltf files that have been optimized for GPU consumption using various cache optimizers and quantization and can also compress textures using Basis Universal format stored in a KTX2 container.

`gltfpack` will output to a subfolder `gltfpacked` in the folder you exported the glTF to.

For the plugin to be able to run the originally exported glTF, you need to have `gltfpack` installed on your system somewhere and configure its path in the plugin preferences, then `gltfpack` relevant options will also appear in the export dialog, where you can also decide if you want to make use of `gltfpack` or not, among the options you can pass through that tool.

Extension in action:
image

Its preferences too:
image

Any feedback is welcome


addons/io_scene_gltf2/__init__.py
```diff
@@ -1756,7 +1927,10 @@ def menu_func_import(self, context):
 
 def register():
     from .blender.com import gltf2_blender_ui as blender_ui
+
     for c in classes:
+        if (hasattr(c, "bl_label") is True and c.bl_label == GLTF_PT_export_gltfpack.bl_label):
```
julienduroure, line 1932: I think we should always register the class, and add a check in poll of UI to avoid display if not needed. This will avoid this check

addons/io_scene_gltf2/__init__.py
```diff
@@ -1765,6 +1939,8 @@ def register():
         blender_ui.variant_register()
     if bpy.context.preferences.addons['io_scene_gltf2'].preferences.animation_ui is True:
         blender_ui.anim_ui_register()
+    if bpy.context.preferences.addons['io_scene_gltf2'].preferences.gltfpack_path_ui is not False:
```
julienduroure, line 1942: Same comment. Always register, add a check in poll

addons/io_scene_gltf2/__init__.py
```diff
@@ -898,6 +1014,52 @@ def draw(self, context):
...
 
+class GLTF_PT_export_gltfpack(bpy.types.Panel):
+    bl_space_type = 'FILE_BROWSER'
+    bl_region_type = 'TOOL_PROPS'
+    bl_label = "gltfpack"
+    bl_parent_id = "FILE_PT_operator"
+    bl_options = {'DEFAULT_CLOSED'}
+
+    @classmethod
+    def poll(cls, context):
```
julienduroure, line 1025: Add a check to display only if the path is provided in preferences

julienduroure (commented):
Hello, Thanks for this PR
First of all, I need to check the Blender policy about calling some external tools.

julienduroure:
(And you should rebase with main branch, as I fixed some bugs after recent changes in Blender)

addons/io_scene_gltf2/__init__.py
deccer: Makes sense.

addons/io_scene_gltf2/__init__.py
deccer: Also makes sense. Had problems with poll first, but that's just me not knowing python at all.

addons/io_scene_gltf2/__init__.py
deccer: Also fixed. I hope this makes sense, I was able to reproduce both cases at least. 

One weird thing i noticed though is that poll seems to be run more than once (2 or 3 times) at all times... which seems weird, but is probably also just me not knowing how Blender calls various things and poll is one of those - I would assume.

deccer:
Thank you for taking a look :)

julienduroure:
Hello,
Can you please rebase with main?
I will merge it soon

addons/io_scene_gltf2/__init__.py
```diff
@@ -147,6 +148,102 @@ def __init__(self):
...
+                ('Normalized', 'Normalized', 'Use normalized attributes for positions'),
+                ('Floating-point', 'Floating-point', 'Use floating-point attributes for positions')),
+        default='Integer',
+    )
+    
+    export_gltfpack_noq: BoolProperty(
+        name='Disable Quantization',
+        description='Disable quantization; produces much larger glTF files with no extensions',
+        default=True,
+    )
```
donmccurdy, line 237: I believe these two defaults are in conflict. If quantization is disabled, integer vertex position attributes won't work. I think the uses of the "Vertex Position Attributes" flag in gltfpack are very limited —  perhaps it would be best to remove that option from the UI, and have quantization enabled by default?

julienduroure:
Thanks @donmccurdy to point it
I will propose a PR soon

PR #1955, opened 2023-07-19
Allow normalization in gather_attribute_change user extension

patrtoth:
This PR allows to mark data as normalized from a user extension like https://github.com/julienduroure/gltf-hooks-examples/blob/main/gltf_attribute_change.py

The motivation being, that one might want to transform a `Float` attribute into a `UnsignedByte` color attribute that requires the `"normalized":true` tag in the glTF 2.0 accessor to be valid.

The only change is exposing the normalization boolean and allowing it to be set in a user extension like
```python
data["normalized"] = True
```


patrtoth:
I added an example based on @julienduroure 's user extension that would make use of this change:
https://gist.github.com/patrtoth/7c1fad78e5264758d8d5a3b6e6cf1ea4
When enabled, this extension changes floating point precision vertex colors into raw unsigned byte colors during export.

PR #1968, opened 2023-08-16
Fix #1956 merge channels (metallic/Roughness) : use default value

### LINKED ISSUE #1956: Expected behavior of pbrMetallicRoughness's metallicRoughness textures
Hi, I have a scene with several materials that only have baseColor, roughnesss, and normal textures. Metallic is set to 0 in the material. When I export from Blender, I noticed that the metallicRoughness textures for these materials have 1.0 (or 255) for the R and the B channel. The G channel had the roughness data in it.

My understand is that R channel doesn't get used for pbrMetallicRoughness. But since I had metallic set to 0, I was expecting the B channel to have a vlaue of 0. Is there a step that I'm missing?

I'm loading through cgltf and I've double checked to make sure there weren't individual flags for metallic and roughness - and their factors. But both attributes are combined as metallicRoughness.

I couldn't find any discssion of this on the web or in this repo, but maybe my search-fu is weak? Curious to know what the expected behavior for the metallicRoughness texture is when one or the other doesn't have a texture?

julienduroure:
Fix #1956 
When we have to concatenate multiple channels into an image (for example MetallicRoughness).
If only 1 texture is provided, the other channel was filled with 1.0
Now this is filled with the default value of the socket


julienduroure:
Tests still need to be updated, as default value from sockets are not taken into account in existing test files

maoliver-amd:
This doesnt seem correct to me. I noticed an issue after updating to blender 4.0 which seems connected to this PR.
From my understanding of the spec the rough/metal value in the presence of a texture is equal to the base factor multiplied by the texture value. Currently the same value is being written to the base factor ASWELL as now to the texture component (e.g. if we had a base roughness of 0.1 then the output gltf has 0.1 written into the texture and 0.1 written to roughnessFactor which results in incorrectly calculated final roughness = 0.1*0.1).

The default value in a compacted texture should be 1.0 when no texture is connected. We actually internally check for this and dump the entire texture channel during decoding to save memory (which no longer works after this PR)

From the spec:
````
The value for each property MAY be defined using factors and/or textures (e.g., baseColorTexture and baseColorFactor). If a texture is not given, all respective texture components within this material model MUST be assumed to have a value of 1.0. If both factors and textures are present, the factor value acts as a linear multiplier for the corresponding texture values.
````

julienduroure:
@maoliver-amd 
Do you mean that you have a Metallic texture and no roughness texture, but a value of 0.1 for roughness?
Can not reproduce in that case.

Can you please open a ticket with a sample .blend file?

PR #1989, opened 2023-09-11
Manage UVMap outside of Materials

### LINKED ISSUE #1975: export&import involving two UV maps malfunctions for any materials referencing the second UV map.
**Describe the bug**
The GLTF export that ships with blender 3.6.2 LTS malfunctions by using the first UV map for all materials, even when the shader nodes refer to the second UV map.

**To Reproduce**
1. export the scene from the attached .blend as a .glb.
2. create a new blend file and import the .glb
3. activate Material Preview shading in the 3D viewport.
4. Grumble about how the materials do not match the original.

**Expected behavior**
The GLTF exporter should create .glb files that reference the correct UV map.

**Screenshots**
original
image

after export/import
image

**.blend file/ .gltf**
attached

**Version**
 - OS: linux
 - Blender Version 3.6.2

**Additional context**
NA

julienduroure:
In Blender, UVMaps are defined outside of Material (but in mesh data).
So for export materials, we now manage it in 2 times
- Export Material without texcoord data
- Then retrieve the correct data, and modify material if needed.

This is now:
- Fix #1975 
- Allow to use custom Attribute for UVMap data (that is a common usage in Geometry Nodes)


hedgehog90:
Just gave this fix a go but ran into an error from `gltf2_blender_gather_materials_clearcoat.py`:

`UnboundLocalError: local variable 'uvmap_info' referenced before assignment`

I think the last line is returning the wrong variable, `uvmap_infos` not `uvmap_info`.
Changing this fixed it and I was able to export successfully.

julienduroure:
@hedgehog90 Thanks, typo fixed!

PR #2014, opened 2023-09-26
Principled v2

julienduroure:
Better specular management


vis-prime:
`Specular IOR value` is 0 by default in principled v2 , is that a typo/bug  ?
old
new

vis-prime:
@julienduroure 

The default `Specular IOR value` is 0 in loaded GLTFs in Principled v2 is it a typo/bug ?, or has it already been fixed?
old
new

julienduroure:
Hello,
The 'olive' material does not have any specular data, so the value of Specular IOR level is set to 0.

Some adjustment may occurs in following weeks, until 4.0 is released.

PR #2023, opened 2023-10-11
Check parent type to determine if a node is really a child of a bone

eoineoineoin:
Fixes a bug where nodes would not be exported if they had previously been using BONE parenting but had later been changed to OBJECT parenting.

To reproduce:
1. Create a mesh
2. Create an armature with one bone
3. Set the mesh's parent to the bone in the armature
4. Change the mesh's "parent type" to object - Blender will change the parent to the armature instead of the bone but will *not* clear the `parent_bone` in the mesh object.
5. Export a glTF - the mesh won't appear in the output file


julienduroure:
Merged, thanks!

PR #2062, opened 2023-11-21
Fix a few typos

PiOverFour:
One of these fixes was already committed in Blender's repo, from blender/blender-addons@9331ce7.


julienduroure:
Thanks !

Issue #2075, opened 2023-12-07
New default bone display unusable for many armatures


DoctorPresto:
With 4.0 defaulting to blender for bone heuristic and using the iconosphere custom shape, while it's nice to not have to spend so much effort correcting the bone rotations, the visualization of the armatures themselves has become completely unusable for any of the armatures I work with in animation. It would be nice if a default size was used instead, with lines drawn between the bones in all views would be even nicer but not necessary.

To reproduce, either import one of the ..glb files or open one of the .blend files included
[human_armature.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/13603136/human_armature.zip)
[vehicle_armature.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/13603137/vehicle_armature.zip)
 or open one of the .blend files.

screenshots:
human_armature current: 
![image](https://github.com/KhronosGroup/glTF-Blender-IO/assets/111536029/7b9e9dc6-f7bc-4982-9ea3-308edd2d5905)

human_armature temperance:
![image](https://github.com/KhronosGroup/glTF-Blender-IO/assets/111536029/e8fb603d-3cc6-4916-b4c5-e50a1783eb04)

vehicle_armature current front:
![image](https://github.com/KhronosGroup/glTF-Blender-IO/assets/111536029/45c19c5b-52d5-4f69-a40e-cc6970f7931b)

vehicle_armature current side:
![image](https://github.com/KhronosGroup/glTF-Blender-IO/assets/111536029/c0c5cf97-8c6f-4da9-907e-bc6159df94f4)

vehicle_armature temperance front:
![image](https://github.com/KhronosGroup/glTF-Blender-IO/assets/111536029/65b5d1d7-1c22-4638-8145-6d33d7f68c44)

vehicle_armature temperance side:
![image](https://github.com/KhronosGroup/glTF-Blender-IO/assets/111536029/a0f9f3a5-7739-45a8-92b8-3441ae00d0cf)

**Version**
 - OS: windows 11
 - Blender Version 4.0

julienduroure:
Agree that the sphere size doesn't have an acceptable size in many cases.
Let's try to find a way to get it ok for all situation.

Currently, the sphere size is related to:
- the Complete Armature size
- The bone Size.

In some cases, the sphere is too large, in some others, it is too small.

See current here:
[...]

PR #2096, opened 2024-01-01
Change scaling factor of default custom bones 

DoctorPresto:
adjusted as discussed [in this issue ](https://github.com/KhronosGroup/glTF-Blender-IO/issues/2075), it's possible these may be slightly too small on some armatures, however it at least avoids the issue of the giant globes preventing use of the armature at all.
Likely the actual scaling factor should be exposed in the importer panel so that an end user can easily adjust the default size one way or the other as best suits the armatures they work with.


julienduroure:
Thanks!
I will maybe add a scale factor option in importer, in order to manage manually tweak if needed

PR #2105, opened 2024-01-09
Change naming format of imported Color Attributes

SSM0770:
Most UV Map imports (including ones for other Blender I/O types) usually follow the format of UV0, UV1, UV2, etc. I noticed that this importer didn't name them that way, so I'm looking to see if this change can be made.

image

(I tested it on Blender 4.1.0 Alpha and it appeared to be working as intended, but let me know if something turns out to be wrong with it)

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
@@ -323,7 +323,7 @@ def do_primitives(gltf, mesh_idx, skin_idx, mesh, ob):
     mesh.polygons.foreach_set('loop_total', loop_totals)
 
     for uv_i in range(num_uvs):
-        name = 'UVMap' if uv_i == 0 else 'UVMap.%03d' % uv_i
+        name = 'UV0' if uv_i == 0 else 'UV%01d' % uv_i
         layer = mesh.uv_layers.new(name=name)
 
         if layer is None:
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
@@ -323,7 +323,7 @@ def do_primitives(gltf, mesh_idx, skin_idx, mesh, ob):
     mesh.polygons.foreach_set('loop_total', loop_totals)
 
     for uv_i in range(num_uvs):
-        name = f'UV{uv_idx}'
+        name = 'UVMap' if uv_idx == 0 else 'UVMap.%03d' % uv_idx
         layer = mesh.uv_layers.new(name=name)
 
         if layer is None:
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
@@ -323,7 +323,7 @@ def do_primitives(gltf, mesh_idx, skin_idx, mesh, ob):
     mesh.polygons.foreach_set('loop_total', loop_totals)
 
     for uv_i in range(num_uvs):
-        name = 'UVMap' if uv_idx == 0 else 'UVMap.%03d' % uv_idx
+        name = 'UVMap' if uv_i == 0 else 'UVMap.%03d' % uv_i
         layer = mesh.uv_layers.new(name=name)
 
         if layer is None:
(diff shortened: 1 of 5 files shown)


scurest:
I don't think this should change. UV maps are named the same they would be named if you created them by hand in the Properties editor. I just downloaded the latest 4.1 and `UVMap` is still the default name (check the default cube for example).

Also the OBJ importer at least also calls them `UVMap`.

(OTOH, color attribute names need an update, since the importer [still calls them `Col`](https://github.com/KhronosGroup/glTF-Blender-IO/blob/2debd75ace303f3a3b00a43e9d7a9507af32f194/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py#L336) even though Blender has since changed to `Color`.)

SSM0770:
> I don't think this should change. UV maps are named the same they would be named if you created them by hand in the Properties editor. I just downloaded the latest 4.1 and `UVMap` is still the default name (check the default cube for example).
> 
> Also the OBJ importer at least also calls them `UVMap`.
> 
> (OTOH, color attribute names need an update, since the importer [still calls them `Col`](https://github.com/KhronosGroup/glTF-Blender-IO/blob/2debd75ace303f3a3b00a43e9d7a9507af32f194/addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py#L336) even though Blender has since changed to `Color`.)

When I refer to "Most UV Map imports" I was more referring to what most other software I've worked with calls them, namely Unity. (The FBX importer also goes by the `UV#` naming scheme, and I'd guess that decision was made Blender themselves, since the manual doesn't mention a place for user contributions)

I had been importing Unity assets for use in a project I'm working on, and for whatever reason get a bit thrown off when the UV I'm selecting doesn't have its name match, I'm probably not the only one that feels that way.

However, I do agree with needing to change `Col` to `Color`, the abbreviation is a bit unnecessary, but I had already made my edit before thinking about it so I'll leave that for someone else to decide if it needs changing.

scurest:
I think the importer should follow Blender's conventions, but it's not my call.

If you do want to do it, you can simplify it to `name = f'UV{uv_idx}'`, and you'll need to update [all the other places that use it too](https://github.com/search?q=repo%3AKhronosGroup%2FglTF-Blender-IO%20UVMap.%2503d&type=code).

SSM0770:
If they decide not to do this, how can I edit my own fork to work in older Blender versions so I can at least personally use it without doing a bunch of weird stuff?

To test this out I literally had to replace the default addon’s files within 4.1’s directory (It wouldn’t show up in the addons page when I tried to import it regularly, even after lowering the version requirement in _init_.py)

julienduroure:
I agree that we need to follow Blender convention, not other software ones.
- So UVMap should be named UVMap, not something else, as this is the name used in Blender
- Col need to be changed to Color, as this is now the default named used in Blender

julienduroure (changes requested):
Please keep only changes on Col => Color
UVMap should not be changed

SSM0770 (commented):
This should be okay I guess, go ahead.

Still, do you have an answer to the question I asked previously about making my fork installable on its own? I at the very least need the UV name changes for personal use with my own project.

addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
```diff
@@ -323,7 +323,7 @@ def __init__(self, skinning: bool = True, skin_into_bind_pose: bool = True):
     mesh.polygons.foreach_set('loop_total', loop_totals)
 
     for uv_i in range(num_uvs):
-        name = 'UVMap' if uv_i == 0 else 'UVMap.%03d' % uv_i
+        name = 'UVMap' if uv_idx == 0 else 'UVMap.%03d' % uv_idx
```
julienduroure, line 326: This is still wrong

julienduroure (changes requested):

addons/io_scene_gltf2/blender/imp/gltf2_blender_mesh.py
SSM0770: Sorry. It’s the middle of the night and I’m doing this entirely on my phone, so I was simply copy/pasting the same thing without paying attention to the fact that there was a difference on mesh. It’s fixed.

julienduroure: Take care ! There is no hurry to manage it on phone in the middle of the night!

julienduroure (approved):

PR #2107, opened 2024-01-10
Refactor material import

scurest:
With the Princpled v2 node, most of the recent KHR material properties can be directly inserted into a Blender socket. This takes the form of either a color factor times a texture color

Or a scalar times one channel of a texture

However there is no function that creates these patterns. Instead this is currently implemented by mass copy/paste/modify.

This PR adds these missing functions in a new material_utils module (first commit) and refactors the code to use them (remaining commits).

Fixes the following bugs I noticed

* specularTexture was not multipled by 0.5 when specularFactor was exactly 1
* misspelled/inconsistent node labels (eg for sheen)
* transmission nodes were placed overlapping other nodes (?)

Here is a test file I used that should exercise all the code paths this touches: [MaterialTest.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/13895216/MaterialTest.zip)


julienduroure:
Thanks,
I will have a look soon

julienduroure (approved):

PR #2123, opened 2024-01-24
Find unique structured array elements more efficiently

Mysteryem:
np.unique works by sorting its input array and structured arrays are
sorted by sorting by each column sequentially with a stable sorting
algorithm. This means that every extra field in the structured array's
dtype adds an extra sort just to find unique elements.

When exporting shape key normals is enabled, 3 extra fields are added to
the `dots` array's structured dtype per shape key on the mesh, which can
quickly end up in over 100 sorts when finding the unique values.

By instead viewing a structured array as a non-structured dtype with the
same itemsize, unique values can be found with a single sort, but at the
cost of the result no longer being sorted according to the structured
dtype.

This non-structured dtype also has to be compared by its bytes, so -0.0
in any floating-point fields must be replaced with 0.0 beforehand and
any NaN values with different byte representations will be considered
unique from other NaN values.

For large meshes and/or meshes with many shape keys, this can reduce
export times by a few seconds.

---

This is heavily based on the similar [`io_scene_fbx.fbx_utils.fast_first_axis_unique`](https://projects.blender.org/blender/blender-addons/src/commit/a99528adb07d0c745b4f1ec5b1997e304b4dc048/io_scene_fbx/fbx_utils.py#L514) function I made for the FBX I/O add-on because finding unique elements along the first axis of a multidimensional array with np.unique internally just views each row and any extra dimensions as a structured dtype and then sorts that structured array view of the multidimensional array.

On a larger humanoid rigged model I have with a few hundred shape keys per mesh, this patch brings the export duration down from 40s to 31s (before this patch, it exports in 6s if I disable exporting shape key normals).

If desired, I can add explicit checks that the fields are not overlapping, though I think this is unlikely to ever be an issue.

If desired, I can add support for `bool` fields, though they don't appear to be used currently. For completeness, `bool` fields should set all values equal to `True` to `True`/`1` because its possible that `bool` fields could be constructed from a view of non-bool data. E.g. `np.arange(3, dtype=np.uint8).view(bool)` results in a `[False, True, True]` array, but the actual values will be `[0, 1, 2]`.

### DIFF
--- a/addons/io_scene_gltf2/blender/com/gltf2_blender_conversion.py
+++ b/addons/io_scene_gltf2/blender/com/gltf2_blender_conversion.py
@@ -163,3 +163,67 @@ def get_anisotropy_rotation_gltf_to_blender(rotation):
 def get_anisotropy_rotation_blender_to_gltf(rotation):
     # glTF rotation is in randian, Blender in 0 to 1
     return rotation * (2 * np.pi)
+
+def fast_structured_np_unique(arr, *args, **kwargs):
+    """
+    np.unique optimized for structured arrays when a sorted result is not required.
+
+    np.unique works through sorting, but sorting a structured array requires as many sorts as there are fields in the
+    structured dtype.
+
+    By viewing the array as a single non-structured dtype that sorts according to its bytes, unique elements can be
+    found with a single sort. Since the values are viewed as a different type to their original, this means that the
+    returned array of unique values may not be sorted according to their original type.
+
+    Float field caveats:
+    All elements of -0.0 in the input array will be replaced with 0.0 to ensure that both values are collapsed into one.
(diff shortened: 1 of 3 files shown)


julienduroure:
Thanks !
I will have a look soon

Mysteryem:
I missed a use of `np.unique` in `gltf2_blender_mesh.merge_duplicate_verts()`, so I have updated it to also use the new `fast_structured_np_unique` function.
I moved `fast_structured_np_unique` into `gltf2_blender_conversion` because it is no longer only used by `gltf_blender_gather_primitives_extract`.

julienduroure (approved):

julienduroure:
LGTM.
Maybe conversion file is not the best place for the function, as this is not a conversion.
Maybe we can create a _utils.py file?

julienduroure:
Merged! Thanks a lot

Issue #2046, opened 2023-11-06
Imported images should be set to "channel packed" alpha


hybridherbst:
**Describe the bug**
Seems "straight" alpha in Blender does something different than it should, the mode that should be used instead is "channel packed".

**To Reproduce**
1. Download 
[ChannelPackedImage.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/13267186/ChannelPackedImage.zip)
2. Import into Blender
3. Enable viewport shading
4. See that it's black (incorrect)
5. In Shading, switch image alpha to "channel packed"
6. Cat's back

**Expected behavior**

**Screenshots**
| After Import | Manually switching alpha to "channel packed" |
| - | - |
| <img width="627" alt="image" src="https://github.com/KhronosGroup/glTF-Blender-IO/assets/2693840/d92b4e3a-8773-445c-91ed-ba1d453f5dfb"> | <img width="860" alt="image" src="https://github.com/KhronosGroup/glTF-Blender-IO/assets/2693840/a90c2c0f-1f7d-46d7-84b5-e0db6cb1fec5"> |

**.blend file/ .gltf**
[ChannelPackedImage.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/13267186/ChannelPackedImage.zip)

julienduroure:
Confirm this behavior. 
I will check with Blender developer why and best practices about this option

scurest:
Cf https://blenderartists.org/t/textures-and-alpha-modes-oh-my/1161715.

AFAICT, when testing with a PNG file from disk...

* "Straight" means when used in a shader node...
    * In Eevee, the color socket emits premultipled color if the alpha socket is not connected (indirectly) to the output, and non-premultipled color if it is. (I had no idea this could happen. I reeeeeeally hate the fact that the value of a socket can be affected by the link topology.)
    * In Cycles, the color socket always emits premultiplied color.
* "Channel Packed" means when used in a shader node, the color socket always emits non-premultipled color.

I think hybridherbst is right, we should switch to "Channel Packed" for everything.

scurest:
```diff
diff --git a/addons/io_scene_gltf2/blender/imp/gltf2_blender_image.py b/addons/io_scene_gltf2/blender/imp/gltf2_blender_image.py
index de42061d..0be9d2ac 100644
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_image.py
[...]

PR #2127, opened 2024-01-28
Import: update pbrSpecularGlossiness for Principled v2

### LINKED ISSUE #1853: Imported Nodes for SpecGlossVsMetalRough sample model are not all using Principled BSDF + Export is broken
**Describe the bug**
Imported Nodes for SpecGlossVsMetalRough sample model are not all using Principled BSDF and the export/reimport of the model via glb/gltf  is broken as the spec/gloss water bottle imports as a black Principled BSDF

**To Reproduce**
1. Import the glb from here: https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/SpecGlossVsMetalRough
2. See the nodes on the materials
3. Export them and reimport them to see the specular gloss 

**Expected behavior**
Imported SpecularGloss model should use Principled BSDF node like other GLTF models, exported/reimported model shouldn't be a plain black material

**Screenshots**
Nodes_Metal1
Nodes_SpecG1

The exported from Blender and reimported model which is black:
Imported

**.blend file/ .gltf**
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/SpecGlossVsMetalRough

Version
OS: Windows 11
Blender Version 3.4.1

**Additional context**
Could be useful to have a Blender file including all sample models or an automated test to verify the import and exports of the sample models are 1:1

### LINKED ISSUE #1857: Convert SpecGloass to MetalRough at import?
Would it make sense to perform a crude conversion of SpecGloss to MetalRough and import that instead? 
The tradeoff would be it wouldn't look as accurate inside Blender, but it would export the same way it imported, and we could get rid of all the poorly exercised code in gltf2_blender_KHR_materials_pbrSpecularGlossiness.py.

For example

```py
# Preserves only the diffuse and a constant glossiness
def specgloss_to_metalrough(specgloss):
    metalrough = {}

    metalrough["baseColorFactor"] = specgloss.get("diffuseFactor")
    metalrough["baseColorTexture"] = specgloss.get("diffuseTexture")

    if "specularGlossinessTexture" not in specgloss:
        glossiness = specgloss.get("glossinessFactor")
        metalrough["roughnessFactor"] = 1 - glossiness
    
    metalrough["extras"] = specgloss.get("extras")
    metalrough["extensions"] = specgloss.get("extensions")

    return metalrough
```

new

_Originally posted by @scurest in https://github.com/KhronosGroup/glTF-Blender-IO/issues/1853#issuecomment-1416127718_

scurest:
Rewrites pbrSpecularGlossiness code to use the new Principled v2 node.

Closes #1853.  
Closes #1857.

Implementation Notes:

* Uses the trick of setting IOR=1000 to make the F0 color equal the specular tint.
* The biggest problem is the glossiness, which is 1-roughness. This patch makes a copy of the specularGlossinessTexture and bakes the roughness into its alpha channel. I would have just used a Math node to do 1-X, but that wouldn't export.
* A `calc_locs` was added to calculate locations for the node blocks. It uses the socket order from the new Principled node.

Import of SpecGlossVsMetalRough from the sample models:

SpecGlossVsMetalRough

You can see the new shader is generally too dark. Some edge reflections are better though. I don't know if there's a way to improve this.

Exporting works (the visual appearance roundtrips). When exporting, the material that used KHR_materials_pbrSpecularGlossiness will use KHR_materials_ior and KHR_materials_specular.

Here is what the shader nodes look like:

Full shader graph for specGloss

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/gltf2_blender_KHR_materials_pbrSpecularGlossiness.py
+++ b/addons/io_scene_gltf2/blender/imp/gltf2_blender_KHR_materials_pbrSpecularGlossiness.py
@@ -12,159 +12,202 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import bpy
 from ...io.com.gltf2_io import TextureInfo
 from .gltf2_blender_pbrMetallicRoughness import \
-    base_color, emission, normal, occlusion, make_output_nodes, make_settings_node
-from .gltf2_blender_texture import texture
+    base_color, emission, normal, occlusion, make_settings_node
+from .gltf2_blender_material_utils import color_factor_and_texture
+from .gltf2_blender_texture import texture, get_source
+from .gltf2_blender_image import BlenderImage
+import numpy as np
 
 
 def pbr_specular_glossiness(mh):
     """Creates node tree for pbrSpecularGlossiness materials."""
-    # This does option #1 from
-    # https://github.com/KhronosGroup/glTF-Blender-IO/issues/303
+    ext = mh.get_ext('KHR_materials_pbrSpecularGlossiness', {})
 
-    # Sum a Glossy and Diffuse Shader
-    glossy_node = mh.node_tree.nodes.new('ShaderNodeBsdfGlossy')
-    diffuse_node = mh.node_tree.nodes.new('ShaderNodeBsdfDiffuse')
-    add_node = mh.node_tree.nodes.new('ShaderNodeAddShader')
-    glossy_node.location = 10, 220
-    diffuse_node.location = 10, 0
-    add_node.location = 230, 100
-    mh.node_tree.links.new(add_node.inputs[0], glossy_node.outputs[0])
-    mh.node_tree.links.new(add_node.inputs[1], diffuse_node.outputs[0])
+    pbr_node = mh.nodes.new('ShaderNodeBsdfPrincipled')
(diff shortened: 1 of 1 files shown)


scurest:
I found out that changing the Specular Tint texture to "Channel Packed" Alpha fixes the too-dark rendering.



See #2046. Presumably this problem would also affect the other extensions that use packed RGB+A textures (KHR_materials_specular and KHR_materials_sheen).

Please let me know whether #2046 will be a blocker for this.

julienduroure:
Thanks.
I will have a look soon.
@donmccurdy @emackey can you also have a look?

emackey:
This seems great to me, thanks @scurest!

node graph screenshot

One thing I noticed, it seems to make a copy of an image, see above.  But this might be the fault of the strange model we've got here, two different shading models on the two bottles sharing the same occlusion map.  Blender probably can't easily figure out what we've done here and how best to optimize it, so that's OK.

In general though this means that Blender can automatically upgrade the legacy specGloss format to the new one, which is a fantastic capability to have!

scurest:
> One thing I noticed, it seems to make a copy of an image, see above.

Yes, that's how it works. The original specular-glossiness texture is used for the Specular Tint socket. For the Roughness socket, a copy of the texture is made and the the roughness is baked into the alpha channel.

emackey (approved):

julienduroure:
Merged! Thanks a lot

PR #2128, opened 2024-01-29
Import: decouple unlit from make_output_nodes

scurest:
Refactors the unlit importer to make its own node setup instead of using `make_output_nodes`.

Should have no functional change on the shader nodes, except only one Transparency node gets created, instead of two. 

-----

A historical note on `make_output_nodes`: This function existed to abstract over certain commonalities between the shader node setups for Blender 2.79, pbrSpecularGlossiness, and unlit. It has since become a grab-bag for output node setups with every callsite using disjoint subsets of its features.

2.79 is long gone, and pbrSpecularGlossiness uses a normal Principled node as of #2127. So this moves the alpha node features of `make_output_nodes` to the only place they are now used, the unlit code.

If both this and #2127 get merged, `make_output_nodes` can be refactored to serve only pbrMetallicRoughness.


julienduroure (approved):

julienduroure:
Thanks!
Seems we can now refactor _make_output_nodes_

PR #2129, opened 2024-01-29
Import: adjust positions of Mix Alpha nodes

scurest:
Very simple patch. Adjusts the position of Mix Alpha nodes in the Base Color node block to accommodate the taller Mix Color nodes in recent Blender versions.

Before (Mix nodes overlap):

After:


julienduroure:
Thanks!

julienduroure:
Merged! Thanks a lot

PR #2132, opened 2024-02-03
Import: remove make_output_nodes

scurest:
Follow up to #2128.

Most of `make_output_nodes` is now unused. Remove it, moving its sole responsibility, creation of the volume node, into the main metal-roughness shader creation function.

Should be a pure refactor with no change in behavior.


julienduroure:
Thanks!

PR #2146, opened 2024-02-24
Export: add NodeNav tool for navigating node trees

scurest:
Adds a new tool to the exporter for searching node trees called a NodeNav (name is up for discussion).

A NodeNav generalizes and unifies the current NodeSocket and ShNode classes. It holds

* a current node
* the output socket through which we arrived at the node
* the input socket where we're going to go next
* the stack of Group nodes we descended through to reach here

You can use it to move backwards through the node tree, like previous_node and previous_socket.

```py
nav.move_back('Base Color')         # moves backwards through the Base Color input
if nav.moved: ...                   # check whether the last move succeeded or not
nav2 = nav.peek_back('Base Color')  # like move_back, but returns a new NodeNav
nav.get_constant('Base Color')      # gets a constant like get_const_from_socket
```

and it handles tracking descent for you so you don't have to manage the group_path manually

```py
# Old
get_const_from_socket(NodeSocket(node.node.inputs[6], node.group_path), kind='RGB')
# New
nav.get_constant(6)
```

NodeNav is intended to eventually be able to replace all the tools in search_node_tree.py, but for now it coexists with them. Only a few places are converted to use it in this PR. This works by converting NodeSockets to NodeNavs.

There are two other nice features to note

1. You do not have to pass eg `kind='VALUE'` to get_constant, it detects this automatically based on the socket's type.
2. You can select sockets by *identifier* instead of name by prefixing the name with a `#`. This lets you select sockets with non-unique names that currently require an index.

    ```py
    # Eg. to select the A input to a Color Mix node
    nav.get_constant(6)           # select nav.node.inputs[6], poor readability
    nav.get_constant('A')         # doesn't work, there are multiple sockets named 'A'
    nav.get_constant('#A_Color')  # works!
    ```

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/material/gltf2_blender_gather_texture_info.py
+++ b/addons/io_scene_gltf2/blender/exp/material/gltf2_blender_gather_texture_info.py
@@ -173,12 +173,13 @@ def __gather_normal_scale(primary_socket, export_settings):
 def __gather_occlusion_strength(primary_socket, export_settings):
     # Look for a MixRGB node that mixes with pure white in front of
     # primary_socket. The mix factor gives the occlusion strength.
-    node = previous_node(primary_socket)
-    if node and node.node.type == 'MIX' and node.node.blend_type == 'MIX':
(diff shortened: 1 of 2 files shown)


julienduroure:
Thanks!

Issue #2111, opened 2024-01-12
Future of alphaMode


scurest:
In Blender's legacy EEVEE renderer, a material has a "Blend Mode" property that controls the interpretation of the alpha value produced from the node graph. For example, the nodes could calculate Alpha=0.5, but if the Blend Mode was "Opaque", the alpha value would be ignored and it would be rendered opaque. Both the importer and exporter have up to this point leaned on this feature to represent glTF's alphaMode+alphaCutoff properties.

However, this has only ever worked in EEVEE. In Cycles, alpha always has the normal interpretation as semi-transparency/coverage. If you want a surface to be opaque, the node graph should calculate Alpha=1, and not rely on an "after the fact" reinterpretation of the alpha value.

With Blender 4.1's introduction of the EEVEE-Next renderer with no Blend Mode, intended to eventually replace current (legacy) EEVEE, this addon should probably start thinking about moving to encoding alphaMode entirely in the node graph. 

That would mean

|alphaMode|Blender Implementation|
|---|---|
|OPAQUE|1.0→Alpha Socket|
|MASK|glTF alpha→[Math node: X >= alphaCutoff]→Alpha socket|
|BLEND|glTF alpha→Alpha socket|

**Import**
* OPAQUE and BLEND should already be correct.
* For MASK, the base color code would learn how to insert a Greater-Than node in front of the Alpha socket (same as it worked in 2.79).
* It can continue setting the Eevee Blend Mode for now.

**Export**
* The alphaMode+alphaCutoff would need to be gathered from the Alpha socket. Constant 1 = OPAQUE, Greater-Than node = MASK, otherwise = BLEND.
* Backwards compatibility considerations with current Blend Mode based setups?

Thoughts?

julienduroure:
Note: EEVEE-next is postponed to 4.2

scurest:
@julienduroure Did you decide if this idea sounds good to you?

scurest:
Okay, the next problem is how to implement MASK mode. Spec says

> 5.19.10. material.alphaCutoff
>
> Specifies the cutoff threshold when in MASK alpha mode. If the alpha value is greater than or equal to this value then it is rendered as fully opaque, otherwise, it is rendered as fully transparent. A value greater than 1.0 will render the entire material as fully transparent.

ie. 

```
alpha = alpha >= cutoff ? 1.0 : 0.0;
```

[...]

PR #2147, opened 2024-02-25
[WIP] Import: use nodes for Alpha clipping

scurest:
Does #2111 for the importer.  
Draft until PRs for the exporter and manual are also done.

Since there wasn't any feedback about the preferred node setup, just uses 1 - (alpha < cutoff) for now. A node frame is used to aid readability.

Example using [AlphaBlendModeTest](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/AlphaBlendModeTest):


julienduroure:
Thanks,
I will have a look soon.
- Do you plan to work on export / doc / test PR?
- About the solution used: There is no perfect solution (until Blender have a >= ), but this one is the more comprehensive solution that give the right answer

scurest:
There's no rush since there's no guarantee Eevee Next will make 4.2 after all.

> Do you plan to work on export / doc / test PR?

Yes. I need #2146 to do the export, since I found the current tools for searching nodes too hard to use.

PR #2157, opened 2024-03-12
Import: add drag & drop support

### LINKED ISSUE #2144: Manage drag & drop
Cf https://docs.blender.org/api/master/bpy.types.FileHandler.html

scurest:
Fixes #2144.

Based on https://docs.blender.org/api/master/bpy.types.FileHandler.html. In particular, to match the example there, the filepath property was moved from ImportHelper because the one in ImportHelper does not have `options={'SKIP_SAVE'}`, and a directory property was added.

Tested with

* import from menu (single file, multiple files)
* import from drag and drop (single file, multiple files)
* import from script (single file)

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -1965,7 +1965,7 @@ def menu_func_export(self, context):
     self.layout.operator(ExportGLTF2.bl_idname, text='glTF 2.0 (.glb/.gltf)')
 
 
-class ImportGLTF2(Operator, ConvertGLTF2_Base, ImportHelper):
+class ImportGLTF2(Operator, ConvertGLTF2_Base):
     """Load a glTF 2.0 file"""
     bl_idname = 'import_scene.gltf'
     bl_label = 'Import glTF 2.0'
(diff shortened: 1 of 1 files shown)


julienduroure:
As discussed in #2144, the Blender API is not fully ready yet for py addons drag&drop, as there is no panel display to choose import preferences before the import.
This enhancement is planned (hopefully) for 4.2 (last time I discussed it with devs).

So, do not expect any review / merge before we know if this will land into 4.2.

If yet => Great, we will merge it (not sure if this will need changes in addon)
If not => We will have to decide if let the user import without being able to change import options

scurest:
The importer doens't use panels, only the exporter does. When you drop, you already get the options in the corner, just like when you eg create a Cube

[](https://github.com/KhronosGroup/glTF-Blender-IO/assets/11024420/f069eb63-34ef-41b6-af7a-6a311b0ed085)

The C++ importers are different, they show a popup where you can change options *before* you do the import. It looks like this.

popup

If you don't want that, it's easy to just show the normal fileselector with the dropped file preselected when a file is dropped by deleting these lines. Does not preselected all files when multiple files are dropped though.

https://github.com/KhronosGroup/glTF-Blender-IO/blob/dbd0103364803f74e7b71c97cde74d193689750a/addons/io_scene_gltf2/__init__.py#L2090-L2091

I don't see a reason to wait on doing _some_ kind of drag&crop.

julienduroure:
> you already get the options in the corner, just like when you eg create a Cube

This panel is changing option **after** the import operator. It means: undoes the import, and does a new import, every time you change an option. I am not sure this is what is best (especially for big imports)

4.1 freeze is tomorrow, I don't want to rush it, and will land in 4.2.

scurest:
Okay, nevermind.

julienduroure:
What did you close it? 
Let's keep it open, this is a good start for 4.2

PR #2181, opened 2024-03-27
Fix #2102 fix hidden object matrices

### LINKED ISSUE #2102: Artifacts when Exporting Scenes with Objects Hidden from Viewport
**Describe the bug**
Hello!

I've noticed that attempting to export scenes which contains objects that have
`hide_viewport == True` yields some strange looking artifacts when re-imported
(see screenshots below).

I'm currently working around this by having a small export script that
essentially does this:

    for obj in bpy.data.objects:
        if obj.hide_viewport:
            obj.hide_viewport = False
    bpy.ops.export_scene.gltf(filepath=f'{output}.gltf',
                              export_cameras=True,
                              export_lights=True,
                              export_format='GLTF_SEPARATE')

This works, but for large scenes (e.g., San Miguel from issue #2101) this can take
a full day to complete.

**To Reproduce**
1. Open included "arcsphere-50.blend" file.
2. Export the scene to glb or glTF.
3. Open a new instance of Blender
4. Import the previously exported scene.

**Expected behavior**
The exported scene should match the "arsphere_expected.glb" file.
(i.e., all objects, including the ones marked with `hide_viewport = True`,
should be included, and at their correct position).

**Screenshots**
Scene with the artifact, presumably from the "hidden" instances:
blender-artifact
Scene exported with no objects with `hide_viewport == True`.
blender-no-artifact

The zipped file includes a .blend file with half of all objects/instances "hidden" with `hide_viewport = True`,
as well as a zip with the expected export result and the received result.
[arcsphere-50.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/13862843/arcsphere-50.zip)

**Version**
 - OS: Linux
 - Blender Versions 3.6, 4.0

julienduroure:


julienduroure:
Note sure if this should be activated by default,
as this take longer to exporter (even if now, result is OK)

julienduroure:
Note: 
- fbx doesn't handle it correctly neither
- usd has "only visible" enable by default (not glTF nor fbx). But when disable this option, seems to export correctly

julienduroure:
Removed from 4.2 milestone, to have more time to find / decide a solution

PR #2193, opened 2024-04-05
Use layout panels instead of PanelTypes

jessey-git:
Convert GLTF to use layout panels instead of PanelTypes for import and
export property drawing. This should match the existing layout used by
in file browser today.

This allows viewport/outliner drag-n-drop support to work fully[1] and
it enables the soon to be committed use of Collection exporters[2].

[1] After a drag-n-drop on the 3d Viewport or the Outliner, this Import dialog will be presented:
gltf-panels-0

[2] For Collection Exporters, this UI will be used (will require a small follow-up PR to fully enable once Collection Exporters are committed):
gltf-panels-1

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -15,8 +15,8 @@
 bl_info = {
     'name': 'glTF 2.0 format',
     'author': 'Julien Duroure, Scurest, Norbert Nopper, Urs Hanselmann, Moritz Becher, Benjamin Schmithüsen, Jim Eckerlein, and many external contributors',
-    "version": (4, 2, 13),
-    'blender': (4, 1, 0),
+    "version": (4, 2, 14),
+    'blender': (4, 2, 0),
     'location': 'File > Import-Export',
     'description': 'Import-Export as glTF 2.0',
     'warning': '',
@@ -61,7 +61,7 @@ from bpy.props import (StringProperty,
                        FloatProperty,
                        CollectionProperty)
 from bpy.types import Operator
-from bpy_extras.io_utils import ImportHelper, ExportHelper
+from bpy_extras.io_utils import ImportHelper, ExportHelper, poll_file_object_drop
 
 
 #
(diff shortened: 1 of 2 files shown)


julienduroure:
Thanks Jesse!

julienduroure:
Hello @jessey-git 

Seems we broke the user extension system with this new layout panel system.
User extension system is a way, in glTF, to add a new addon that can enhancement / change the glTF exporter.
This system allowed to register some classes that change the behavior, but also some Panel classes that add some panels in glTF exporter or importer, by setting a Panel class with bl_parent_id
This system is no more working. How can now access the operator layout to add a new layout panel ?
Here is an example of addons, with the old system : https://github.com/KhronosGroup/glTF-Blender-IO/blob/main/example-addons/example_gltf_exporter_extension/__init__.py

jessey-git:
@julienduroure 
Hmm, I didn't know something like this was being used here. Supporting this is not directly possible right now.
Something custom could be made to work, but it would be a breaking change for the ecosystem of folks using this mechanism.

If that is too much for the upcoming Blender release the only choice will be to revert the Panel and File-Handler changes. That would mean disabling Drag-and-drop Import and collection Export for glTF but at least there's no regression in capability.

One possible custom solution / design that I think could work (if I'm understanding all that's involved):
- Instead of your users registering an extension PanelType, they will register a single callback function (with some new API that you provide)
- Your API would keep a list of these callbacks and invoke them, as needed, passing in the appropriate layout/context for drawing
- The callback function would draw their properties like normal using those passed in layout/context parameters
- An additional change is that Exporters need to store their properties differently. With collection export, more than 1 operator may be active at once. Your users would need the ability to store multiple `ExampleExtensionProperties` structures on the scene and access them accordingly

Pseudo code:
```
# For example, at the bottom of the 'ExportGLTF2_Base.draw' function the following would be added:
for user_extension in export_extension_callbacks:
    user_extension(context, layout)
```

julienduroure:
Thanks for your answer.
I will go with your custom proposition, even if this will need some update for any user using this mechanism.
I do not plan to rollback your work.

I will come back to you if I have more question or remarks.
Thanks for your suggested design, I will try to implement it

PR #2203, opened 2024-04-16
Enable the Collection export feature

jessey-git:
This enables GLTF to be used as a Collection exporter [0].

To enable this, two general changes were made:
- The `export_scene.gltf` operator was configured on the File Handler
- A `collection` string property was added to the export Operator

This `collection` property is akin to the special `filepath` and
`directory` properties in that this is how Blender will communicate with
the addon. The string is simply the name of the collection to be used.

The layout is slightly changed when using Collection export: the various
"Limit To" options are not drawn/used. Feedback was received about this
during development, with our other exporters, where the users found them
confusing when the exporter was set on a collection. Basically, if a
collection exporter is being used, we assume the user really wants
everything regardless of selection or renderability etc.

[0] To test, go to `Collection Properties` -> `Exporters` -> `Add` and
select "glTF 2.0" after applying this patch.

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -558,6 +558,12 @@ class ExportGLTF2_Base(ConvertGLTF2_Base):
         default=False
     )
 
+    collection: StringProperty(
+        name="Source Collection",
+        description="Export only objects from this collection (and its children)",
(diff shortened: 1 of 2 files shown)


julienduroure:
Thanks !

PR #2227, opened 2024-05-14
gather animation hook for all animations

y-khodja:


julienduroure:
Hello,

There is already an `gather_animation_hook` hook in the code, you can not use the same name.
Root issue is that there are hooks for sampled export, but not for fcurves/not sampled ones. Not sure adding a new hook that will be performed for both cases is the best/right solution, as there will be 2 hooks for sampled, and only 1for not sampled. 
Maybe it can be better to add proper hooks for not sampled export.
(And you can not say this is a hook for all animations, as your PR handle only "action" export mode, not NLA or Scene mode)

addons/io_scene_gltf2/blender/exp/animation/gltf2_blender_gather_action.py
```diff
@@ -426,6 +426,9 @@ def gather_action_animations(  obj_uuid: int,
             link_samplers(animation, export_settings)
             animations.append(animation)
 
+            # export user extensions for all types of animation
+            export_user_extensions('gather_animation_hook', export_settings, animation, blender_action, blender_object)
```
julienduroure, line 430: There is already a hook with the same name

julienduroure (changes requested):

y-khodja:
Hello,

Thank you for your answer, i now see the issue. I thought it would be better to have one hook that take in all animation export types as it would be used to add extensions or extras to the exported animation its self.

addons/io_scene_gltf2/blender/exp/animation/gltf2_blender_gather_action.py
y-khodja: Yes this was a suggestion to encounter exported animations extension problem

julienduroure:
2 years without any news on this PR, closing.
Feel free to open a new PR if you want

PR #2233, opened 2024-05-16
Option to export VC even if not used in node tree

julienduroure:
History of this PR: 

>  In the common usage / historically / simplicity of usage, Vertex Colors are used not only for Vertex Color, but also for transmission of other custom data.
    The glTF Specification propose a way to transmit custom attributes, but seem for a number of reasons, this workflow is not commonly used:
        Not very well know
        Not perfect from Blender side yet (need to use name starting with underscore to avoid exporting internally used custom attributes)
        Lack of implementation / easy way to retrieve data for some engine/viewer
        This make the end to end workflow of custom attributes not very used (from what I know)

> All of these make that VC are commonly used for custom attributes transmission too, even if this goes against the glTF Specification, which says that VC must act as a multiplier of Base Color.

> Setting some custom attributes in VC COLOR_0 will generate some valid glTF files, but a fully compliant glTF 2.0 viewer must use these data as a base color multiplier, probably resulting in weird shading.

Even if we truly think we need to put energy setting a global solution around custom attributes (When we say globally, it means from Blender, Unity, Godot, UE, threejs, Babylonjs, etc...), we also know that you have some pipeline and workflow already in production right now, and that the ideal solution will take some time to be implemented in all these softwares.

So here is our proposition:

This PR add a new panel with more options to choose at export:

image

This PR adds an option to export active VC as COLOR_0, even if not use in node tree as a multiplier of base color.
The generated glTF can look wrong in viewer that are fully compliant with glTF, as the VC will be a multiplier of base color, where it is not in Blender viewport.

The default value is still "MATERIAL", as this is the option fully compliant with glTF Specification
image

When choosing ACTIVE, a warning will explaining that your workflow can be more robust with another solution:

image

You can also choose to never export VC:

image

You can now choose to export additional VC as COLOR_1, COLOR_2, etc...

image

This also add an option to choose what to do when there is no material. You can now choose to export active VC as COLOR_0, or not export any VC.

image


bertodelrio256:
Im not a dev but this looks like the optimal solution for everyone. I would say the "export all vertex colors" should include the alpha channel of the VC as well. thank you! Im sure game engines will eventually switch to using custom attributes instead for vertex colors, but who knows how long that will be. (unreal might already but still supports VC)

digimbyte:
I did make this suggestion way back when this whole thing started. glad to see someone with more skill and common sense step up.

Takanu:
I think you've covered all the bases and the extra explanation for the feature is important and helpful, thanks!

TheSofox:
Which set of options would most closely mirror `export_colors` from previous versions?

We're having a bug with the Godot engine where Vertex colours weren't being exported when not attached to a material: https://github.com/godotengine/godot/issues/97953

I seem to have fixed it by changing `export_vertex_color` from `MATERIAL` to `ACTIVE` (for Blender 4.2+ obviously), but I'd like some confirmation that this is the right approach.

PR #2286, opened 2024-07-19
Fix crash when exporting animations with Full Collection Hierarchy option enabled.

sumihiro:
Hi,

This pull request addresses an issue where the exporter crashes when attempting to export animations with the Full Collection Hierarchy option enabled.

Changes made:
- Modified the animation caching logic to properly handle collections in the hierarchy
- Ensured that collection objects are skipped in the object_caching function

This fix allows users to successfully export animations while using the Full Collection Hierarchy option without experiencing crashes.

Please review and let me know if any further changes or information are needed.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/animation/gltf2_blender_gather_action.py
+++ b/addons/io_scene_gltf2/blender/exp/animation/gltf2_blender_gather_action.py
@@ -519,7 +519,7 @@ def __get_blender_actions(obj_uuid: str,
     if export_settings['gltf_animation_mode'] == "BROADCAST":
         return __get_blender_actions_broadcast(obj_uuid, export_settings)
 
-    if blender_object and blender_object.animation_data is not None:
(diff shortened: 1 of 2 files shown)


julienduroure:
Can you please confirm that you tested on 4.2, because this bug should already been solved between 4.1 and 4.2
Also, please provide a test .blend file that crash before your PR and not after your PR

sumihiro:
I tested this on both 4.2 and the latest main branch, but it still crashed.

I'm attaching the blend file that causes the crash.

You can reproduce the crash by checking both 'Scene Graph > Full Collection Hierarchy' and 'Animation' options when exporting."

[car.blend.zip](https://github.com/user-attachments/files/16309694/car.blend.zip)

julienduroure:
Hello,
I can confirm the crash.
I will go with my own implementation:
- Early checks, avoiding checking hasattr for each object
- Checks style more aligned with other part of the code ( checking vtree type instead of real Blender object type)
- 
See #2287

Issue #2267, opened 2024-06-06
log level setting in export


ronh991:
**Is your feature request related to a problem? Please describe.**
There is a log level property in the ImportGLTF2 class, but I do not see one in the ExportGLTF2.

I am trying to stop the flooding of INFO messages. so my blender start 

F:\blender-4.2.0\blender.exe --start-console --debug-value 1

WARNING

**Describe the solution you'd like**
a similar way to get the app debug value with set_debug_log for the export classes.  The ExportGLTF2_Base already has

        export_settings['loglevel'] = logging.INFO

however set to INFO and can't be changed like import?

**Describe alternatives you've considered**
mod the _init_.py

**Additional context**

PR #2299, opened 2024-07-26
Add logging level control for gltf exports

buildxyz-git:
**Issue**: Exporting scene as gltf when using python built as a module is very noisy with `INFO` logging entries.

I have made changes to `ExportGLTF2_Base` in:
`blender/scripts/addons_core/io_scene_gltf2/__init__.py` 

changes support setting the log level instead of using the hardcoded `INFO` value  [here](https://github.com/blender/blender/blob/0d70481d15da54560f70f8de8d26c973585aa3e7/scripts/addons_core/io_scene_gltf2/__init__.py#L1054) which produces way to much output when exporting.

`ImportGLTF2` already has a pattern for this [here](https://github.com/blender/blender/blob/0d70481d15da54560f70f8de8d26c973585aa3e7/scripts/addons_core/io_scene_gltf2/__init__.py#L1866) so I created a global `set_debug_level` function they can both use in the same manner.

The user can now set the log level before exporting by setting `bpy.app.debug_value` like:

```python
bpy.app.debug_value = 2
bpy.ops.export_scene.gltf(...)
```

This will solve [issue 2267](https://github.com/KhronosGroup/glTF-Blender-IO/issues/2267)

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -170,6 +170,18 @@ def is_draco_available():
 
     return is_draco_available.draco_exists
 
+def set_debug_log():
+    import logging
+    if bpy.app.debug_value == 0:      # Default values => Display all messages except debug ones
+        return logging.INFO
+    elif bpy.app.debug_value == 1:
+        return logging.WARNING
+    elif bpy.app.debug_value == 2:
+        return logging.ERROR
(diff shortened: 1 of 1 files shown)


julienduroure:
Thanks for the PR
I will have a look soon

julienduroure:
Merged, thanks!

PR #2304, opened 2024-08-06
Account for library when gathering image URI

etacarinaea:
When exporting a scene that has linked objects with materials that have relative paths to images, the absolute path of those images was being determined by taking that path relative to the current scene's path, instead of relative to the original path of the library, which would result in the images not being found if the current scene and the library weren't in the same directory. This PR fixes that.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/material/image.py
+++ b/addons/io_scene_gltf2/blender/exp/material/image.py
@@ -56,7 +56,11 @@ def gather_image(
         uri, factor_uri = __gather_uri(image_data, mime_type, name, export_settings)
     else:
         # Retrieve URI relative to exported glTF files
-        uri = __gather_original_uri(image_data.original.filepath, export_settings)
+        uri = __gather_original_uri(
+            image_data.original.filepath,
+            blender_shader_sockets[0].socket.id_data.library,
+            export_settings
+        )
         # In case we can't retrieve image (for example packed images, with original moved)
         # We don't create invalid image without uri
         factor_uri = None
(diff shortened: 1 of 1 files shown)


julienduroure:
Thanks for the PR
I will have a look soon

julienduroure:
Some files are now renamed in main branch.
Can you please rebase your PR?
Thanks :)

etacarinaea:
Here's the zip: [gltf-blender-io_library-uri_test.zip](https://github.com/user-attachments/files/16978115/gltf-blender-io_library-uri_test.zip)
```
.
├── assets
│   ├── asset.bin
│   ├── asset.blend
│   └── asset.gltf
├── img
│   └── test.png
├── test.bin
├── test.blend
└── test.gltf
```

I think I forgot to mention that this only seems to happen when "keep original" is checked while exporting as a gltf. The zip includes the resulting gltf files from exporting the original scene (asset.blend) and the scene with the object linked (test.blend). The gltf of the original one correctly includes the image path, the one with the linked object doesn't.

Output from `git diff -- test.gltf assets/asset.gltf`:
```diff
@@ -44,6 +47,19 @@
                        ]
                }
        ],
+       "textures":[
+               {
+                       "sampler":0,
+                       "source":0
+               }
+       ],
+       "images":[
+               {
+                       "mimeType":"image/png",
+                       "name":"test.png",
+                       "uri":"../img/test.png"
+               }
+       ],
        "accessors":[
                {
                        "bufferView":0,
```

julienduroure:
> when "keep original" is checked

It was the missing part! Thanks :)

julienduroure:
Merged, thanks!

Issue #2317, opened 2024-08-14
Duplicate Mesh with Collection Instances with Geometry Nodes Instances


slyedoc:
When using exporting With Geometry Nodes Instances, collection instances are duplicated.  A simple Cube collection instance is exporting as:

```
		{
			"mesh":0,
			"name":"CubeInInstance.001"
		},
		{
			"mesh":1,
			"name":"GN Instance"
		},
		{
			"children":[
				0,
				1
			],
			"name":"InstanceCube"
		}
```
The collection instance ended up with 2 children, each with its own mesh.

[Blend File - instance_collections.zip](https://github.com/user-attachments/files/16606282/instance_collections.zip)
Resulting Gltf
[instance.zip](https://github.com/user-attachments/files/16606313/instance.zip)

**To Reproduce**
1. Create Empty Scene, create collection called Cube, put cube mesh in it,
2. Create another Empty Scene, add collection instance of "Cube"
3. Export Scene with Geometry Nodes Instances enabled, (tired with many other settings to no effect), see 
![options](https://github.com/user-attachments/assets/39adcd2b-2e2c-471c-89fe-93557a1043a9)

**Version**
 - OS: [linux]
 - Blender Version [4.2]

Happens on main and #2311 as well

slyedoc:
fixed.

huge help, thanks.

PR #2318, opened 2024-08-14
fix_2317

julienduroure:
Make sure to not check instances inside instances collection when GN instance is enable.
This was exporting data twice


julienduroure:
Fix #2317

PR #2324, opened 2024-08-18
Fix #2321

### LINKED ISSUE #2321: blender 4.2 cannot export shapekey with amature
**Describe the bug**
 Cannot export shapekey with Amature

**To Reproduce**
Add a Cube
Set some shape keys
Add a Amature
Attach the Cube to Amature
Export to gltf
import the gltf file, the cube does not have any shape keys

**Expected behavior**
shapekey can be exported in 4.1.1 

**Screenshots**
the cube with shape key and amature
image
modifier
image
gltf export settings
image
import glft file
image

**Version**
 - OS: Win10 19045.4651
 - Blender Version 4.2

julienduroure:
Export SK when apply modifiers with only Armature modifier


Jontte (changes requested):
I believe we need a similar fix in `gltf2_blender_gather_mesh.py:120` to make sure the names of the shape keys also get exported in the mesh extras.

PR #2332, opened 2024-08-30
Note requirements for exact class names

Phrogz:
The example docs and code previously did not make it clear that the exact class names must be used for the add-on extension to function.

### DIFF
--- a/example-addons/example_gltf_exporter_extension/__init__.py
+++ b/example-addons/example_gltf_exporter_extension/__init__.py
@@ -56,6 +56,7 @@ def draw(context, layout):
         body.prop(props, 'float_property', text="Some float value")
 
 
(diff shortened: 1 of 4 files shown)


Phrogz:
These changes might seem a little pedantic; after all, the README for the exporter does already say "Add a class named …". However, the required class names looked to me like examples, and I spent 3.5 hours trying to figure out why my class (with a different name) wasn't being called, trying to find what I was missing to register my extension so it would be found. I never guessed convention over configuration would be in play. After all, the example files already have other classes like `ExampleExtensionProperties`.

So, I hope that these changes save some other developer some grief, by being abundantly clear.

julienduroure:
Merged, thanks!

PR #2340, opened 2024-09-02
Describe mesh_node_idx more clearly

Phrogz:
The previous comment made it unclear whether `mesh_node_idx` refers to the index in the `meshes` array or the `nodes` array.

Further, the use of the word "Indices" implies that multiple, different index values are present for a single VNode. As best as I can tell, two of the three attributes will always be `None`, with only one attribute getting the index in the "nodes" collection, depending on the type of object (camera, light, or "mesh" node).

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/vnode.py
+++ b/addons/io_scene_gltf2/blender/imp/vnode.py
@@ -60,7 +60,7 @@ class VNode:
         self.rotation_after = Quaternion((1, 0, 0, 0))
         self.rotation_before = Quaternion((1, 0, 0, 0))
 
-        # Indices of the glTF node where the mesh, etc. came from.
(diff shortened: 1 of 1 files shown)


scurest:
> Further, the use of the word "Indices" implies that multiple, different index values are present for a single VNode

There are, early in the vnode creation. It's only in [`fixup_multitype_nodes`](https://github.com/KhronosGroup/glTF-Blender-IO/blob/3a06fa0bd0d085affde102ea8be9818b92e88301/addons/io_scene_gltf2/blender/imp/vnode.py#L356) that the Blender-compatible constraint that at most of these is defined is enforced.

julienduroure:
Hello,
Closing this PR, as the assumption is not true.
The glTF specification make it possible to have a node with light, camera, etc... at the same time, where this is not possible in Blender

PR #2390, opened 2024-10-30
viewport material export

julienduroure:
New option to export simple material, based on viewport display information.

image

Exported values:
- BaseColor (RGBA)
- Metallic (float factor)
- Roughness (float factor)
- 
image

=> This is very fast, as we don't check the node tree.
=> Custom properties of the material can still be exported


julienduroure:
Note: We are close to 4.3 release, so we can't make this (new feature) into it.
This is planned for 4.4

HailToDodongo:
Thanks for looking into this so quickly!

I did gave the PR a test on the latest 4.3 beta, and it does seem to fix my issue.
Export works again, and the basic material + custom properties are still there for me to process.

The only thing i noticed is that certain older models have wrong vertex colors, but i can create a separate issue for that.
(As a test, .obj export has correct colors)

julienduroure:
Has impact on  #2405

Issue #2386, opened 2024-10-25
Feature Request: "Force keeping channel for data/properties" (exported via KHR_animation_pointer)


The0Dev:
**Is your feature request related to a problem? Please describe.**
Some animation channels get lost when reexporting a glTF file with Blender. 

**Describe the solution you'd like**
I'd like to have an option that would preserve animation channels of properties (exported via KHR_animation_pointer) even when keyframes are identical (similarly to existing options for preserving bone and object channels). 

**Describe alternatives you've considered**
It's possible to work around this by slightly changing keyframes but that's not ideal and I'd like to have a proper solution.

**Additional context**
I'm dealing with a game engine that can target some animation channels (with identical or non-identical keyframes) and additionally alter/animate them. It would be nice if I could always keep all channels.

PR #2400, opened 2024-11-12
Added the 'force keeping channel for data' feature

ghost:
Implements the feature request #2386

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,7 +6,9 @@
             "type": "python",
             "request": "attach",
             "port": 5678,
-            "host": "localhost"
+            "host": "localhost",
+            "stopOnEntry": true,
+            "justMyCode": false
         }
     ]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,9 +6,7 @@
             "type": "python",
             "request": "attach",
             "port": 5678,
-            "host": "localhost",
-            "stopOnEntry": true,
-            "justMyCode": false
+            "host": "localhost"
         }
     ]
 }
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -12,15 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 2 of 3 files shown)


julienduroure:
Hello,
Thanks, I will have a look soon.
Plan to merge it for Blender 4.4

Seems your format options is not aligned with the ones that we are using here, and on Blender repository.
Can you please fix it (avoiding to change the current code, and make it easy to see your diff).
For info, what is used can be found here: https://github.com/KhronosGroup/glTF-Blender-IO/blob/main/tools/autopep8.md

The0Dev:
Hello. Thank you for replying. 

> Plan to merge it for Blender 4.4

Could you please clarify why it is not possible to merge it for Blender 4.3.1? The changes are minor. 

> Seems your format options is not aligned with the ones that we are using here

It should be fixed now.

julienduroure:
> Could you please clarify why it is not possible to merge it for Blender 4.3.1? The changes are minor.

Branch freeze before the release will happen in a few hours

The0Dev:
>Branch freeze before the release will happen in a few hours

In a few hours Blender 4.3 is going to be released if I'm not mistaken. It seems that Blender 4.4 is not expected until March 2025, which seems quite a long wait. I kindly hope that it might be possible to include this change in an earlier version (there will probably be at least 4.3.**1**?). Thank you for considering my request!

julienduroure:
Any 4.3.x is only for bug fixing.

This will be merge in main branch that will become 4.4

The0Dev:
I've realized that there is an issue. The animation pointer feature is currently only available in the "Scene" and "NLA tracks" modes. In these modes, all objects and **data** properties (when Animation Pointer is toggled) are sampled and baked, **even when** they are not animated by the user. As the result, using the "force keeping channel for objects/data" options causes **every** object and property (that can be exported via KHR_animation_pointer) to always have at least 2 keyframes and clutters resulting gltf files with multiple unnecessary channels. That seems to be **the intended behavior**. However, I'd like to only keep channels and keyframes for properties that I animate, which is currently impossible using those modes. So unfortunately this PR is not very useful for me at the moment unless it becomes possible to use Animation Pointer in the aforementioned modes or if some other changes are made.

PR #2413, opened 2024-11-22
Expose the ability to limit exporting renderable objects via Exporters

ottah:
This change adds the ability to limit exporting objects that are disabled from rendering like in the file exporter.

blender_2MTrP1FWcV


julienduroure:
Hello,

The "include" section of the exporter (via file export) is hidden when exporting via Collections.
This is not a bug in glTF exporter, but a design done at Blender level for all exporters (usd, fbx, obj, etc...)

So closing this PR, as we will keep this exporter aligned with others.
Feel free to open a discussion thread on https://devtalk.blender.org/ if you want to discuss it with Blender devs

PR #2469, opened 2025-01-23
Add pre-export & post-export hooks

julienduroure:
- Add 2 regular hooks for pre-export & post-export
- Add example for legacy functions that already doing it


julienduroure:
Hum... Just discover that there are some glTF2_pre_export_callback & glTF2_post_export_callback

PR #2480, opened 2025-02-01
Followup 133868

julienduroure:
- Use slot identifier instead of slot handle
- change API calls after changes in Blender (in PR 133868)
- Refactoring, avoiding duplicate functions


julienduroure:
Note: This PR will failed until https://projects.blender.org/blender/blender/pulls/133868 is merged

Issue #2446, opened 2024-12-18
Avoid Redundant Mesh Exports for Multiple Geometry Modifiers


PMZeroSkyline:
Summary: When exporting to GLTF 2.0 format, if multiple geometry node modifiers reference the same mesh through properties, the current exporter generates multiple copies of the same mesh, one for each modifier. This results in redundant meshes being included in the exported GLTF file. The requested feature is to optimize the export process so that the referenced mesh is only exported once, and subsequent references in different geometry node modifiers simply point to the original mesh.

Detailed Description: In scenarios where a single mesh is referenced multiple times through different geometry node modifiers (e.g., in a procedural setup or instances), the current exporter exports multiple identical copies of the same mesh for each reference. This redundancy can lead to larger file sizes and unnecessary duplication in the GLTF output.

It would be more efficient if the exporter could recognize when multiple geometry node modifiers refer to the same mesh (via properties), and instead of exporting the mesh multiple times, it would export it once and reference it in the node hierarchy. This would reduce the size of the exported GLTF file and improve the performance when loading the model in a 3D viewer or engine.

Expected Behavior:

    If a mesh is referenced by multiple geometry node modifiers (via property references), the exporter should export the mesh once.
    In the GLTF file, the mesh should be defined only once, and each reference to it (from different geometry node modifiers) should point to the original mesh.
    The exporter should handle this optimization automatically without requiring manual adjustments from the user.

Benefits:

    Smaller GLTF file sizes.
    Faster export times due to reduced duplication.
    Better memory management when the GLTF file is loaded into a 3D engine or viewer.

Example Scenario:

    Mesh Mesh_A is referenced by geometry modifier 1 and geometry modifier 2 through a property link.
    Current exporter generates two separate copies of Mesh_A in the GLTF file.
    With the requested feature, the exporter would only include Mesh_A once, and both geometry modifiers would reference the same mesh.

Additional Notes:

    This feature would be especially helpful for complex procedural node setups or instances that reuse the same geometry multiple times.
    The GLTF format already supports referencing objects such as meshes via their ID or URI, so this optimization should be straightforward to implement.

julienduroure:
Hello,
Please provide an example of .blend file, with your export settings saved inside the file.

Note that exporting Geometry Node instances is experimental. The GN node datamodel is far from being compatible with glTF Specification, so we will not be able to fully support all Blender features

PMZeroSkyline:
Hello, thanks for your reply

[...]

PR #2490, opened 2025-02-18
Fix instance eval object

MattiasFredriksson:
Description
----
Fix for #2446. Geometry node (GN) instances are currently exporting the 'evaluated depsgraph object' rather then the instantiated (or 'original') mesh data block. Proposed changes will instead export the original un-evaluated mesh when the object has no modifiers applied.

Initially did not consider modified instantiated objects, generally the evaluated object will be exported. While a slight improvement it is a bit tricky as it would be an edge case. Due to how 'apply modifier' option works it will still export separate mesh entries for those cases even if they are both evaluations of the same object.

Changes
----
- Export the original GN instantiated mesh object when possible (it has no modifiers)
- Unittest with .blend validating that there are no additional `meshes` entries exported for the unmodified case and copies are exported in the case with modifiers!
- Tested on Blender 4.4 beta version.


julienduroure:
Hello,
Thanks for the PR.
I will have a look on it soon

MattiasFredriksson:
No worries, it is an improvement to current behavior but I think it's more a bandaid and doesn't directly solve the underlying problem.

The core issue is mainly that the importer has two options that are handled independently: 'Apply Modifier' and 'GN Instances'. Selecting one and not the other exports parts of the evaluation but not the other part (e.g. the GN modified mesh output + GN instances). What happends are these issue where one part is 'evaluated' and the other part isn't.

Addressing those issues are a bit complicated though because the mesh node is re-evaluated at export and is another copy. Improving all of those behaviors is a bit of a larger issue so it would be a good idea to get your opinion on the topic and these changes before commiting to trying to do a larger change.

julienduroure:
We are currently working with Blender core dev to get an exporter evaluation graph that get all data needed for geometry export, making this PR and current GN related code no more relevant.
Let's keep this PR open until I get more info about it, and until I tested it

MattiasFredriksson:
Yes [this change](https://projects.blender.org/blender/blender/pulls/135318) should make things better. Will close this as I intended to do so but forgot about it!

PR #2491, opened 2025-02-18
Fix 'dev/null' for windows

MattiasFredriksson:
Description
----
Corrected stdio output target to "NUL" for windows platform, currently unittests does not run on Windows.

Comments
---
Might be a better solution to pass the 'silent' option to exec() call rather then piping to dev/null but this seemed like the simplest solution as my javascript isn't great.

Could run most test but some tests fail, unsure of this is windows specific or not. Tests that fail are mostly related to the name of the exported '.png' files for scenes/08 tests (see image below) and some UDIM counter issue.

image

### DIFF
--- a/tests/test/test.js
+++ b/tests/test/test.js
@@ -250,7 +250,8 @@ describe('Exporter', function () {
 
         variants.forEach(function (variant) {
             var args = variant[1];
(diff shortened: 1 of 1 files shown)


julienduroure:
Thanks!
Merging it.

The UDIM name is strange, as this is not happen locally here (linux) and in github

PR #2503, opened 2025-02-27
Fix #423 - import multiple scenes as collections

### LINKED ISSUE #423: Import gltf scenes as collection, all in current scene
Import gltf scenes as collection, all in current scene

julienduroure:
Fix #423 - import multiple scenes as collections

- For single scene import => All is created in active collection
- For multiple scenes import => Each scene is imported as a collection. Non default scene are excluded from view_layer
- If there are some orphan nodes (not in any scenes) => An Orphan Collection is created (excluded too)

image

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -1931,6 +1931,12 @@ class ImportGLTF2(Operator, ConvertGLTF2_Base, ImportHelper):
         default=True,
     )
 
+    import_scene_as_collection: BoolProperty(
+        name='Import Scene as Collection',
+        description='Import the scene as a collection',
+        default=True,
+    )
+
     def draw(self, context):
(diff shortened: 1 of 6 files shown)


aaronfranke:
Is there a reason that scenes aren't imported as scenes, like the exporter already does?

Screenshot 2025-02-27 at 9 24 20 AM

julienduroure:
Blender put more and more effort in collection management instead of scene management.

Cf Blender Collection Exporters added in 4.2
And Collection Importers may come one day (see https://projects.blender.org/blender/blender/issues/132171)

At some point, I want to experiment exporting direct children collection as scenes in exporter too (not planned for now)

Blender glTF exporter of multiple scene is not perfect, we have some issue regarding update of depsgraph & mesh evaluation

Anyway, I just added a fallback to create multiple scenes

PR #2530, opened 2025-04-12
Fix #1353

### LINKED ISSUE #1353: Import: exception when importing glTFs with long extra names
This was reported at https://blender.stackexchange.com/questions/215658/glb-import-error-due-to-idproperty-limitation.

Attempting to import a file with long extra names produces this error

    File "/home/scurest/blender-2.93.0-f362dad19bf4-linux64/2.93/scripts/e_gltf2/blender/com/gltf2_blender_extras.py", line 91, in set_extras
        blender_element[custom_property] = value
    KeyError: 'the length of IDProperty names is limited to 63 characters

<details>

<summary>Sample .gltf</summary>

```
{
  "asset": { "version": "2.0" },
  "nodes": [
    {
      "extras": {
        " But the perturbed Man away turns down the valleys dark;\n[Saying. We are not One: we are Many, thou most simulative]": 0
      }
    }
  ]
}
```

</details>

The code is line 91 here

https://github.com/KhronosGroup/glTF-Blender-IO/blob/bb488b245f0c83fdd566b43c2c762d469c044326/addons/io_scene_gltf2/blender/com/gltf2_blender_extras.py#L90-L93

The try shows this code was already problematic once before (#884), so I think **the catch should be broadened from `TypeError` to `Exception`** to prevent future problems like this from making the whole export fail.

As for actually importing the property with a too-long name, FBX does this to make the name good for use in Blender. We could do the same thing.

```py
def validate_blend_names(name):
    assert(type(name) == bytes)
    # Blender typically does not accept names over 63 bytes...
    if len(name) > 63:
        import hashlib
        h = hashlib.sha1(name).hexdigest()
        n = 55
        name_utf8 = name[:n].decode('utf-8', 'replace') + "_" + h[:7]
        while len(name_utf8.encode()) > 63:
            n -= 1
            name_utf8 = name[:n].decode('utf-8', 'replace') + "_" + h[:7]
        return name_utf8
    else:
        # We use 'replace' even though FBX 'specs' say it should always be utf8, see T53841.
        return name.decode('utf-8', 'replace')
```

julienduroure:
Fix #1353 - Fix set extras with long name at import

### DIFF
--- a/addons/io_scene_gltf2/blender/com/extras.py
+++ b/addons/io_scene_gltf2/blender/com/extras.py
@@ -89,11 +89,26 @@ def set_extras(blender_element, extras, exclude=[]):
         if custom_property in exclude:
             continue
 
+        secured_custom_property = _set_unique_name(blender_element, custom_property)
+
         try:
-            blender_element[custom_property] = value
+            blender_element[secured_custom_property] = value
         except Exception:
             # Try to convert to string
             try:
-                blender_element[custom_property] = str(value)
+                blender_element[secured_custom_property] = str(value)
             except Exception:
                 print('Error setting property %s to value of type %s' % (custom_property, type(value)))
+
+def _set_unique_name(blender_element, name):
+    if len(name) >= 64:
+
+        # Find the first available name of form truncated_name.XXX,
+
+        suffix = 1
+        name = name[:59] + '.' + str(suffix).zfill(3)
+        while name in blender_element.keys():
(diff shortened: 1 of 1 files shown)


scurest:
The reason the FBX importer uses the hash is to keep the name unique. Otherwise if you have "long name A" and "long name B", when you see "long n.001" and "long n.002", you can't know which one was which.

PR #2546, opened 2025-05-08
Fix #2542

### LINKED ISSUE #2542: Variant issue
Not sure how this happen, but file from #2525 has an issue in the definition of variants (for mesh with modifier).
This leads to a non valid glTF file

- [X] Add a check to prevent exporting double variant for same primitive
- [X] Investigate & fix the root cause
- [x] Variant with same names are not exported correctly => Will be managed in another ticket

julienduroure:
Fix #2542


julienduroure:
Must be merged after #2543

Issue #2423, opened 2024-11-27
Exporting the GPU instance together with other simple TRS animations will cause the animation to play incorrectly.


Maiev-42:
**Describe the bug**
If the gpu instance is exported together with the animation of other objects, it will cause errors in various viewers or the animation cannot be played.

**To Reproduce**
1. I now make a set of gpu instances according to the specifications in blender, and then create a new object to make a simple TRS animation
![1](https://github.com/user-attachments/assets/501f36fb-ad1a-4594-ac5b-2078b3e59793)

2.Then I checked the gpu instance, exported one copy via NLA, exported one copy via merge animation and viewed them in multiple editors
![2](https://github.com/user-attachments/assets/add5a89f-2239-4baf-bda7-90bef5cb0b70)

3. The following are links to editors used for testing, none of which work properly
If using merge animation, half of them cannot be imported, and the other half cannot view the animation
If using nla, most of them can be imported and see the animation, but cannot be played
https://sandbox.babylonjs.com/
https://github.khronos.org/glTF-Sample-Viewer-Release/
https://playcanvas.com/viewer
https://gltf-viewer.donmccurdy.com/
https://threejs.org/editor/
![3](https://github.com/user-attachments/assets/caf781be-b257-47b2-a7c6-0bb9d5bf2dba)

**Expected behavior**
Hopefully the animation can be exported and played normally.

**.blend file/ .gltf (mandatory)**
[test.zip](https://github.com/user-attachments/files/17928749/test.zip)

**Version**
 - OS: window11
 - Blender Version  4.3.1 rc
 - addon version 4.3.47

ramirofages:
I want to add that animating the parent node of a group that is to be exported as instances, breaks the entire export.

<img width="792" height="319" alt="Image" src="https://github.com/user-attachments/assets/e50225a8-f82a-4424-a7ba-f0510540f3aa" />

In the above image only the animated_root has an animated position. When exporting it and using [donmccurdy viewer](https://gltf-viewer.donmccurdy.com/)

in the report section I see the following error:

<img width="1138" height="210" alt="Image" src="https://github.com/user-attachments/assets/25464497-81c9-4017-a37c-b90ccde44194" />

I wish I could upload the blend file but the test case is pretty simple. Let me know if I can be of help in any way

Issue #2199, opened 2024-04-12
glTF Exporter Hangs with Large-Scale Instancing


Kyozon:
**Describe the bug**
When attempting to export a Blender scene containing a single, heavily instanced object (over 1 million instances) using the glTF exporter, the process hangs indefinitely with no errors shown in Blender's output log.

**To Reproduce**

1. Open the provided .blend file (glTF_Instancing_Stress.zip).
2. In the 'GScatter -> Systems' collection, select the 'glTF Bug' System.
3. Open the glTF Exporter.
4. Configure export settings:
      - Enable "Export Selection Only".
      - Under "Data/Scene Graph", enable "Geometry Nodes" and "GPU Instancing".
      - Disable "Animation", "Skinning", and "Shape Keys".
5. Click "Export".

**Expected behavior**
glTF file exported successfully. 

**Actual Behavior**

The glTF exporter hangs indefinitely, with no progress updates or error messages.

**Screenshots**

Here is a screenshot showing the object to export. Please be aware that once you open the project, you won't have the usual 3D viewport. That is because the number of instances is too much for the viewport to handle, so upon export I change to another context to workaround this issue.

Additionally, the GScatter plugin I'm using in the original NDA scene is bundled to repro the issue 1:1.

![glTF_Bug - Object to Export](https://github.com/KhronosGroup/glTF-Blender-IO/assets/35878729/08da5702-50a0-4c66-a52d-a4a182afd5ad)

**.blend file/ .gltf (mandatory)**

[glTF_Instancing_Stress.zip](https://github.com/KhronosGroup/glTF-Blender-IO/files/14952765/glTF_Instancing_Stress.zip)

**Version**
 - OS: Windows 11 23H2 
 - Blender Version: Blender 4.1 
 - Workstation with 256GB RAM/56 CPU Cores. 

**Additional context**
I'm specifically interested in whether there are known instance count limitations within glTF, or if this behavior suggests a potential bug in the exporter.

julienduroure:
Hello,
I am not even able to open the file on my computer.
Can you please create a smaller file?
I know that we are talking about performance here, but without able to open the file...

Maybe you can decrease the instance number ?
Have a file that I can open, and take some time to export (between 3 and 10 minutes) will be ok to check the performance profile, but, at least, I can open it

Kyozon:
Hi there, 

The previous file had about 9 million instances, this new one should have about 187K approximately. 

[...]

PR #2577, opened 2025-06-25
GPU instances nodes removal fixes

2frac:
It reworks code dealing with nodes instancing - in particular part which removes nodes added to instancing.

### What it does

- Gathers removed nodes from all scenes and applies deletion at once.
(in old code self.nodes_idx_to_remove is cleaned up in loop(not before it), which will cause to only delete nodes from last processed scene)

- Uses sets to check if node is deleted, instead of lists.
This probably has the most effect on speed for big-sized instance collections(should help with https://github.com/KhronosGroup/glTF-Blender-IO/issues/2199 )

- Instead of calculating shifts for node indices it uses single dict mapping.
I haven't checked exact performance effect, but to iterate all deleted objects for every remaining object can result in bad speed for specific scenes. There is still some O(n^2)-ish speed behavior, but i suspect it rather related to node creation and better be investigated separately.
Also results in simpler code.

- Applies new node index to animation's target.
Previously it was just checking if node is deleted, without changing index to new.
Should solve https://github.com/KhronosGroup/glTF-Blender-IO/issues/2423

### Examples to reproduce

All made in blender 4.4.3. For every example, both GPU instances and GN instancing should be enabled.

- [big_instancing.zip](https://github.com/user-attachments/files/20908874/big_instancing.zip)

amount of instances can be controlled by density
for 40k instances it takes 25 seconds(before fix) vs 10 second(after)for me.
for 100k - 155 vs 48.

- [two_scenes.zip](https://github.com/user-attachments/files/20908930/two_scenes.zip)

When exported, result is looking strange for "Scene", which likely to be caused by bad indices.
Also some "GN Instance" objects can be seen in node list in JSON, even if GN instancing is enabled as option and it should remove it.

- [animation_deleted.zip](https://github.com/user-attachments/files/20908933/animation_deleted.zip)

When exported, animation target has node ids in JSON, which is too big to be real.

### Note

There is more optimizations i would like to try(beside said creation), but better be done as separate pull request[s].

Instancing generally looks like it be better done using direct transform gathering from GN/instancing/particles rather than generating and unifying nodes - but that will require entirely different approach with its own problems.

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/exporter.py
+++ b/addons/io_scene_gltf2/blender/exp/exporter.py
@@ -278,26 +278,23 @@ class GlTF2Exporter:
             self.__traverse(holder.extensions)
 
             # Remove children from original Empty
-            new_children = []
-            for child_idx in node.children:
-                if child_idx not in insts:
-                    new_children.append(child_idx)
-            node.children = new_children
+            insts_set = set(insts)
+            node.children = [child_idx for child_idx in node.children if child_idx not in insts_set]
 
             self.nodes_idx_to_remove.extend(insts)
 
         for child_idx in node.children:
-                child = self.__gltf.nodes[child_idx]
-                self.manage_gpu_instancing(child, also_mesh=child.mesh is not None)
+            child = self.__gltf.nodes[child_idx]
+            self.manage_gpu_instancing(child, also_mesh=child.mesh is not None)
 
     def manage_gpu_instancing_nodes(self, export_settings):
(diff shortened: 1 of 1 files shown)


julienduroure:
Can you please rebase with main branch?
This should solve the tests now

2frac:
(i meant to rebase instead of merge main, but result should be same)

julienduroure:
Thanks for the PR, I will have a look soon!

PR #2578, opened 2025-06-26
get 'allow_embedded_format' as property instead of as mapping entry

2frac:
In [that commit](https://github.com/blender/blender/commit/7276b2009a8819619263a1b897389662307b0ee0) AddonPreferences(and some other classes) became inaccessible as dict-like object. Instead, fields should be access as attributes.

Property in question should work also for versions <5, still.

### DIFF
--- a/addons/io_scene_gltf2/__init__.py
+++ b/addons/io_scene_gltf2/__init__.py
@@ -151,9 +151,8 @@ def get_format_items(scene, context):
               'Exports multiple files, with separate JSON, binary and texture data. '
               'Easiest to edit later'))
 
-    if bpy.context.preferences.addons['io_scene_gltf2'].preferences \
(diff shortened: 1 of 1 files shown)


julienduroure:
Thanks ! I will have a look

julienduroure (changes requested):
I confirm that we also must change this line: 

https://github.com/KhronosGroup/glTF-Blender-IO/blob/d97e93200cff331b7d58bb8347237740fd7ccd89/addons/io_scene_gltf2/blender/com/gltf2_blender_ui.py#L351

To test it:
Create a variant, and assign the material(s) to current Variant.
In Material property:

- If "Please create a variant first" => KO
- If display a list => OK

julienduroure (approved):

PR #2585, opened 2025-08-06
Collection Exporter: Allow choosing which axis to center the collection on collection export

gguishe:
When exporting collections with the `Export at Collection Center`
option enabled, this enables to choose which axis to center the collection,

----
image


julienduroure:
Hello,
I am trying to understand what is your need about aligning only some axis?

We have to be careful about adding more and more options ( time of maintenance, difficulty for users to manage so many options, etc...)

Issue #2591, opened 2025-08-26
Linked collection instances are exported as empty while using "Full Collection Hierarchy" and "Limit to: Visible Objects" options


rpaciorek:
**Describe the bug**

While exporting with "Full Collection Hierarchy" and "Limit to: Visible Objects" options enabled linked (to other blend file, e.g. asset library) collections are exported as empty. Even though the collection is visible in Blender.

**Expected behavior**

Turning on/off the "Limit to: Visible Objects" option should not affect the export of objects that are visible at the Blender level.

**Cause**

This is caused by `visible_get()` returning `false` when called on objects inside the collection instance.

Using the "Full Collection Hierarchy" option causes `recursive_node_traverse` is call without `is_children_in_collection=True` for objects inside the collection during processing of the scene hierarchy:

https://github.com/KhronosGroup/glTF-Blender-IO/blob/6347351d688b2969cf6c48d8a949c9a5416c6c9e/addons/io_scene_gltf2/blender/exp/tree.py#L394-L395

This results in the `CHILDREN_IS_IN_COLLECTION` flag not being set in:

https://github.com/KhronosGroup/glTF-Blender-IO/blob/6347351d688b2969cf6c48d8a949c9a5416c6c9e/addons/io_scene_gltf2/blender/exp/tree.py#L208-L210

(because `is_children_in_collection` is `False`) and the node not being processed correctly in:

https://github.com/KhronosGroup/glTF-Blender-IO/blob/6347351d688b2969cf6c48d8a949c9a5416c6c9e/addons/io_scene_gltf2/blender/exp/tree.py#L607-L608

If we do not use the "Full Collection Hierarchy" option `recursive_node_traverse` is call with `is_children_in_collection=True`:

https://github.com/KhronosGroup/glTF-Blender-IO/blob/6347351d688b2969cf6c48d8a949c9a5416c6c9e/addons/io_scene_gltf2/blender/exp/tree.py#L375-L382

and linked collection is exported correctly.

**Solution**

Adding `is_children_in_collection=True` to `recursive_node_traverse` in tree.py:394

https://github.com/KhronosGroup/glTF-Blender-IO/blob/6347351d688b2969cf6c48d8a949c9a5416c6c9e/addons/io_scene_gltf2/blender/exp/tree.py#L394-L395

fixes this issue

**Screenshots**
Original scene:
[...]

Issue #2589, opened 2025-08-24
some instances of collection are exported as empty while using "Full Collection Hierarchy" option


rpaciorek:
**Describe the bug**
While exporting with "Full Collection Hierarchy" option enabled only original collection and zero or one instance content are exported.

Only content of instance with the same name as original collection are exported. All other instances are exported as empty object. In particular, a collection from the asset library may not be exported at all (depending on if its name has been changed or not).

**Expected behavior**
Content of all instances of collection are exported using the same meshes.

**Screenshots**
Original scene:
![Image](https://github.com/user-attachments/assets/51ed08c6-7fc3-4e13-8ae6-3daa2b6036aa)

After Export-Import:
![Image](https://github.com/user-attachments/assets/21b5caad-40d4-4ea1-bb99-67870dddabbc)

**.blend file/ .gltf (mandatory)**

[gltf_export_collection_test.zip](https://github.com/user-attachments/files/21956442/gltf_export_collection_test.zip)

**Cause, Solution, ...**

This is caused by condition in tree.py:L392 comparing collection names: https://github.com/KhronosGroup/glTF-Blender-IO/blob/6347351d688b2969cf6c48d8a949c9a5416c6c9e/addons/io_scene_gltf2/blender/exp/tree.py#L392

I don't know the idea behind this condition. If it was to prevent "duplicate collections" from being exported, this condition doesn't guarantee that (a collection can be exported 0, 1, or 2 times).

Removing this condition solves the problem with missing collection content. Meshes (with modifiers) are duplicated (when use apply modifiers option), but that's a separate issue (#1952). Other meshes are linked correctly.

**Version**
 - OS: Linux
 - Blender Version: 5.0.0-alpha+main.bd6c4d73a0fd-linux.x86_64-release
- glTF-Blender-IO: current main (6347351d688b2969cf6c48d8a949c9a5416c6c9e)

julienduroure:
I can confirm

PR #2602, opened 2025-10-19
fix 2589 and 2591

rpaciorek:
fix proposition for #2589 and #2591


julienduroure:
Hello,
Any chance to add some test cases?
I will help to avoid any regression later

rpaciorek:
Hi,
added.

I'm not sure in which numerically prefixed group they should be, so I put its to `01`.

PR #2603, opened 2025-10-27
Improve unlinking of objects from collections

harpoonlobotomy:
Handle case where the custom bone icosphere cannot be unlinked from collection, by unlinking from gltf.active_collection instead. Also ensures the active collection is then unhidden from the viewport so newly imported objects aren't hidden inside. 

If a collection is active but hidden, after creating the Icosphere Blender will give error message
"...\scripts\addons_core\io_scene_gltf2\blender\imp\node.py", line 170, in armature_display
    bpy.context.collection.objects.unlink(bpy.context.object)
RuntimeError: Error: Object 'Icosphere.001' not in collection 'Scene Collection' "
and halt, leaving the icosphere linked to both the active+hidden collection and the glTF_not_exported collection, with no object imported. 

This PR fixes that. 

Tested on Blender 5.1.0A.

### DIFF
--- a/addons/io_scene_gltf2/blender/imp/node.py
+++ b/addons/io_scene_gltf2/blender/imp/node.py
@@ -177,7 +177,18 @@ class BlenderNode():
                 1, 1, 1))
         bpy.data.collections[BLENDER_GLTF_SPECIAL_COLLECTION].objects.link(bpy.context.object)
         gltf.bone_shape = bpy.context.object.name
-        bpy.context.collection.objects.unlink(bpy.context.object)
+        try:
+            bpy.context.collection.objects.unlink(bpy.context.object)
+        except:
+            # If the collection was not in context, use active_collection instead.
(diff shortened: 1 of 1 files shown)


julienduroure:
I am really sorry, but I am still not able to reproduce :(
Can you maybe create a video, starting from an empty file?

harpoonlobotomy:
> I am really sorry, but I am still not able to reproduce :( Can you maybe create a video, starting from an empty file?

I'm sorry too, I didn't for this to be so long-running! And yes, I'll just do a screen recording from the very start on a fresh file and link it here.

harpoonlobotomy:
Sorry for the delay, things have been... hectic, let's say.  Reading my original PR back, I'm not surprised at all you couldn't reproduce anything, my descriptions were awful. It's only now I've forgotten enough that I can see it. Thank you for your patience.

Once I can reproduce it myself, I'll report back if the bug is still present in the current version of Blender or not, and if it does I'll finally get you that screen recording.

Issue #870, opened 2020-01-10
User extensions: empty collections and null


scurest:
Any key-value pair whose value is `{}`, `[]`, or `None` gets scrubbed out of the glTF data when exporting. That means a user-extension can't do things like this:

````
    "extensions": { "KHR_materials_unlit": {} }
    # gets turned into "extensions": {}
````

The way the in-tree code works around this is to use a hard-coded whitelist of key-names that shouldn't be scrubbed.

https://github.com/KhronosGroup/glTF-Blender-IO/blob/3ab37ba38a3ae483d69a029f979286ded8b9b94b/addons/io_scene_gltf2/blender/exp/gltf2_blender_export.py#L108-L115

That's not very robust and it doesn't help user extensions anyway. Here's my [proposed fix](https://github.com/KhronosGroup/glTF-Blender-IO/compare/master...scurest:empty-cols-in-json):

* Add some new objects you should use wherever you want an empty collection/null

    ````py
    EmptyDict = object()  # use this where you want an empty dict
    EmptyList = object()  # use this where you want an empty list
    Null = object()  # use this where you want a null
    ````
* Make `from_extension` believe that `EmptyDict` is a dict so it will allow it in the `"extensions"` object.
* Change `__fix_json` to convert `EmptyDict` into `{}` etc. Get rid of the whitelist.
* Whenever you want to put an empty dict, just use `EmptyDict` instead. Example:

    ````py
    extensions['KHR_materials_unlit'] = Extension(
        'KHR_materials_unlit',
        EmptyDict,
        required=False,
    )
    ````

What's everyone think?

cc @jjcasmar

jjcasmar:
I agree. I have the same doubts when I first looks at the __fix_json. I think your solution is a good one.

emackey:
Do 3rd-party addons have to import `EmptyDict` and friends?  What does that code look like?

I certainly agree that having a hard-coded list there is not acceptable now for 3rd-party extensions.

scurest:
They just do it the same way they import everything else

````py
class glTF2ExportUserExtension:
    def __init__(self):
        import io_scene_gltf2
        self.Extension = io_scene_gltf2.io.com.gltf2_io_extensions.Extension
        self.EmptyDict = io_scene_gltf2.io.com.gltf2_io.EmptyDict
        
    def gather_material_hook(self, material, blender_material, export_settings):
        if material.extensions is None:
            material.extensions = {}
[...]

PR #2631, opened 2025-12-22
Allow Extensions to pass data through without __fixing

ChristopherBiscardi:
Compared to extras, extensions currently don't allow bypassing the `__fix_json` function, which results in extras being able to export values with empty objects, but extensions can not.

Here is an example of the same data exported through extras and extensions. Note how the extension data is empty. This is because keys with empty dicts are filtered out.

```json
"nodes":[
    {
        "extensions":{
            "EXT_my_extension":{
                "components":[
                    {}
                ]
            }
        },
        "extras":{
            "components":[
                {
                    "test_components::Marker":{}
                }
            ]
        },
        "mesh":0,
        "name":"Cube"
    }
],
```

This PR enables export extensions to define which of their glTF extensions should be passed through without using `__fix_json`. The result is that the keys with empty dict values are kept in the same way extras keeps them.

```json
"nodes":[
    {
        "extensions":{
            "EXT_my_extension":{
                "components":[
                    {
                        "test_components::Marker":{}
                    }
                ]
            }
        },
        "extras":{
            "components":[
                {
                    "test_components::Marker":{}
                }
            ]
        },
        "mesh":0,
        "name":"Cube"
    }
],
```

This is an example implementation for an export extension which writes a single glTF extension:

```python
def passthrough_extension_data(self, passthrough_extensions, gltf2_plan, export_settings):
    passthrough_extensions.append(glTF_extension_name)
```
---

This fix is essentially the same as was done for extras in https://github.com/KhronosGroup/glTF-Blender-IO/pull/1429

The implementation is a hook because it seems better for backward compatibility to allow extensions to opt-in.

I am currently trying to port some functionality from extras to an extension, and this difference in exporting data is blocking the transition. I can go into my use case more if it's relevant. I'm also happy to make any naming changes to the hook names, etc.

This PR also fixes https://github.com/KhronosGroup/glTF-Blender-IO/issues/870

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/export.py
+++ b/addons/io_scene_gltf2/blender/exp/export.py
@@ -76,7 +76,8 @@ def __export(export_settings):
 
     export_user_extensions('gather_gltf_extensions_hook', export_settings, exporter.glTF)
     exporter.traverse_extensions()
-
+    passthrough_extensions = []
+    export_user_extensions('passthrough_extension_data', export_settings, passthrough_extensions, exporter.glTF)
     # Detect extensions that are animated
     # If they are not animated, we can remove the extension if it is empty (all default values), and if default values don't change the shader
     # But if they are animated, we need to keep the extension, even if it is empty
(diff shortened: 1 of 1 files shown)


julienduroure:
Thanks for the PR.
I will have a look soon

PR #2638, opened 2026-01-13
Fix  #2569 - fix variant export

### LINKED ISSUE #2569: Variant with same names are not exported correctly

julienduroure:
Avoid 2 variants with same name
This will not fix older files, we need to manually update your variant names


julienduroure:
Note: This is revert on Blender, as this crash on some plateforms

PR #2651, opened 2026-02-16
Also export children of a instanced collection when gltf_hierarchy_fu…

HarrievG:
Currently, when an instanced collection has children and` gltf_hierarchy_full_collections` is set to `True` children of that object are NOT exported. 

To illustrate : 
image

In the image above, the SM_SignBorder_Medical_3 will not be exported, and the node SM_Bld_Wall_Doorframe_Door_02.013 will have 0 children.

With the fix SM_Bld_Wall_Doorframe_Door_02.013 will have a single child, SM_SignBorder_Medical_3

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/tree.py
+++ b/addons/io_scene_gltf2/blender/exp/tree.py
@@ -383,18 +383,32 @@ class VExportTree:
 
                 # Some objects are parented to instance collection
                 for child in blender_children[blender_object]:
-                    self.recursive_node_traverse(child, None, node.uuid, node.matrix_world,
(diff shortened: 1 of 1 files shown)


HarrievG:
I have added an commit that is the full fix . 
This works in both cases : with and without full collection hierarchy selected.

I have included the blend file i used to test it with, it contains 2 scenes. 1 "entity" scene with an collection that is instanced in several ways, also nested. 

I have used https://gltf-viewer.donmccurdy.com/ to test with and without full collection hierarchy selected.

[blend_and_glb.zip](https://github.com/user-attachments/files/25351587/blend_and_glb.zip)

Issue #2140, opened 2024-02-16
Perf on node tree traversal


julienduroure:
With this file: https://github.com/KhronosGroup/glTF-Blender-IO/files/11376083/vrm1_0.zip
**Without exporting animation** (that is another subject)

- Exporting materials : 21s
- Exporting with placeholder: 0.5s

Materials exported are really simple, without any textures.
The node tree is quite complex, but not compatible with glTF.

We may have some performance enhancement for these cases (complex node setup not compatible)

HailToDodongo:
Are there any update on this?
Since blender 4.1 up to the latest beta either vertex colors or gltf export has been broken for me and other people (referring to the issue in  #2278)
Meaning that any more complex node setups are unable to be exported without freezing blender.

I'm mainly asking since i need to decide to either write a custom exporter an abandon gtlf, or wait for a fix.
Right now i have to recommend people to keep using blender 4.0 because of this which is not ideal.

julienduroure:
@HailToDodongo This ticket and the one you are referencing are about performance, not vertex color.
Vertex Color "issue" is fixed in 4.2, cf #2233 

I think the performance about navigating in material node tree is here from the beginning. If you noticed some performance drop between 4.0 and 4.1, can you please create a new ticket with you test blend file?

HailToDodongo:
Ah sorry, by vertex colors i meant the change in behaviour in 4.1 that also caused issues. This was indeed fixed in 4.2.
As for performance (or rather the inability to export at all since 4.1) the model in #2278 also has this problem, the issue there is describing that.

I just wasn't sure where to comment, since a few issues where linked together as duplicates.

julienduroure:
4.1 introduces node group traversal, where 4.0 and prior does not go inside node group.
That explains why the file export instantly in 4.0, because the node tree is full of blender group. But nothing is retrieve and exporting in glTF material.

4.1 can go inside node groups. It takes time (like any other node traversal), but can retrieve more information to export.
In case you now that no material data can be retrieve, I suggest to not export any materials.

glTF specification can not handle all blender node tree, so you may want to adapt your node tree to get data, or ignore material at export.

HailToDodongo:
The issue with omitting materials is that i'm still interested in custom properties attached to it.
The plugin mostly uses nodes for visualization only (partially emulating N64 hardware), whereas the actual materials settings are in custom properties.
[...]

PR #2692, opened 2026-04-29
io_scene_gltf2: cache shader-graph traversal in `from_socket`

marciogranzotto:
# Cache shader-graph traversal in `from_socket`

## What this fixes

The recursive socket→shader-node walk in `from_socket` (`addons/io_scene_gltf2/blender/exp/material/search_node_tree.py`) re-walks shared sub-graphs once per upstream input branch. With no caching at the recursion level, materials whose shader graphs share nodes — a node group used by multiple BSDF inputs, or a `Mix` / `MapRange` / `Color Ramp` between an `Image Texture` and a `Principled BSDF` input — get traversed once *per branch they participate in*. Cost grows quadratically (or worse) in the number of shared nodes.

This addresses the long-standing `# TODO: cache these searches` comment that has lived above `from_socket` and gives a concrete fix to the perf concerns surfaced in #2140 ("I think the performance about navigating in material node tree is here from the beginning").

## Measured impact

On a real production scene — 185 materials, 217 textures, 24 linked libraries (mix of Poly Haven materials, custom Geometry Nodes assets, complex BSDF setups):

| Metric | Before | After |
|---|---|---|
| Blender export phase (BEGIN EXPORT → "Exported successfully") | **1h 56m 50s** | **27 s** |
| Speedup | — | **~260×** |
| Output `scene.glb` size | 6,151,248 bytes | 6,150,812 bytes (0.007% delta, all in vendor metadata) |
| Materials / textures / images / lights | 185 / 217 / 218 / 8 | identical |
| `extensionsUsed` | (full list) | identical |

End-to-end (Blender export + downstream Needle build pipeline that runs Basis/KTX2 / Draco compression on top of the glTF output) dropped from ~2h 8m to ~8.5 min on the same scene.

py-spy sampling of the unpatched export showed `__search_from_socket` recursion **~45 frames deep on every sample** (909 frames across 20 samples taken 30 s apart), with `<genexpr>` from `_bpy_types.py` iterating `socket.links` as the active leaf 85 % of the time. After this patch, the same scene's export is too short for sampling to be statistically meaningful.

## Side benefit: UI no longer freezes

The previous quadratic walk held the GIL for its entire duration (pure Python, single-threaded). On scenes where the walk took an hour or more, Blender's UI thread starved completely — the UI froze, no progress feedback, hard to tell whether the export was hung or making progress. With memoization, the work completes fast enough that the UI thread keeps responding and addon-side status logs stream in real time.

## Implementation

* **The inner recursion now returns relative ("suffix") paths** from its `start_socket` to each match, so identical sub-trees produce identical cached results regardless of the outer caller's traversal prefix. The previous code carried a mutable `search_path` accumulator and built absolute paths inline — incompatible with caching because the same sub-tree visited from two different prefixes would have produced two different cached results.

* **A per-call `memo` dict** lives for the duration of one outer `from_socket` call and is keyed on `(start_socket.as_pointer(), group_path_key)`. `group_path_key` is a tuple of node pointers for the materials traversal (where `group_path` is a `List[bpy.types.Node]`) or the tree's own pointer for the lamps traversal (where `gather_lights_punctual` passes the lamp's `ShaderNodeTree` directly instead of a list).

* **`GROUP_INPUT` is now a no-op** when the recursion is not inside a recorded enclosing group. The original code accessed `group_path[-1].inputs` unconditionally, which would `IndexError` on a top-level `GROUP_INPUT`. In practice the only callers that pass a non-list `group_path` (the lamps path) never ascend a group, so this is a defensive clarification rather than a behavior change. It keeps the cache key well-typed.

## What is *not* changed

* No new dependencies.
* No new public API. `from_socket`'s signature is identical.
* Filter behavior unchanged (same `Filter` / `FilterByName` / `FilterByType` semantics).
* Output is byte-near-identical: 436-byte delta on a 6 MB GLB, all in vendor metadata (Needle build-pipeline timestamps/hashes) that isn't emitted by the gltf2 exporter itself.

## Testing

* End-to-end export of a real production scene with deeply-shared shader graphs: passes with **0 errors, 0 warnings**, produces structurally equivalent `scene.glb`.
* Material counts, texture counts, image counts, light counts, `extensionsUsed` array, sample material color factors all match the unpatched run exactly.
* Materials path exercised (185 materials including Poly Haven assets with deep node-group chains).
* Lamps path exercised (12 area lights with emission node trees, where `from_socket` is invoked with a `ShaderNodeTree` as `group_path`).

## Branch strategy

This PR targets `blender-v5.1-release` for backport since the perf issue ships in 5.1.x. An equivalent change against `main` is on `perf/cache-from-socket-traversal` in the same fork (same diff modulo the `__get_socket_index` helper that exists in `main` but not in `blender-v5.1-release`). Happy to file the `main` PR separately if you'd prefer to review them in sequence.

cc @julienduroure

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/material/search_node_tree.py
+++ b/addons/io_scene_gltf2/blender/exp/material/search_node_tree.py
@@ -70,7 +70,6 @@ class NodeTreeSearchResult:
         self.group_path = group_path
 
 
-# TODO: cache these searches
 def from_socket(start_socket: NodeTreeSearchResult,
                 shader_node_filter):
     """
@@ -80,12 +79,29 @@ def from_socket(start_socket: NodeTreeSearchResult,
     :param shader_node_filter: should be a function(x: shader_node) -> bool
     :return: a list of shader nodes for which filter is true
     """
-    # hide implementation (especially the search path)
+    # Per-call memo so that shared shader sub-graphs (group nodes, mix nodes,
+    # multi-input nodes, etc.) are walked once. Without this the recursion
+    # below is at least quadratic in the number of shared input branches and
+    # dominates export time on complex materials.
+    memo: typing.Dict[typing.Tuple, typing.List[NodeTreeSearchResult]] = {}
+
+    def _gp_key(group_path):
+        # `group_path` is a list of group nodes for the materials traversal,
+        # but lamps pass their ShaderNodeTree directly. Hash by element
+        # pointers for the list case; otherwise by the tree's own pointer.
+        if isinstance(group_path, list):
+            return tuple(n.as_pointer() for n in group_path)
+        return ("tree", getattr(group_path, "as_pointer", id)())
+
     def __search_from_socket(start_socket: bpy.types.NodeSocket,
                              shader_node_filter: typing.Union[Filter, typing.Callable],
-                             search_path: typing.List[bpy.types.NodeLink],
-                             group_path: typing.List[bpy.types.Node]) -> typing.List[NodeTreeSearchResult]:
-        results = []
+                             group_path) -> typing.List[NodeTreeSearchResult]:
+        cache_key = (start_socket.as_pointer(), _gp_key(group_path))
+        cached_results = memo.get(cache_key)
+        if cached_results is not None:
+            return cached_results
+
+        results: typing.List[NodeTreeSearchResult] = []
         for link in start_socket.links:
             # follow the link to a shader node
             linked_node = link.from_node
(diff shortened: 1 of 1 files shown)


marciogranzotto:
@julienduroure can you take a look? This has been a game changer for me

julienduroure:
Hello,
Thanks a lot for this PR.

Performance of material node tree traversal is currently having a big refactoring (see [here](https://projects.blender.org/blender/blender/issues/152848)), planned for this spring/summer.

So these PR will not be merged as-is, but I will make it fit the refactoring.

Any chance to get access to your test file?

marciogranzotto:
@julienduroure good to know! I'll try to export a .blend with the project I was using, I changed it lot since I made this PR, but I'll test with and without the fix and come back to you with a file. Thank you for the review!

marciogranzotto:
@julienduroure huge unoptimized file: https://drive.google.com/file/d/1c9TRiLU5NvohYK_rVRVFG0zWsVu57h21/view?usp=sharing

On test runs I got:

| Scene state | Patch | Wall-clock export | Output GLB | Status |
|---|---|---|---|---|
| sala2_packed.blend (self-contained, instances-real, 1 489 objs) | **patched** | **80.7 s** | 2 712 453 836 B | ✓ completed |
| sala2_packed.blend (same as above) | unpatched | **≥ 30 min** | — | killed at 30-min cap, **206** primitives processed before the same single-primitive stall |

julienduroure:
Hello,
I started to have a look on your file.
Unfortunately, I don't have enough RAM to open and export it entirely.
I created a subset of the file, appending only a few collections into a new file.

For this subset, I don't have any enhancement of the performance.

- Blender 5.1 : 15,52s
- main ( 5.2 beta) : 16,09s
- your PR on main : 15,84s
- My implementation PR on main : 15,88s

Do you have a reference to one or more Collection(s) of the file where there is a signification enhancement of the performance ?

julienduroure:
We merged last week some enhancement (with caching and more).
Can you please test on recent build (5.3 alpha) ?
Closing this ticket, as we will not merge to 5.1

Feel free to test in recent build and propose a new PR

PR #2698, opened 2026-05-10
KHR_texture_basisu export

julienduroure:
This PR needs KTX2 management in Blender, this is not the case for now.

TODO: manage options of KTX2 for glTF export


walcht:
@julienduroure, I only saw this PR now (I had a local branch that handles KTX2 export options).

> TODO: manage options of KTX2 for glTF export

There are tons of details here. Exporting a KTX2 file has easily dozens or so of options per codec (e.g., for Basis Universal UASTC you have around a dozen parameters, same for BCn+RDO export (coming to libktx soon), same for ASTC export, etc).

Since I have been working on this since a long time, I can either open a PR against your `ktx_export` branch or submit the patch(es) here.


There are also a lot of caveats, for instance:

- if a user reads a glTF with DDS file then export it to KTX2 file, a warning should be shown that this operation will result in (significant?) quality loss. Likewise for reading KTX2 files, then re-exporting them (yes, even if unmodified).
- there are just so many parameters, that it is not ideal to write a KTX2 file without constant feedback on how the end texture will look like (obviously, what you see in Blender is not the same as the texture you will get when exported in KTX2. Currently, I have no idea how to mitigate this...).
- users may use this glTF addon to just bunder their KTX2 texture within a glTF format. There should be a way to detect that if the original input is KTX2 and un-modified => do not re-write and simply copy that input KTX2 file as is. Is there something like this already in this glTF export plugin? (I'm not familiar with this code base or even the functionalities of glTF export plugin).

addons/io_scene_gltf2/__init__.py
```diff
@@ -360,6 +360,8 @@ class ExportGLTF2_Base(ConvertGLTF2_Base):
                 'Be aware of a possible loss in quality'),
                ('WEBP', 'WebP Format',
                 'Save images as WebPs as main image (no fallback)'),
+               ('KTX2', 'KTX2 Format',
+               'Save images as KTX2 with BasisU supercompression'),
```
walcht, line 364: > Save images as KTX2 with BasisU supercompression

Not necessarily. KTX2 can be saved with uncompressed formats directly, or with ASTC/BCn native GPU-formats.

I think something like this is better:

> Save images as Khronos Texture container format (KTX2) with Basis Universal GPU texture codecs or direct native GPU texture formats.

Also, something like this should be added:
`Be aware of a possible loss in quality`

julienduroure: glTF extension KHR_texture_basisu is more retricted than the possibility of KTX

[See https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_texture_basisu](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_texture_basisu#ktx-v2-images-with-basis-universal-supercompression)

PR #2711, opened 2026-06-16
Fix #2691: keep image name suffixes like ".001" on export

### LINKED ISSUE #2691: GLB Export - Image Names Can't Contain Periods "."
**Describe the bug**
On export, the GLB exporter removes everything past the last `.` in image names. This easily results in situations where two images will get exported as the exact same name, and this corrupts the image names on GLB import into Blender (in some cases images will get a completely different random name). This makes it impossible to reliably reference image names in GLB JSON extras, as the names can change underneath you and become ambiguous.

**To Reproduce**
1. Open the attached [GLB Image Names.zip](https://github.com/user-attachments/files/27070753/GLB.Image.Names.zip)
2. Export as GLB
3. Import the GLB
4. Notice the image names are now scrambled.

**Expected behavior**
Export image names with periods:
`Image`
`Image.001`
`Image.Other.001`

**Screenshots**
Image
Resulting GLB JSON:
```JSON
  "images": [
    {
      "bufferView": 4,
      "mimeType": "image/png",
      "name": "Image" // <-- /!\
    },
    {
      "bufferView": 5,
      "mimeType": "image/png",
      "name": "Image" // <-- /!\
    },
    {
      "bufferView": 6,
      "mimeType": "image/png",
      "name": "Image.Other"
    }
  ],
```

**.blend file/ .gltf**
[GLB Image Names.zip](https://github.com/user-attachments/files/27070753/GLB.Image.Names.zip)

**Version**
 - OS: Windows 10
 - Blender Version 5.1.1

Builder106:
Fixes #2691.

### Problem
On export, image datablock names containing a `.` (Blender's own duplicate suffix, e.g. `Image.001`, `Image.Other.001`) lose everything after the last dot. Distinct images then collapse to the same glTF image `name` (`Image`, `Image`), which is ambiguous and, as reported, scrambles image names on re-import.

### Cause
In `exp/material/image.py`, `__gather_name` runs `os.path.splitext(img.name)` when combining names:

```python
for img in imgs:
    name, extension = os.path.splitext(img.name)   # "Image.001" -> "Image"
    names.append(name)
```

`img.name` is a **datablock** name, not a filename, so `splitext` wrongly treats `.001` / `.Other.001` as a file extension. (The shared-filepath branch a few lines above is already correct — it only strips a whitelisted real image extension.)

### Fix
Apply that same whitelist guard to the datablock-name loop: only strip a genuine image extension (`.png` / `.jpg` / `.jpeg` / `.webp`); otherwise keep the full name. This follows the guidance on the issue — "when possible, we need to do our best to export correctly the name, avoiding ambiguity."

### Verification
Three images named `Image`, `Image.001`, `Image.Other.001`, each on its own plane, exported to GLB (Blender 5.x):

| | `images[].name` in the GLB |
|---|---|
| before | `["Image", "Image", "Image.Other"]` (collision) |
| after  | `["Image", "Image.001", "Image.Other.001"]` |

Genuine file extensions are still stripped (`wood.png` → `wood`), and dotted names that are not extensions are preserved (`v1.5_diffuse` stays intact).

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/material/image.py
+++ b/addons/io_scene_gltf2/blender/exp/material/image.py
@@ -205,6 +205,11 @@ def __gather_name(export_image, use_tile, export_settings):
         names = []
         for img in imgs:
             name, extension = os.path.splitext(img.name)
+            # img.name is a datablock name, not a filename, so only strip a
+            # genuine image extension. Otherwise names like "Image.001" lose
+            # their suffix and collide with "Image" (#2691).
+            if extension.lower() not in ['.png', '.jpg', '.jpeg', '.webp']:
+                name = img.name
             names.append(name)
         name = '-'.join(names)
         return name or 'Image'


Waffle1434:
Thank you!

Builder106:
Hey @julienduroure, it's me again. I wanted to bump this one, it's been about three weeks. Someone ran into the same bug independently and left a thanks in the thread yesterday, so this looks like a real pain point beyond just the linked issue. The fix reuses the extension whitelist guard that's already applied a few lines up in the same function, just extends it to the datablock-name loop, so the diff stays small and consistent with the existing pattern rather than introducing a new approach.

julienduroure:
Merged, thanks!

PR #2716, opened 2026-06-25
Add operator presets for glTF import operation

dupoxy:


dupoxy:
The branch was renamed, and it broke everything. See https://github.com/KhronosGroup/glTF-Blender-IO/pull/2717 now.

PR #2721, opened 2026-07-05
Take into account strength from object data with node tree

baardaap-blender:
When a nodetree is present for a light, only the light strength from the nodetree was used for the GLTF export. However, both the strength in the object data and the strength in the nodetree influence the final light strength. This commit multiplies the strength from the nodetree with the object data strength/power.

I checked with Brecht, and muliplying is the right thing to do.


julienduroure:
Merged, thanks @baardaap-blender !

PR #2733, opened 2026-07-28
Fix #2732 - pass the original material to hook

### LINKED ISSUE #2732: gather_material_hook no longer provides access to Material properties
**Describe the bug**

I believe https://github.com/KhronosGroup/glTF-Blender-IO/pull/2664 changed the type of the `blender_material` argument (of `gather_material_hook`) from `Material` to `InlineShaderNodes`. Data placed on `bpy.types.Material` seems to no longer be available in the `gather_material_hook` export hook as a result.

**To Reproduce**

1. register a property on `Material`
	```python
	bpy.types.Material.test_gltf_property = bpy.props.IntProperty()
	```
2. try to access said property in the `gather_material_hook`. (in this example, the field `test_gltf_property` will not show up, because `blender_material` is *not* a `bpy.types.Material`; It is a `InlineShaderNodes`)
	```python
    def gather_material_hook(self, gltf2_material, blender_material, export_settings):
        print("gather_material_hook")
        print(blender_material)
        print(dir(blender_material))
	```

The Blender 5.2 output shows an `InlineShaderNodes` *without* the `test_gltf_property` prop, which is correct since the type is different, but the data in that prop from the `Material` is no longer accessible.

```
gather_material_hook
<InlineShaderNodes object at 0x171cbe4d0>
['__class__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', 'from_light', 'from_material', 'from_world', 'node_tree']
```

<details><summary>full test plugin for reproduction</summary>

A `blender_manifest.toml` and an `__init__.py` based on [the demo example in the `glTF-Blender-IO` repo](https://github.com/KhronosGroup/glTF-Blender-IO/tree/7551a35b146ab62ebc503ec661f88a74dc611392/example-addons/example_gltf_exporter_extension).

```toml
schema_version = "1.0.0"
id = "GLTF_EXT_example_extension_export"
name = "glTF user extension example for export"
version = "0.0.1"
tagline = "glTF user extension example for export"
maintainer = "Community"
type = "add-on"
permissions = ["files"]
tags = ["Import-Export"]
blender_version_min = "4.2.0"
license = ["SPDX:GPL-2.0-or-later"]
website = "https://github.com/KhronosGroup/glTF-Blender-IO"
copyright = ["The glTF-Blender-IO authors"]
```

```python
import bpy

bl_info = {
    "name": "Example glTF Extension",
    "category": "Generic",
    "version": (1, 0, 0),
    "blender": (5, 0, 0),
    'location': 'File > Export > glTF 2.0',
    'description': 'Example addon to add a custom extension to an exported glTF file.',
    'tracker_url': "https://github.com/KhronosGroup/glTF-Blender-IO/issues/",  # Replace with your issue tracker
    'isDraft': False,
    'developer': "(Your name here)", # Replace this
    'url': 'https://your_url_here',  # Replace this
}

# glTF extensions are named following a convention with known prefixes.
# See: https://github.com/KhronosGroup/glTF/tree/main/extensions#about-gltf-extensions
# also: https://github.com/KhronosGroup/glTF/blob/main/extensions/Prefixes.md
glTF_extension_name = "EXT_example_extension"

# Support for an extension is "required" if a typical glTF viewer cannot be expected
# to load a given model without understanding the contents of the extension.
# For example, a compression scheme or new image format (with no fallback included)
# would be "required", but physics metadata or app-specific settings could be optional.
extension_is_required = False

class ExampleExtensionProperties(bpy.types.PropertyGroup):
    enabled: bpy.props.BoolProperty(
        name=bl_info["name"],
        description='Include this extension in the exported glTF file.',
        default=True
        )

def register():
    # Register the properties
    bpy.utils.register_class(ExampleExtensionProperties)

    # If you plan to use the Scene Exporter (File => Export => glTF 2.0), you probably want to register the properties to the scene
    bpy.types.Scene.ExampleExtensionProperties = bpy.props.PointerProperty(type=ExampleExtensionProperties)

    bpy.types.Material.test_gltf_property = bpy.props.IntProperty()
    
    # If you plan to use Collection Exporters, you may want to register the properties to the collection, instead of the scene
    # bpy.types.Collection.ExampleExtensionProperties = bpy.props.PointerProperty(type=ExampleExtensionProperties)

    # Use the following 2 lines to register the UI for this hook
    from io_scene_gltf2 import exporter_extension_layout_draw
    exporter_extension_layout_draw['Example glTF Extension'] = draw_export # Make sure to use the same name in unregister()

def unregister():
    # Unregister the properties
    bpy.utils.unregister_class(ExampleExtensionProperties)
    del bpy.types.Scene.ExampleExtensionProperties

    # If you registered the properties to the collection, you should unregister them here
    # del bpy.types.Collection.ExampleExtensionProperties

    # Use the following 2 lines to unregister the UI for this hook
    from io_scene_gltf2 import exporter_extension_layout_draw
    del exporter_extension_layout_draw['Example glTF Extension'] # Make sure to use the same name in register()


def draw_export(context, layout):

    # Note: If you are using Collection Exporter, you may want to restrict UI for some collections
    # You can access the collection like this: context.collection
    # So you can check if you want to show the UI for this collection or not, using
    # if context.collection.name != "Coll":
    #     return


    header, body = layout.panel("GLTF_addon_example_exporter", default_closed=False)
    header.use_property_split = False

    props = bpy.context.scene.ExampleExtensionProperties

    header.prop(props, 'enabled')


# Note: the class must have this exact name

    def __init__(self):
        # We need to wait until we create the gltf2UserExtension to import the gltf2 modules
        # Otherwise, it may fail because the gltf2 may not be loaded yet
        self.Extension = Extension
        self.properties = bpy.context.scene.ExampleExtensionProperties
        self.is_critical = True

    def gather_material_hook(self, gltf2_material, blender_material, export_settings):
        print("gather_material_hook")
        print(blender_material)
        print(dir(blender_material))
```

</details>

**Expected behavior**
Prior to 5.2, properties defined on a `Material` were accessible in the `gather_material_hook`

**Screenshots**

A .blend (or .gltf) is not required for reproduction. Here is the plugin reproduction shown previously. Exporting to gltf with the default blender scene is enough to view the results in the console.

[temp-gltf-extension-repro.zip](https://github.com/user-attachments/files/30438452/temp-gltf-extension-repro.zip)

**Version**
 - OS: all platforms
 - Blender Version: 5.2.0 LTS

**Additional context**

I'm unsure if there's a new way to access the `Material` data in the export hooks from the material `InlineShaderNodes::from_material` uses, or if the change in type for the `blender_material` argument was unintentional.

julienduroure:
Make sure to not expose the inline material, as material properties are not exposed

### DIFF
--- a/addons/io_scene_gltf2/blender/exp/material/materials.py
+++ b/addons/io_scene_gltf2/blender/exp/material/materials.py
@@ -796,6 +796,8 @@ def __export_unlit(bmat, export_settings):
     if export_settings['gltf_extras'] and export_settings['gltf_export_anim_pointer']:
         export_settings['KHR_animation_pointer']['extras']['materials'][bmat.id]['glTF_extras'] = material
 
+    # Make sure to expose bmat.material (the original material), so users can retrieve additional proporties
+    # (These properties are not available on the inline material)
     export_user_extensions('gather_material_unlit_hook', export_settings, material, bmat.material)
 
     # Now we have exported the material itself, we need to store some additional data
--- a/addons/io_scene_gltf2/blender/exp/material/materials.py
+++ b/addons/io_scene_gltf2/blender/exp/material/materials.py
@@ -331,7 +331,9 @@ def gather_material(bmat, export_settings):
 
     mat_unlit, uvmap_info, vc_info, udim_info = __export_unlit(bmat, export_settings)
     if mat_unlit is not None:
-        export_user_extensions('gather_material_hook', export_settings, mat_unlit, bmat)
+        # Make sure to expose bmat.material (the original material), so users can retrieve additional proporties
+        # (These properties are not available on the inline material)
+        export_user_extensions('gather_material_hook', export_settings, mat_unlit, bmat.material)
         return mat_unlit, {"uv_info": uvmap_info, "vc_info": vc_info, "udim_info": udim_info}
 
     orm_texture = __gather_orm_texture(bmat, export_settings)
@@ -420,7 +422,9 @@ def gather_material(bmat, export_settings):
     if material.emissive_factor is not None and bmat.get_socket("Base Color").socket is None:
         material.pbr_metallic_roughness = gltf2_pbr_metallic_roughness.get_default_pbr_for_emissive_node()
 
-    export_user_extensions('gather_material_hook', export_settings, material, bmat.get_used_material())
+    # Make sure to expose bmat.material (the original material), so users can retrieve additional proporties
+    # (These properties are not available on the inline material)
+    export_user_extensions('gather_material_hook', export_settings, material, bmat.material)
 
     # Now we have exported the material itself, we need to store some additional data
     # This will be used when trying to export some KHR_animation_pointer
--- a/addons/io_scene_gltf2/blender/exp/material/materials.py
+++ b/addons/io_scene_gltf2/blender/exp/material/materials.py
@@ -796,7 +796,7 @@ def __export_unlit(bmat, export_settings):
     if export_settings['gltf_extras'] and export_settings['gltf_export_anim_pointer']:
         export_settings['KHR_animation_pointer']['extras']['materials'][bmat.id]['glTF_extras'] = material
 
-    export_user_extensions('gather_material_unlit_hook', export_settings, material, bmat.get_used_material())
+    export_user_extensions('gather_material_unlit_hook', export_settings, material, bmat.material)
 
     # Now we have exported the material itself, we need to store some additional data
     # This will be used when trying to export some KHR_animation_pointer


addons/io_scene_gltf2/blender/exp/material/materials.py
Exairnous: I think this should be modified also?
EDIT: Since GitHub doesn't seem to be showing the previous contents of line 799 properly, they are:
`    export_user_extensions('gather_material_unlit_hook', export_settings, material, bmat.get_used_material())`
```suggestion
    export_user_extensions('gather_material_unlit_hook', export_settings, material, bmat.material)
```

Exairnous (commented):
@julienduroure Thanks very much for the fix!  It looks mostly good, I have left one inline comment.

addons/io_scene_gltf2/blender/exp/material/materials.py
julienduroure: Thanks, fixed!

Exairnous:
With the latest commit, I think everything's good.  Thank you.
(Note: I tested with the Hubs Blender add-on, not the sample one from @ChristopherBiscardi in https://github.com/KhronosGroup/glTF-Blender-IO/issues/2732 but I expect it should have the same results)

Any chance this fix will get backported to Blender 5.2 LTS?

julienduroure:
> Any chance this fix will get backported to Blender 5.2 LTS?

@Exairnous Yes, I will merge it into Blender repo tomorrow, and put it on the list on 5.2.x LTS backport

Exairnous:
@julienduroure Excellent!  Thanks!

ChristopherBiscardi (approved):
This does seem to fix the issue, and my existing extension was able to regain the ability to handle data attached to Material in the gather_material_hook

PR #2766, opened 2026-09-19
Cleanup: Spelling in code

Xury46:
Fix misspelled words in class names, function names, and variable names throughout the code base.

To the best of my understanding, these are all internal names, and shouldn't break anything.
The `self.tree_troncated` variable seems to be unused, but I'm limiting this PR to spelling changes only, so I did not remove it.


julienduroure:
Merged, thanks!

PR #2767, opened 2026-09-19
Fix: Bugs due to typos

Xury46:
Fixes a few misspelled attribute names that would cause silent failures.

The var.name = "VariantName" change isn't really a bug, just inconsistent in style.
I noticed the default value is "Variant Name", so this line could actually be removed altogether so that it just uses the default name when creating a new variant.


julienduroure:
Merged, thanks!

